PR #140, opened 2024-12-25
ICICI Data API

psymbio:
Function for fetching data from ICICI Breeze API is added in: `broker/icici/api/data.py` and the call is changed from Fyers to ICICI here: `restx_api/history.py`, however, I believe this needs to be done in such a way that the switching between different brokers is handled in this file.

For testing I have added: `strategies/testing_icici_data_api.ipynb` but in this you'll be able to see the data is fetched but with a processing problem:

```bash
{'status': 'error',
 'message': "Failed to process historical data: 'timestamp'",
 'error_type': 'processing_error',
 'raw_data': [{'close': 242.4,
   'datetime': '2024-12-20 11:22:00',
   'high': 242.4,
   'low': 242.2,
   'open': 242.2,
   'volume': 16220},
```

wasn't sure exactly where in the code this is done but it's just that instead of timestamp ICICI calls it datetime.

### DIFF
--- a/broker/shoonya/api/data.py
+++ b/broker/shoonya/api/data.py
@@ -193,39 +193,15 @@ class BrokerData:
                     "to": str(end_ts)
                 }
                 
-                # Add debug logging
-                print(f"EOD Request payload: {payload}")
-                
-                response = get_api_response("/NorenWClientTP/EODChartData", self.auth_token, payload=payload)
-                
-                # Parse EOD response
-                if isinstance(response, list):
-                    data = []
-                    for item in response:
-                        if isinstance(item, str):
-                            item = json.loads(item)
-                        data.append({
-                            'timestamp': int(item['ssboe']),
-                            'open': float(item['into']),
-                            'high': float(item['inth']),
-                            'low': float(item['intl']),
-                            'close': float(item['intc']),
-                            'volume': float(item['intv'])
-                        })
-                    return pd.DataFrame(data)
-                else:
-                    raise Exception(f"Invalid response format for EOD data: {response}")
-            
-            # For intraday data, use TPSeries endpoint
+                print("EOD Payload:", payload)  # Debug print
+                try:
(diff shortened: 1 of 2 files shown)


psymbio:
Additional features to be added:

1. product_type: This will not always be "Cash" and needs to be updated as a string variable with possible values ("Futures", "Options", "Cash"). For simplicity, it could default to "Cash".
2. expiry_date: A string representing the expiration date.
3. right: A string variable to specify the type ("call", "put", or "others").
4. strike_price: A string to represent the strike price.

Based on: https://api.icicidirect.com/breezeapi/documents/index.html#gethistoricalchartslist

marketcalls:
Hi Recently added Added AngelAPI and removed the Hardcoded Class FyersData from the restxapi and renamed the class to BrokerData. Can you please refer and change the Class Name. I will test and approve the pr accordingly.

broker/icici/api/data.py
```diff
@@ -0,0 +1,165 @@
...
+        'Authorization': f'Bearer {AUTH_TOKEN}',
+        'Content-Type': 'application/json'
+    }
+
+    conn.request(method, endpoint, payload, headers)
+    res = conn.getresponse()
+    data = res.read()
+    return json.loads(data.decode("utf-8"))
+
+class ICICIDirectData:
```
crypt0inf0: Rename `ICICIDirectData` to `BrokerData` for common class name.

restx_api/history.py
```diff
@@ -56,7 +56,8 @@ def post(self):
 
             try:
                 # Initialize broker's data handler
-                data_handler = broker_module.FyersData(AUTH_TOKEN)
+                # data_handler = broker_module.FyersData(AUTH_TOKEN)
+                data_handler = broker_module.ICICIDirectData(AUTH_TOKEN)
```
crypt0inf0: Rename `ICICIDirectData` to `BrokerData`

PR #157, opened 2024-12-31
fixing log write once the order is sent

marketcalls:
fixing log write once the order is sent for placesmartorder

### Update - Fixed in Latest Release

I've published a new release that addresses the logging timestamp issue. Here are the key changes:

1. **Fixed Order Logging Sequence**
   - Order logs are now written immediately after order placement
   - Timestamps in logs now accurately reflect when orders were sent to the broker
   - The configured delay (SMART_ORDER_DELAY) no longer affects the logged timestamp

2. **Clarification on Delays**
   The smart order process has two types of delays:
   - Position check delay (before order placement) - Required API call to check current positions
   - Configured SMART_ORDER_DELAY (after order placement) - Configurable rate limiting

3. **More Accurate Timestamps**
   - Previous: Timestamps included both position check and configured delays
   - Now: Timestamps reflect actual order placement time

This change gives you more accurate data for order timing analysis while maintaining the rate limiting protection. The SMART_ORDER_DELAY still applies before sending the response, but it no longer affects the logged timestamps.

Please update to the latest version to get these improvements.

### DIFF
--- a/restx_api/place_smart_order.py
+++ b/restx_api/place_smart_order.py
@@ -208,6 +208,17 @@ class PlaceSmartOrder(Resource):
                         'message': ' Positions Already Matched. No Action needed.'
                     })
                     return make_response(jsonify(order_response_data), 200)
+
+                # Log successful order immediately after placement
+                if res and res.status == 200:
+                    order_response_data = {'status': 'success', 'orderid': order_id}
+                    executor.submit(async_log_order, 'placesmartorder', order_request_data, order_response_data)
+                    socketio.emit('order_event', {
+                        'symbol': order_data.get('symbol'),
+                        'action': order_data.get('action'),
+                        'orderid': order_id,
+                        'mode': 'live'
+                    })
                 
             except Exception as e:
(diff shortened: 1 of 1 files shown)


raghava4u:
Addressed your #157 review on `c722ccf01` (Mac checkout).

**Per-test:** all four extras are intentional structure+enrich on `/optionchain`, not duplicated work. Test 4’s live-poll +1 is that pair; `getOptionChain(far)` count across visibility refresh is unchanged (#152 guard kept). Expectations updated with `with_quotes` asserts + comments.

**Abort race:** `ignoreNextAbortRef` boolean → `ignoreAbortControllerRef` keyed to the aborted controller, so a settle-wins-race cannot swallow the next 25s timeout. Unit test covers the no-op abort case.

Vitest: StrategyBuilder + useOptionChainPolling + useOptionChainLive = 50 pass. Collab note: `collab/messages/2026-09-20-grok-to-claude-157-review.md`. Ready for re-review — no merge from me.

PR #274, opened 2025-06-07
FlatTrade WSS Implement

crypt0inf0:
This is a rough attempt to implement flattrade wss into openalgo using the free version `claude 3.5`, `gemini 2.5 pro`, `SWE-1` on windsurf.

I thing everything works fine. But Order subscription is incomplete.

Fine tune it with the paid version to match close with the other brokers wss implementation.

It has detailed logging and debug info and module test, Once fine tuned we can remove those thing.

I'm not going to work on it any further. And once it got fine tuned we can simply port to `Shoonya` as both of them use the same backend.


crypt0inf0:
Deleting it completely and try to implement a new flattrade wss which is compatible with openalgo during the weekend.

PR #296, opened 2025-06-27
Run ngrok on specified port

MananGandhi1810:
Currently, ngrok runs on port 80 by default. This change points ngrok to the port where the flask process is running. For example, I run OpenAlgo Flask process on 5001 port, but when I have ngrok tunnelling turned on, the ngrok process points to port 80.

### DIFF
--- a/app.py
+++ b/app.py
@@ -174,7 +174,7 @@ def create_app():
 
 def setup_environment(app):
     with app.app_context():
-        #load broker plugins
+        # load broker plugins
         app.broker_auth_functions = load_broker_auth_functions()
         # Ensure all the tables exist
         ensure_auth_tables_exists()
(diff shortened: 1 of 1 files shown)


marketcalls:
This config will impact our Ubuntu Server Side Scripts as some deploys at port 80 and port 443 based on whether SSL is enabled or not.

MananGandhi1810:
Oh, but then how can we allow the user to configure for a specific port? A new env var called NGROK_PORT?

marketcalls:
thats a best idea. let me think

marketcalls:
We are recommending users to use devtunnel for any testing and which is simpler than ngrok.
And for time being we are not going to do any updates to ngrok setups. The existing configs
will be functional. and going forward after static ip implementation the ngrok will become less relevant

PR #304, opened 2025-06-30
added icici-breeze

vijaywnirmal:


marketcalls:
Hi Vijay thanks for your contribution to openalgo. Unfortunately ICICI Direct asked all the algo platforms including openalgo to remove the configurations and setups. We had already built ICICI Direct with the support of ICICI Direct team and it is available till 1.0.23 version. But the later versions we have to remove it from openalgo based on their request.

PR #318, opened 2025-07-11
update

ayush20180:


crypt0inf0:
We can add any stocks we want without openalgo standard symbol,
image
It will cause error during placing orders. Can we add search script functionality on `Stock Configuration`

marketcalls:
May i know what this update is all about?

PR #328, opened 2025-07-16
Fixed issue with displaying Issues in analyzer tab due to single quotes…

p2c2e:
I noticed when testing with backdated option symbols, where the symbols have expired that there is a message like "Invalid symbol 'NIFTY07JUL2527000PE' for exchange 'NFO'"

But when setting it to the 'data-response' field in the button, the single quote interferes when we just use JSON.stringify. Change encodes the content and then sets it. Tested locally and it works.


p2c2e:
Incomplete change. Will resubmit.

PR #347, opened 2025-07-22
Attempt to make non-market order request to zerodha to be tried as AMO

p2c2e:
Attempts to make non-market smartorder request to zerodha to be tried as AMO variety. 

Not sure this is the right thing to do - caller may want to explicitly request for AMO. But, since SmartOrder DAO is not currently extensible (may affect other brokers), did not touch that.

### DIFF
--- a/broker/zerodha/api/order_api.py
+++ b/broker/zerodha/api/order_api.py
@@ -2,11 +2,14 @@ import http.client
 import json
 import os
 import urllib.parse
+from datetime import datetime, time
+import pytz
 from database.auth_db import get_auth_token
 from database.token_db import get_br_symbol, get_oa_symbol
 from broker.zerodha.mapping.transform_data import transform_data, map_product_type, reverse_map_product_type, transform_modify_order_data
 from utils.httpx_client import get_httpx_client
 from utils.logging import get_logger
+from broker.zerodha.api.data import BrokerData
 
 logger = get_logger(__name__)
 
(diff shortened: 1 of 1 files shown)


marketcalls:
I dont want to hardcode the exchange timings. Because at time we will have unplanned sessions, special sessions on saturndays or even Sunday. and sometimes special sessions like muhurat trading days in the evenings. 

And there are instances where exchange conducted sessions after 3.30 due to some outage issues

p2c2e:
Makses sense. 

Wondering why the BrokerData with market_timings was already there - hence was tempted to try it out (it helps with testing BTW - since analyzer cannot catch all the combinations / issues that the real backend catch.)

Would the right approach be to modify the SmartOrderSchema to get addl attributes (beyond few mandatory ones)? That way, we can pass the Variety attribute used by kite

p2c2e:
Side note: Validation of request/response etc is much easier in Pydantic (no validation specific code) than Marshmallow - not sure what the reason for choosing was ?? !

marketcalls:
> Makses sense.
> 
> Wondering why the BrokerData with market_timings was already there - hence was tempted to try it out (it helps with testing BTW - since analyzer cannot catch all the combinations / issues that the real backend catch.)
> 
> Would the right approach be to modify the SmartOrderSchema to get addl attributes (beyond few mandatory ones)? That way, we can pass the Variety attribute used by kite

Everytime we tend to use AI tools it purposefully add market timings. Later instructed Agentic Coding tools not to include any market timings as practically this could be big issue if markets extend beyond that fixed timings.

p2c2e:
Will find another way to support AMO from caller side

PR #351, opened 2025-07-25
Handle cases where average price of holding is zero

p2c2e:
In Zerodha, there are many instances where holdings return avg price of zero. For e.g. when there are corporate bonds, G-Secs etc. In such cases, it takes upto T+2 days for them to reflect. 

Earlier logic was resulting in dividebyzero due to this. Logic handles this situation bit more gracefully.

This maybe happening in other brokers as well. Haven't checked.

### DIFF
--- a/broker/zerodha/mapping/order_data.py
+++ b/broker/zerodha/mapping/order_data.py
@@ -207,14 +207,23 @@ def transform_positions_data(positions_data):
 
 def transform_holdings_data(holdings_data):
     transformed_data = []
-    for holdings in holdings_data:
+    for holdings in holdings_data:  
+        # Handle zero average price case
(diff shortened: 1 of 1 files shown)


p2c2e:
Bumping this up. I understand this is not going to be an issue with folks who only do Equity, F&O etc. BUT, folks like me who have Corporate Bonds, G-Secs have multiple days when the code fails. 

Merging this will allow me to reduce patches in my own fork and help a few others.. I am attaching two examples where issues happen as reference:

First is a Corporate Bond that is in the Demat:
image
and this one is a TBill that was having the same issue for few days (until the auction is completed and becomes available on Holdings)
image

Thanks in advance!

marketcalls:
Will test this on Monday and will update it.

PR #369, opened 2025-07-30
Update the ltp wss test with csv file

crypt0inf0:
## Summary by cubic
Added a CSV file with 978 NSE symbols and updated the test script to load symbols from this file using a corrected file path.


marketcalls:
Can you bring the exchange column as well instead of hardcoding it?

marketcalls:
some how the CSV file is missing in the list

PR #418, opened 2025-08-30
Fix logging error on debug mode

crypt0inf0:
## Summary by cubic
Fixes a debug-mode logging crash caused by the sensitive-data filter converting all args to strings. We now only scrub string args, so %-style formatting (e.g., %d) no longer raises TypeError.

- **Bug Fixes**
  - Preserve original types for non-string log args; apply redaction only to strings.
  - Added a note recommending %s for string formatting to avoid type issues.


crypt0inf0:
logging errors are fixed.

PR #453, opened 2025-09-10
Fix Date Type Error

crypt0inf0:
## Summary by cubic
Fixes date handling in get_history so start_date/end_date can be strings or date/datetime objects. Prevents crashes during timestamp conversion and keeps behavior unchanged for string inputs.

- **Bug Fixes**
  - Normalize start/end dates to "YYYY-MM-DD" when given as date/datetime; then convert to epoch.
  - Updated docstring and relaxed type hints to reflect supported input types.


crypt0inf0:
It is still not merged.

Issue #561, opened 2025-11-06
Ability to extend brokers functionalities


kalai-de:
Zerodha provides api for  Margin calculation, creating the alert and alert triggered order. 

<img width="1109" height="560" alt="Image" src="https://github.com/user-attachments/assets/85544bdf-7ae5-411c-97be-94af75283dcb" />

<img width="1109" height="560" alt="Image" src="https://github.com/user-attachments/assets/9219f6e6-f55d-4540-8fe6-3eb9765e1be2" />

It would be better we can extend the broker individual broker capabilities and we can executes it through python client or API..

marketcalls:
Margin API is now enabled for selected Brokers

PR #562, opened 2025-11-06
feat(broker): add custom broker API endpoint and refactor zerodha API calls

kalai-de:
- Introduce new RESTX endpoint for custom broker API requests
- Extract common API response logic into shared module for zerodha
- Refactor funds and order APIs to use shared API response module

https://github.com/marketcalls/openalgo/issues/561, We can extend broker functionalities directly with this new endpoint, 

image

image

image

we can extend individual brokers cabality depending on broker that we are using.

---
## Summary by cubic
Adds a custom broker RESTX endpoint and centralizes Zerodha API request handling to cut duplication and simplify maintenance. This makes funds and order APIs cleaner and makes it easier to extend support for more brokers.

- **New Features**
  - New POST /broker/custom endpoint to forward broker API requests.
  - Validates input with CustomBrokerSchema and returns consistent error messages.
  - Dynamically routes to broker modules via broker_custom_service.execute_custom.

- **Refactors**
  - Moved shared get_api_response to broker/zerodha/api/api_response.py with pooled httpx client and unified error handling.
  - Updated funds.py and order_api.py to use the shared helper; removed duplicate request code.
  - Registered the new namespace in restx_api/__init__.py.

<sup>Written for commit 3b9bebad4f7e4bc417d342079ad12b460132f122. Summary will update automatically on new commits.</sup>

### DIFF
--- a/broker/zerodha/api/api_response.py
+++ b/broker/zerodha/api/api_response.py
@@ -1,0 +1,75 @@
+from utils.httpx_client import get_httpx_client
+from utils.logging import get_logger
+
+logger = get_logger(__name__)
+
+def get_api_response(endpoint, auth, method="GET", payload=None):
+    """
+    Make an API request to Zerodha's API using shared httpx client with connection pooling.
+    
+    Args:
+        endpoint (str): API endpoint (e.g., '/orders')
+        auth (str): Authentication token
+        method (str): HTTP method (GET, POST, etc.)
+        payload (dict/str, optional): Request payload
+        
+    Returns:
+        dict: API response data
+    """
+    AUTH_TOKEN = auth
+    base_url = 'https://api.kite.trade'
+    
+    # Get the shared httpx client with connection pooling
+    client = get_httpx_client()
+    
+    headers = {
+        'X-Kite-Version': '3',
+        'Authorization': f'token {AUTH_TOKEN}'
+    }
+    
+    url = f"{base_url}{endpoint}"
+    
+    try:
+        # Handle different HTTP methods
+        if method.upper() == 'GET':
+            response = client.get(
+                url,
+                headers=headers
+            )
+        elif method.upper() == 'POST':
+            if isinstance(payload, str):
+                # For form-urlencoded data
+                headers['Content-Type'] = 'application/x-www-form-urlencoded'
+                response = client.post(
+                    url,
+                    headers=headers,
+                    content=payload
+                )
+            else:
+                # For JSON data
+                headers['Content-Type'] = 'application/json'
+                response = client.post(
+                    url,
+                    headers=headers,
+                    json=payload
+                )
+        else:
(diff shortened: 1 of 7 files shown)


kalai-de:
@marketcalls if it is okay for zerodha, I can do the same for other brokers too.

crypt0inf0:
What's the benefit of using the custom broker endpoint?

kalai-de:
Example: Open algo dont have feature to calculate the margin required to execute the option strategies, create the alerts or alert triggered order.

In my case I use zerodha, they have api to 
1. calculate the margin required to execute basket of orders.
2. Alerts and alert triggered order(ATO), I found ATO is really good feature and I use regularly in zerodha 

Similarly Dhan have option to get the option chain, option expiry list, option geeks. 

with my above implementation,using openalgo api I can directly execute my broker's api, without needing to go through extra authentication layer with broker again. 
```
    endpoint = fields.Str(required=True)
    method = fields.Str(required=False, default="GET")
    payload = fields.Raw(required=False, default=None)

```

I can pass the endpoint, method and payload to execute my brokers api call through openalgo api.. 

we can execute brokers' api through openalgo without needing to go through seperate implementation

crypt0inf0:
Was it any different from https://github.com/marketcalls/openalgo/pull/564

kalai-de:
@crypt0inf0  yes, it is different, 
https://github.com/marketcalls/openalgo/pull/564 covers only margins, but still i cannot execute `alert triggered order` offered my broker,   

image

Also dhan offers, historical price for expired options, powerfull option chain with geeks. 

image

image

I can execute all these, without waiting for openalgo to implement, say for example, we need to wait for openalgo to implement https://github.com/marketcalls/openalgo/pull/564 for margin calculation but with this custom broker api execution we can directly use zerodha one without needing to wait for openalgo implementation

crypt0inf0:
I understand your idea and your implementation but most of the openalgo users are non technical user's. They want a simple user experience. 

Can you give a example of how to use this custom API?

crypt0inf0:
I understand your idea and your implementation but most of the openalgo users are non technical user's. They want a simple user experience. 

Can you give a example of how to use this custom API?

kalai-de:
@crypt0inf0 , I agree this is not for non-technical users, but technical people can easily call this, 

```
from openalgo import api

client = api(api_key='api_key', host='http://127.0.0.1:5000')

client._make_request(endpoint='/broker/custom', payload={
    'endpoint': '/alerts',
    "method":"GET",
    "payload":{}
})
```

We can update the python, node sdk to access this easiy without using `_make_request`

crypt0inf0:
@kalai-de Now I understand we can directly use the broker API endpoint via openalgo @marketcalls Can you test this and merge it?

marketcalls:
im exploring it

Issue #848, opened 2026-02-04
Frontend: Implement Shared WebSocket Connection Manager


marketcalls:
## Problem Statement

Currently, each React component that needs real-time market data creates its own WebSocket connection to the backend proxy server (port 8765). This leads to:

1. **Multiple redundant connections** - Opening the Option Chain page with PlaceOrderDialog creates 2+ WebSocket connections
2. **Duplicate subscriptions** - Same symbols may be subscribed multiple times across connections
3. **Resource waste** - Each connection has authentication overhead and memory footprint
4. **Noisy logs** - Multiple "WebSocket client connected" messages in production

### Evidence

When opening Option Chain with PlaceOrderDialog:
```
WebSocket client connected: <session_id_1>
WebSocket client connected: <session_id_2>
```

### Current Architecture

```
OptionChain page
  └── useOptionChainLive → useMarketData → WebSocket #1 (100+ symbols)

PlaceOrderDialog
  └── useLiveQuote → useMarketData → WebSocket #2 (1 symbol)
```

Each `useMarketData` hook instance creates its own WebSocket connection in `frontend/src/hooks/useMarketData.ts` (line 282).

---

## Proposed Solution

Implement a **SharedWebSocketProvider** React context that maintains a single WebSocket connection per user session.

### Architecture

```
┌─────────────────────────────────────────────────────────────┐
│                    Frontend Architecture                     │
├─────────────────────────────────────────────────────────────┤
│                                                             │
│   SharedWebSocketProvider (Context)                         │
│   ├── Single WebSocket connection to backend:8765           │
│   ├── Tracks all subscribed symbols from all components     │
│   ├── Distributes received data to registered callbacks     │
│   └── Exposes: isConnected, subscribe(), unsubscribe()      │
│                                                             │
│   ┌─────────────────┐     ┌─────────────────┐              │
│   │  OptionChain    │     │ PlaceOrderDialog │              │
│   │  (100 symbols)  │     │  (1 symbol)      │              │
│   └────────┬────────┘     └────────┬─────────┘              │
│            │                       │                        │
│            ▼                       ▼                        │
│   ┌─────────────────────────────────────────────────────────┐
│   │  useMarketData / useLiveQuote                           │
│   │  - Subscribes via shared provider                       │
│   │  - Receives data via callback                           │
│   │  - Manages own REST fallback                            │
│   └─────────────────────────────────────────────────────────┘
│                                                             │
└─────────────────────────────────────────────────────────────┘

[...]

PR #854, opened 2026-02-05
Feat : Sharedwebsocket provider Frontend Update

Kalaiviswa:
Feat : Sharedwebsocket provider Frontend Update

---
## Summary by cubic
Introduces a shared WebSocket manager for the React frontend to use a single connection across pages, with ref-counted subscriptions, auto pause/resume on tab visibility, and REST API fallback after repeated failures. Adds auto-reconnect and stronger connection lifecycle controls while keeping existing market data hooks working as-is.

- **New Features**
  - Added MarketDataManager singleton with subscription deduplication and caching.
  - Added MarketDataContext provider and wrapped the app to share the connection.
  - Refactored useMarketData to delegate to the shared manager (API unchanged).
  - Auto pause after 5s when the tab is hidden; auto resume on visibility.
  - Auto-reconnect with exponential backoff; connection guard to avoid duplicate connects.
  - REST API fallback after 3 WebSocket failures, polling multiquotes every 5s; exposed isFallbackMode in hooks and context.
  - Added documentation: frontend/docs/frontend/websocket-manager.md.

- **Bug Fixes**
  - Prevented zombie WebSocket connections by aborting in-flight requests on disconnect.
  - Improved timestamp parsing in market_data_service to handle string and ms values.

<sup>Written for commit 2d8ca896111f5522237ed47f2f927d6c1cb7d163. Summary will update on new commits.</sup>

### DIFF
--- a/frontend/src/contexts/MarketDataContext.tsx
+++ b/frontend/src/contexts/MarketDataContext.tsx
@@ -61,12 +61,14 @@ export function MarketDataProvider({
   const managerRef = useRef<MarketDataManager>(MarketDataManager.getInstance())
   const { isVisible } = usePageVisibility()
 
-  const [connectionState, setConnectionState] = useState<ConnectionState>('disconnected')
-  const [isConnected, setIsConnected] = useState(false)
-  const [isAuthenticated, setIsAuthenticated] = useState(false)
-  const [isPaused, setIsPaused] = useState(false)
-  const [isFallbackMode, setIsFallbackMode] = useState(false)
-  const [error, setError] = useState<string | null>(null)
+  // Initialize state from manager's current state
+  const initialState = managerRef.current.getState()
+  const [connectionState, setConnectionState] = useState<ConnectionState>(initialState.connectionState)
+  const [isConnected, setIsConnected] = useState(initialState.isConnected)
+  const [isAuthenticated, setIsAuthenticated] = useState(initialState.isAuthenticated)
+  const [isPaused, setIsPaused] = useState(initialState.isPaused)
+  const [isFallbackMode, setIsFallbackMode] = useState(initialState.isFallbackMode)
+  const [error, setError] = useState<string | null>(initialState.error)
 
   const pauseTimeoutRef = useRef<ReturnType<typeof setTimeout> | null>(null)
   const wasConnectedRef = useRef(false)
--- a/frontend/src/hooks/useMarketData.ts
+++ b/frontend/src/hooks/useMarketData.ts
@@ -58,7 +58,7 @@ export function useMarketData({
     isConnected: context?.isConnected ?? false,
     isAuthenticated: context?.isAuthenticated ?? false,
     isPaused: context?.isPaused ?? false,
-    isFallbackMode: false,
+    isFallbackMode: context?.isFallbackMode ?? false,
     error: context?.error ?? null,
   })
 
--- a/frontend/src/lib/MarketDataManager.ts
+++ b/frontend/src/lib/MarketDataManager.ts
@@ -807,9 +807,9 @@ export class MarketDataManager {
           }
           this.dataCache.set(dataKey, updatedSymbolData)
 
-          // Fan out to all callbacks for this symbol
+          // Fan out to all callbacks for this symbol (case-insensitive comparison)
           this.subscriptions.forEach((entry) => {
-            if (entry.symbol === symbol && entry.exchange === exchange) {
+            if (entry.symbol.toUpperCase() === symbol && entry.exchange.toUpperCase() === exchange.toUpperCase()) {
               entry.callbacks.forEach((callback) => {
                 callback(updatedSymbolData)
               })


marketcalls:
## PR Review: Shared WebSocket Connection Manager

Thanks for implementing the shared WebSocket architecture as per issue #848! This is a solid implementation. Here's a detailed review:

---

### ✅ What's Good

**Architecture:**
- Singleton pattern for `MarketDataManager` ensures single connection
- Ref-counted subscriptions prevent duplicate WebSocket subscriptions
- Clean separation: Manager (connection) → Context (React state) → Hook (component API)
- Backward compatible - existing hooks API unchanged

**Performance:**
- Single WebSocket connection instead of N connections per component
- Subscription deduplication across modes
- Efficient batch resubscription on reconnect (`resubscribeAll()`)
- Data cache with fan-out to multiple callbacks

**Resilience:**
- Auto-reconnect with exponential backoff (max 30s)
- REST API fallback after 3 consecutive failures
- Tab visibility handling (pause after 5s hidden)
- Abort controller to prevent zombie connections during async operations

---

### ⚠️ Issues to Address

#### 1. **Double Visibility Handling (Bug)**
Both `MarketDataContext.tsx:92-122` AND `useMarketData.ts:98-130` implement visibility change handlers:

```typescript
// MarketDataContext.tsx
useEffect(() => {
  if (!isVisible && (isConnected || isAuthenticated)) {
    pauseTimeoutRef.current = setTimeout(() => {
      manager.pauseConnection()
    }, pauseDelay)
  }
  // ...
}, [isVisible, ...])

// useMarketData.ts
useEffect(() => {
  const handleVisibilityChange = () => {
    if (document.hidden) {
      pauseTimeout = setTimeout(() => {
        manager.pauseConnection()
      }, pauseDelay)
    }
    // ...
  }
  document.addEventListener('visibilitychange', handleVisibilityChange)
  // ...
}, [pauseWhenHidden, pauseDelay, enabled])
```

**Problem:** When used together (which is the intended pattern via `MarketDataProvider`), both handlers fire, potentially causing race conditions or double pause/resume calls.

**Fix:** Remove visibility handling from `useMarketData.ts` since `MarketDataContext` already handles it. The hook should only subscribe to state changes, not manage connection lifecycle.

---

#### 2. **API Key Stored in Memory (Security)**
`MarketDataManager.ts:128`:
```typescript
private apiKey: string | null = null
```

The API key is stored in the singleton for REST fallback. This persists in memory even after the user might expect it to be cleared.

**Recommendation:** 
- Clear `apiKey` on `disconnect()`
- Consider fetching fresh on each fallback poll rather than caching

---

#### 3. **Case Sensitivity Inconsistency (Bug)**
WebSocket path (`handleMessage`) doesn't normalize case:
```typescript
// MarketDataManager.ts:550
const symbol = (data.symbol as string).toUpperCase()
```

But REST fallback does extra normalization:
```typescript
// MarketDataManager.ts:812
if (entry.symbol.toUpperCase() === symbol && entry.exchange.toUpperCase() === exchange.toUpperCase())
```

**Problem:** Symbols could be stored differently in cache depending on data source, causing lookup misses.

**Fix:** Normalize consistently in `subscribe()` method so keys are always uppercase.

---

#### 4. **Fallback Polling Rate (Performance)**
`MarketDataManager.ts:127`:
```typescript
private fallbackPollingRate: number = 5000 // Poll every 5 seconds
```

5-second polling on REST `/api/v1/multiquotes` could be heavy on the server, especially with many symbols.

**Recommendation:** Consider 10-15 second intervals for fallback, or make it configurable via context props.

---

#### 5. **Memory Leak Risk in useMarketData**
`useMarketData.ts:122-130`:
```typescript
document.addEventListener('visibilitychange', handleVisibilityChange)
return () => {
  document.removeEventListener('visibilitychange', handleVisibilityChange)
  if (pauseTimeout) {
    clearTimeout(pauseTimeout)
  }
}
```

The `pauseTimeout` is declared inside `handleVisibilityChange` closure, not in the effect scope. If visibility changes rapidly, timeouts could stack.

**Fix:** Move `pauseTimeout` to a ref:
```typescript
const pauseTimeoutRef = useRef<ReturnType<typeof setTimeout> | null>(null)
```

---

#### 6. **consecutiveFailures Not Reset on Manual Disconnect**
`MarketDataManager.ts:467-493`:
```typescript
disconnect(): void {
  // ...
  this.consecutiveFailures = 0  // ✅ This is good, but...
}
```

Actually, this IS reset - ignore this point. The code is correct.

---

### 📝 Minor Suggestions

1. **Documentation file location**: `frontend/docs/frontend/websocket-manager.md` - consider moving to repo root `docs/` for consistency

2. **TypeScript strict null checks**: Several places use optional chaining that could be simplified with proper type guards

3. **Logging**: The `console.log` statements in production code should use a configurable logger or be removed:
   ```typescript
   console.log('[MarketDataManager] Switching to REST API fallback mode')
   ```

---

### 🧪 Testing Recommendations

Before merging, please verify:
1. Open Option Chain + PlaceOrderDialog → should see only 1 WebSocket connection in Network tab
2. Hide tab for 6+ seconds → connection should pause
3. Unhide tab → should resume and resubscribe
4. Kill WebSocket server → should switch to REST fallback after 3 failures
5. Restart WebSocket server → should auto-reconnect and disable fallback

---

### Summary

| Category | Status |
|----------|--------|
| Performance | ✅ Good - single connection, deduplication |
| Security | ⚠️ Minor - API key caching |
| Bugs | ⚠️ Double visibility handling needs fix |
| Architecture | ✅ Good - clean separation |
| Backward Compat | ✅ Good - API unchanged |

**Recommendation:** Address the double visibility handling bug (#1) before merging. Other issues are minor and can be addressed in follow-up PRs.

Great work on this implementation! 🚀

marketcalls:
## Detailed Explanation: Double Visibility Handling Bug

### The Problem

Currently, **two separate places** handle tab visibility changes:

**1. MarketDataContext.tsx (lines 92-122):**
```typescript
useEffect(() => {
  if (!isVisible && (isConnected || isAuthenticated)) {
    // When tab is hidden → schedule pause
    pauseTimeoutRef.current = setTimeout(() => {
      manager.pauseConnection()  // ← PAUSE #1
    }, pauseDelay)
  } else if (isVisible && isPaused) {
    manager.resumeConnection()  // ← RESUME #1
  }
}, [isVisible, ...])
```

**2. useMarketData.ts (lines 98-130):**
```typescript
useEffect(() => {
  const handleVisibilityChange = () => {
    if (document.hidden) {
      pauseTimeout = setTimeout(() => {
        manager.pauseConnection()  // ← PAUSE #2 (same manager!)
      }, pauseDelay)
    } else {
      manager.resumeConnection()  // ← RESUME #2 (same manager!)
    }
  }
  document.addEventListener('visibilitychange', handleVisibilityChange)
}, [pauseWhenHidden, pauseDelay, enabled])
```

### Example: What Happens When User Switches Tabs

**Scenario:** User has Option Chain page open (uses `useMarketData`), wrapped in `MarketDataProvider`

```
User switches to another tab (t=0)
    │
    ├─► MarketDataContext detects !isVisible
    │   └─► Schedules pauseTimeout #1 (5 seconds)
    │
    └─► useMarketData's visibilitychange listener fires
        └─► Schedules pauseTimeout #2 (5 seconds)

After 5 seconds (t=5s):
    │
    ├─► Timeout #1 fires → manager.pauseConnection()
    │   └─► WebSocket closes, state = 'paused'
    │
    └─► Timeout #2 fires → manager.pauseConnection()
        └─► Called AGAIN on already-paused connection
            (Harmless but wasteful)

User returns to tab (t=10s):
    │
    ├─► MarketDataContext detects isVisible
    │   └─► Calls manager.resumeConnection() ← RESUME #1
    │
    └─► useMarketData's visibilitychange listener fires
        └─► Calls manager.resumeConnection() ← RESUME #2
            └─► Called AGAIN, potentially triggering double connect!
```

### Why This Is a Problem

1. **Race Condition on Resume:**
   ```
   resumeConnection() checks: if (connectionState !== 'paused') return
   ```
   But if RESUME #1 sets state to 'connecting' and RESUME #2 fires immediately after, the guard may not catch it in time, causing:
   - Two `connect()` calls
   - Two WebSocket connections attempted
   - Potential "WebSocket already connecting" errors

2. **Unnecessary Work:**
   - Two timeouts scheduled instead of one
   - Two event listeners consuming memory
   - Two state checks on every visibility change

3. **Inconsistent Behavior:**
   - If `pauseWhenHidden` is `false` in the hook but `true` in context (default), behavior is unpredictable
   - Different `pauseDelay` values between context and hook could cause partial pauses

### The Fix

Remove visibility handling from `useMarketData.ts` completely. The context already handles it globally:

```typescript
// useMarketData.ts - REMOVE this entire useEffect block (lines 98-130):
- useEffect(() => {
-   if (!pauseWhenHidden || !enabled) return
-   const handleVisibilityChange = () => { ... }
-   document.addEventListener('visibilitychange', handleVisibilityChange)
-   return () => { ... }
- }, [pauseWhenHidden, pauseDelay, enabled])
```

### What This Solves

| Before Fix | After Fix |
|------------|-----------|
| 2 visibility listeners per hook instance | 1 global listener in context |
| Race condition on resume | Single, controlled resume |
| Inconsistent pauseDelay handling | One source of truth |
| Memory overhead from duplicate listeners | Minimal footprint |

### Architecture After Fix

```
┌─────────────────────────────────────────────────────┐
│  MarketDataProvider (wraps entire app)              │
│  └─► Handles ALL visibility changes                 │
│      └─► Single pauseConnection() / resumeConnection()
└─────────────────────────────────────────────────────┘
                    │
        ┌───────────┴───────────┐
        ▼                       ▼
┌───────────────┐       ┌───────────────┐
│ useMarketData │       │ useMarketData │
│ (Option Chain)│       │ (PlaceOrder)  │
│               │       │               │
│ Only handles: │       │ Only handles: │
│ - subscribe   │       │ - subscribe   │
│ - data state  │       │ - data state  │
│ - NO visibility│      │ - NO visibility│
└───────────────┘       └───────────────┘
```

This is the proper separation of concerns: **Context manages connection lifecycle, hooks manage subscriptions.**

PR #866, opened 2026-02-06
fix(dhan): use eventlet-compatible rate limiter to prevent 805 errors

aravindgandavadi:
## Problem

The Dhan API rate limiter uses `threading.Lock` which doesn't work correctly with Flask/Gunicorn eventlet workers. When multiple concurrent requests are made (e.g., from strategies using synthetic future calculations), the lock fails to properly synchronize across greenlets, causing rate limit errors (805).

## Root Cause

`threading.Lock` is not compatible with eventlet greenlets. When multiple greenlets try to acquire the lock, they can bypass the synchronization mechanism, leading to:
- Multiple concurrent API calls exceeding Dhan's rate limit
- Error 805 (too many requests) from Dhan API

## Solution

- Use `eventlet.Semaphore` instead of `threading.Lock` when eventlet is available
- Use `eventlet.sleep()` instead of `time.sleep()` for non-blocking delays  
- Fall back to `threading.Lock` for non-eventlet environments (backwards compatible)

## Testing

Tested with 5 concurrent strategies making synthetic future API calls:
- **Before fix**: 12 rate limit (805) errors
- **After fix**: 3 rate limit (805) errors (75% reduction)

Log confirmation:
```
[DHAN_RATE_LIMIT] Using eventlet Semaphore for rate limiting
```

## Reference

[Dhan API Rate Limits](https://dhanhq.co/docs/v2/#rate-limit):
- Quote APIs: 1 request per second
- Data APIs: 5 requests per second

## Changes

- `broker/dhan/api/data.py`: Replace threading.Lock with eventlet.Semaphore

---
## Summary by cubic
Switch the Dhan API rate limiter to an eventlet-compatible implementation so requests serialize correctly under Flask/Gunicorn eventlet workers. This prevents 805 (Too Many Requests) spikes with concurrent strategies.

- **Bug Fixes**
  - Use eventlet.Semaphore and eventlet.sleep when eventlet is available; fall back to threading.Lock and time.sleep otherwise.
  - Log which synchronization path is used for visibility.
  - Maintains 1 req/sec limit; 805 errors dropped ~75% in a 5-concurrency test.

<sup>Written for commit 801b2a80454c4ac2ed9c568dc4e206b92fa4ee5e. Summary will update on new commits.</sup>

### DIFF
--- a/broker/dhan/api/data.py
+++ b/broker/dhan/api/data.py
@@ -1,6 +1,5 @@
 import json
 import os
-import threading
 import time
 import urllib.parse
 from datetime import datetime, timedelta
@@ -8,6 +7,16 @@ from datetime import datetime, timedelta
 import httpx
 import jwt
 import pandas as pd
+
+# Import eventlet for greenlet-compatible synchronization
+try:
+    import eventlet
+    from eventlet.semaphore import Semaphore as EventletSemaphore
+
+    _USE_EVENTLET = True
+except ImportError:
+    _USE_EVENTLET = False
+    import threading
 
 from broker.dhan.api.baseurl import get_url
 from broker.dhan.mapping.transform_data import map_exchange_type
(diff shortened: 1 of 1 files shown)


aravindgandavadi:
Closing this PR - after further testing, the eventlet Semaphore fix doesn't resolve the rate limit issues due to a systemic greenlet thread-switching bug in the codebase. Will create a separate issue to track the root cause.

PR #913, opened 2026-02-12
feat: trending oi tool

Azhagesan-dev:
## Summary by cubic
Adds a new Trending OI tool with a symbol-agnostic timeseries API and a UI for intraday OI/volume/PCR and futures sentiment across selected strikes. Centralizes IST time handling with new utils and vectorized conversions; endpoints require a TradingView API key in /apikey.

- **New Features**
  - Backend: timeseries blueprint with POST /timeseries/api/chain and /timeseries/api/data; strict validation and session/API‑key checks; symbol‑agnostic history service returning aligned columnar data with batched fetches, rate limits, and retries; futures contract lookup; blueprint registered.
  - Frontend: new /trendingoi page with strike selection (range/custom, ATM auto/fixed), interval controls (1m–1h), date picker, PCR and PE‑CE metrics, futures LTP/OI/volume, heatmap and sentiment tags, column visibility toggles; API client added; route and Tools tile wired.

- **Bug Fixes**
  - History API timezone fixes: centralized IST helpers (get_ist_now, to_ist_epoch, to_ist_epoch_series); vectorized series conversion; added tests; corrected end‑of‑day handling.
  - Numeric handling: to_ist_epoch_series now preserves numeric epoch inputs (no double conversion); test added.
  - React state mutation fix to prevent stale selections and mis‑sorted rows.
  - Hardening: added per‑endpoint rate limiting (10/min), capped /timeseries/api/data to 100 symbols, sanitized symbols payload, and set the frontend date to IST to avoid off‑by‑one selection.
  - UI: added proper colors for light mode for better readability.

<sup>Written for commit 3627315194a608ad6afbbc19a4a70422c0d61934. Summary will update on new commits.</sup>


marketcalls:
the historify service you are modifying to IST. Will it impact other brokers as well? Any different brokers are providing different style of timestamps formats. Will the change in the file history_service will it break for other brokers?

marketcalls (commented):
## Comprehensive Code Review — PR #913: feat: trending oi tool

Thank you for this contribution, @Azhagesan-dev. This is a well-structured new feature with clean separation between backend (symbol-agnostic) and frontend (classification/aggregation). However, there are several issues across **architecture, security, correctness, and codebase consistency** that need to be addressed before merge.

---

## 1. SECURITY ISSUES (Must Fix)

### 1.1 No limit on `symbols` list size (`blueprints/timeseries.py:153`)
The `/timeseries/api/data` endpoint accepts a `symbols` list with **no upper bound**. An authenticated user can send thousands of symbols, causing the backend to make thousands of sequential broker API calls (each with 500ms delay + retries), effectively DoS-ing the server thread for minutes.

**Required fix:** Add a max symbols limit (e.g., 100) with validation:
```python
MAX_SYMBOLS = 100
if len(symbols) > MAX_SYMBOLS:
    return jsonify({"status": "error", "message": f"Maximum {MAX_SYMBOLS} symbols allowed"}), 400
```

### 1.2 No rate limiting on the new endpoints
Both `/timeseries/api/chain` and `/timeseries/api/data` lack Flask-Limiter rate limiting. Compare with other blueprints — most tool endpoints are rate-limited. The `/data` endpoint is especially expensive (N sequential broker calls).

**Required fix:** Add `@limiter.limit()` decorators consistent with other blueprints.

### 1.3 `@cross_origin()` without restrictions
The `@cross_origin()` decorator on both endpoints allows requests from **any origin**. The project already has CORS configured globally via `cors.init_app(app, **get_cors_config())`. Using `@cross_origin()` per-route **overrides** the global config and opens the endpoints to any domain. Check how other blueprints like `oitracker.py` and `straddle_chart.py` handle this — they also use `@cross_origin()`, so this is consistent with existing code but worth noting as a project-wide concern.

### 1.4 Error message leaks internal details (`straddle_chart.py` pattern vs this PR)
The straddle blueprint returns `str(e)` in error responses, which can leak stack traces. This PR correctly returns generic messages (`"An error occurred processing your request"`) — good practice. However, the `timeseries_service.py` logs `logger.exception()` which includes full tracebacks in logs, which is fine for debugging.

---

## 2. IST TIMEZONE CHANGE — CROSS-BROKER IMPACT (Critical Review)

### 2.1 `history_service.py` change affects ALL brokers
The change at `services/history_service.py:184-186` replaces:
```python
start_timestamp = int(start_dt.timestamp())
end_timestamp = int(end_dt.timestamp())
```
with:
```python
start_timestamp = to_ist_epoch(start_dt)
end_timestamp = to_ist_epoch(end_dt)
```

This change is in `get_history_from_db()` which queries the DuckDB/Historify database — used by **all brokers** when `source="db"`. The `to_ist_epoch()` function treats naive datetimes as IST and localizes them before converting to epoch.

**Impact analysis:**
- `start_dt` and `end_dt` are created from user-supplied `YYYY-MM-DD` strings via `datetime.strptime()` — these are **naive** datetimes (no timezone).
- Previously: `int(start_dt.timestamp())` used the **system's local timezone** (Python default for naive datetimes).
- Now: `to_ist_epoch()` **forces IST** regardless of system timezone.

**This is correct for Indian market data** but creates a **behavioral change**:
- On servers in IST timezone: **No change** (same result).
- On servers in UTC or other timezones (cloud deployments, Docker): **Different epoch values** — a ~5.5 hour shift. This could cause queries to miss data or return extra data near date boundaries.

**Verdict:** This fix is arguably correct for an Indian trading platform, but it **must be documented as a breaking change** for non-IST deployments. Users running OpenAlgo on UTC servers (common in cloud) will see different Historify query results.

### 2.2 `broker/angel/api/data.py` changes — Angel-only, safer
The changes to Angel's `get_history()` and `get_oi_history()` replace:
```python
current_time = pd.Timestamp.now()
# and
df["timestamp"] = df["timestamp"].astype("int64") // 10**9
```
with:
```python
current_time = get_ist_now()
# and
df["timestamp"] = df["timestamp"].apply(to_ist_epoch)
```

**Impact:** Angel-only. The `get_ist_now()` change is correct — market hours should be IST regardless of server TZ. The `to_ist_epoch` conversion is semantically better but:
- `.apply(to_ist_epoch)` is **significantly slower** than vectorized `.astype("int64") // 10**9` for large DataFrames. For a symbol with 375 1-minute candles this is fine, but worth noting.
- The old code (`astype("int64") // 10**9`) treats timestamps as UTC nanoseconds. If Angel returns IST timestamps (which they appear to), the old code was **already wrong** (off by 5.5 hours) for intraday data but happened to work because both sides had the same bug. The new code is technically more correct.

### 2.3 Other brokers NOT updated — inconsistency
**16 broker modules** use `.astype("int64") // 10**9` and **6 broker modules** use `pd.Timestamp.now()`. Only Angel was updated. This creates an inconsistency — the same timestamp handling differs across brokers. Either:
- Update all brokers (risky, large scope), or
- Document that Angel is the reference and others will follow incrementally.

**Required action:** At minimum, document this as a known inconsistency in the PR description.

---

## 3. CODE QUALITY ISSUES

### 3.1 React state mutation bug (`TrendingOI.tsx:833`)
As noted by cubic's review, `customStrikes` sorting should not be done in-place. Line 833:
```tsx
if (strikeMode === 'custom') return [...customStrikes].sort((a, b) => a - b)
```
This is actually **already correct** — `[...customStrikes]` creates a copy before sorting. The cubic review flagged line 373 which appears to be a different version. Verify the latest commit doesn't have an in-place `.sort()` on React state.

### 3.2 No CSRF exemption for the new blueprint
The `timeseries_bp` POST endpoints are called from the React frontend (same origin). The project uses CSRF protection. Other similar blueprints (oitracker, straddle) work because `@cross_origin()` + session cookies handle this. **Verify** that CSRF tokens are properly sent from the frontend `webClient` — if the `webClient` doesn't include CSRF tokens for POST requests, these endpoints will fail with 400 errors in production.

### 3.3 Missing `pytz` dependency check
The new `utils/datetime_utils.py` imports `pytz`. Verify this is in `pyproject.toml` dependencies. If it's not already a dependency, it needs to be added.

### 3.4 `time.sleep()` in request handler thread
`timeseries_service.py` uses extensive `time.sleep()` calls (500ms per symbol + batch delays + retry backoff). For 50 symbols, a single request could take **25+ seconds** while blocking the Flask worker thread. With a single Gunicorn worker (`-w 1` as recommended), this blocks **all other requests** to the app.

**Required fix:** Either:
- Document the expected response time prominently in the UI
- Add a maximum symbols cap (see Security 1.1)
- Consider async/background processing for large requests

### 3.5 Test file is not pytest-compatible
`test/verify_timezone_fix.py` uses `print()` statements and manual assertions with `if __name__ == "__main__"`. The project uses pytest. Tests should follow pytest conventions:
```python
def test_ist_now():
    now = get_ist_now()
    assert now.tzinfo is not None
    assert str(now.tzinfo) == "Asia/Kolkata"
```

### 3.6 Frontend hardcodes `new Date().toISOString().split('T')[0]` (`TrendingOI.tsx:785`)
This uses the browser's local timezone, not IST. An Indian user in a non-IST timezone (or traveling) would get a different date than expected. Should use the same IST logic as backend:
```tsx
// Use IST date
const getISTDate = () => {
    const now = new Date()
    const ist = new Date(now.getTime() + (5.5 * 60 * 60 * 1000) - (now.getTimezoneOffset() * 60 * 1000))
    return ist.toISOString().split('T')[0]
}
```

---

## 4. ARCHITECTURE & CONSISTENCY

### 4.1 Blueprint naming inconsistency
Existing tools follow the pattern: `blueprints/{tool_name}.py` with routes at `/{tool_name}/api/...`. The new blueprint is named `timeseries.py` but the feature is "Trending OI" and the route is `/trendingoi`. This is confusing — the blueprint should either be named `trending_oi.py` or the routes should be at `/timeseries/api/...` (which they are, but the frontend page is `/trendingoi`). Pick one name.

**Suggestion:** Since the service is designed to be reusable ("timeseries" is generic), keeping `timeseries` for the backend and `trendingoi` for the frontend page is acceptable, but document this intentional split.

### 4.2 `NSE_INDEX_SYMBOLS` / `BSE_INDEX_SYMBOLS` duplication
`timeseries_service.py` defines its own copy of `NSE_INDEX_SYMBOLS` and `BSE_INDEX_SYMBOLS`. These same sets exist in multiple other services. There should be a single source of truth (likely in `utils/constants.py`).

### 4.3 No auto-refresh / WebSocket integration
All other tools (Option Chain, OI Tracker, Straddle) have either auto-refresh timers or WebSocket integration for live data. Trending OI is purely manual "click Load" — which is fine for v1 but should be noted as a limitation.

### 4.4 `1137 line` single component file
`TrendingOI.tsx` at 1137 lines is very large. Consider breaking out:
- Data processing functions into a separate utility file
- Column definitions into a separate config
- The table component into a sub-component

This isn't blocking but would improve maintainability.

---

## 5. SUMMARY — Required Changes Before Merge

| Priority | Issue | Location |
|----------|-------|----------|
| **P0 - Security** | Add max symbols limit (e.g., 100) | `blueprints/timeseries.py` |
| **P0 - Security** | Add rate limiting to endpoints | `blueprints/timeseries.py` |
| **P1 - Correctness** | Document IST timezone behavioral change for non-IST servers | PR description / CHANGELOG |
| **P1 - Correctness** | Document that only Angel broker is updated (16 others unchanged) | PR description |
| **P1 - Testing** | Convert test to pytest format | `test/verify_timezone_fix.py` |
| **P2 - Quality** | Fix frontend date to use IST | `TrendingOI.tsx:785` |
| **P2 - Quality** | Verify CSRF token handling for POST endpoints | Frontend `webClient` |
| **P2 - Quality** | Verify `pytz` in dependencies | `pyproject.toml` |
| **P3 - Style** | Consider breaking up 1137-line component | `TrendingOI.tsx` |
| **P3 - Style** | Deduplicate `NSE_INDEX_SYMBOLS` | `timeseries_service.py` |

**Overall verdict:** This is a well-designed feature with good separation of concerns. The symbol-agnostic backend is reusable. However, the **security gaps (no symbol limit, no rate limiting)** and **cross-broker IST impact** need to be addressed before merge.

marketcalls:
## Multi-Broker Compatibility Analysis

Following up on the review — here's a detailed breakdown of how this PR's timezone changes interact with all 25+ brokers.

---

### Change #1: `services/history_service.py` (line 184-186) — Affects ALL Brokers

This is the most critical change. `get_history_from_db()` is **shared infrastructure** used by every broker when `source="db"` (Historify/DuckDB queries).

**Before:**
```python
start_timestamp = int(start_dt.timestamp())
end_timestamp = int(end_dt.timestamp())
```
Python's `.timestamp()` on a naive datetime uses the **system's local timezone**. On an IST server, this produces IST-based epochs. On a UTC server (AWS, GCP, Docker), it produces UTC-based epochs.

**After:**
```python
start_timestamp = to_ist_epoch(start_dt)
end_timestamp = to_ist_epoch(end_dt)
```
`to_ist_epoch()` **forces IST** regardless of system timezone, shifting the query window by ~5.5 hours on non-IST servers.

**Impact:**
- IST servers: No behavioral change
- UTC/cloud servers: Historify queries will return different data ranges near date boundaries (a ~5.5 hour shift). This could cause missing data or extra data for users running OpenAlgo on UTC cloud deployments or Docker containers with UTC system time.

**Recommendation:** Revert this change from this PR. If IST normalization in `history_service.py` is needed, it should be a separate dedicated PR with proper testing across all brokers and deployment environments.

---

### Change #2: `broker/angel/api/data.py` — Angel-Only, Safe

These changes are scoped entirely within Angel's broker module:
- `pd.Timestamp.now()` → `get_ist_now()` (correct — market hours are always IST)
- `.astype("int64") // 10**9` → `.apply(to_ist_epoch)` (semantically better, minor perf difference)

No other broker is affected. This is fine to keep.

---

### Change #3: `utils/datetime_utils.py` — New Utility, Safe

New file, only imported by Angel and the new timeseries service. No existing code is affected unless it explicitly imports this module.

---

### Current Broker Timestamp Landscape (Why a Global Fix is Risky)

Each broker handles timestamps differently. A quick audit of all 29 broker `data.py` files shows **at least 7 distinct patterns**:

| Pattern | Brokers | How Timestamps Work |
|---------|---------|-------------------|
| ISO 8601 → `int64//10^9` | Zerodha, Angel, AliceBlue, Samco, Mstock, IBulls, FivePaisa, FivePaisaXTS, Firstock, CompositEdge, JainamXTS, TradeJini, Wisdom | Parse ISO string, vectorized nanosecond conversion |
| Direct epoch (UTC) | Dhan, Dhan Sandbox | API returns epoch integers, adds 19800s (5:30h) for IST |
| Epoch (no conversion) | Fyers | Already in epoch format from API |
| Mixed format handling | Upstox | Handles strings, milliseconds, nanoseconds with fallbacks |
| CSV custom format | DefinedGe | Parses `ddMMyyyyHHmm`, explicit IST→UTC localization |
| Epoch + market open offset | Firstock | Adds 9:15 AM offset for daily candles |
| `pd.Timestamp.now()` for today | Angel, Fyers, Nubra, Samco, Mstock, DefinedGe | Uses pandas (system TZ) to detect "is today" |

Key observations:
- **16 brokers** use `.astype("int64") // 10**9` — vectorized, treats nanoseconds as-is (no TZ awareness)
- **6 brokers** use `pd.Timestamp.now()` for "is today" checks — these would benefit from `get_ist_now()` but each needs individual testing
- Daily candle offsets vary: +5:30h (most), +9:15h (Firstock), none (some)

---

### Recommended Path Forward

1. **For this PR:** Remove the `history_service.py` change (the global one). Keep the Angel-specific changes and the new Trending OI feature — they are safely scoped.

2. **For a future PR (timezone normalization):** A separate dedicated PR that:
   - Updates all 29 broker modules to use `get_ist_now()` instead of `pd.Timestamp.now()`
   - Standardizes `to_ist_epoch()` across all brokers that use `.astype("int64") // 10**9`
   - Tests each broker individually (this is a large effort since each broker has different API response formats)
   - Documents the expected behavior for UTC vs IST server deployments

This keeps the Trending OI feature mergeable while avoiding a risky cross-cutting change to the shared history service.

Azhagesan-dev:
## Key Changes & Addressed Review Comments

### 1. Security (P0)
- **Denial of Service (DoS) Prevention**: Added a strict `MAX_SYMBOLS = 100` limit to the `/timeseries/api/data` endpoint. Requests exceeding this limit will be rejected with a 400 Bad Request.
- **Rate Limiting**: Applied `@limiter.limit("10 per minute")` to both `/timeseries/api/chain` and `/timeseries/api/data` endpoints, consistent with other blueprints.

### 2. Correctness & Timezones (P1)
- **Frontend Date Fix**: Updated `TrendingOI.tsx` to use IST-aware date defaults (`new Date().toLocaleDateString('en-CA', { timeZone: 'Asia/Kolkata' })`) instead of the browser's local timezone.
- **Backend Timestamp Consistency**: Documented the cross-broker timestamp inconsistency. The **Angel** broker module now correctly uses `get_ist_now()` and `to_ist_epoch()`. Other brokers remain on naive implementation for now (to be addressed in follow-up).
- **CORS**: Confirmed usage of `@cross_origin()` matches existing project patterns (`oitracker.py`, `straddle_chart.py`).

### 3. Performance (P2)
- **Vectorized Timestamp Conversion**: Updated `broker/angel/api/data.py` to use `to_ist_epoch_series()` (vectorized) instead of `.apply(to_ist_epoch)`. This addresses the review comment about performance, ensuring efficient processing for large datasets.
- **Rate Limit Safety**: Retained `time.sleep()` in `timeseries_service.py` to strictly adhere to broker rate limits and avoid `429 Too Many Requests` errors (stability prioritized over concurrency).

### 4. Testing & Quality (P1)
- **Unit Tests**: Updated `test/test_datetime_utils.py` to use `pytest` assertions, ensuring that `get_ist_now` returns timezone-aware datetimes and `to_ist_epoch` handles naive inputs correctly.
- **CSRF Protection**: Confirmed that `frontend/src/api/client.ts` automatically attaches the `X-CSRFToken` header for all POST requests, so no backend exemption is needed for `timeseries_bp`.
- **Dependencies**: Verified `pytz` is present in `pyproject.toml`.

## Known Inconsistencies
> [!IMPORTANT]
> **Broker Timestamp Inconsistency**: 
> Currently, only the **Angel** broker module has been updated to use the new IST-aware timestamp logic. 
> - **Angel**: Uses `to_ist_epoch()` (Correct)
> - **Other Brokers (16 modules)**: Still use `pd.Timestamp.now()` / `.astype("int64") // 10**9` (Potentially incorrect on non-IST servers).
> 
> This inconsistency is known and accepted for this PR to keep the scope focused. Other brokers will be updated in follow-up PRs.

marketcalls (commented):
Thanks for the thorough work on this, @Azhagesan-dev — the symbol-agnostic service design and IST centralization are well thought out. A few things to address before merging:

---

## Architecture (Positive)

- Clean separation: backend is symbol-agnostic, frontend owns CE/PE/FUT classification
- Reusable `get_multi_symbol_history` service design
- Proper IST centralization in `utils/datetime_utils.py` with good test coverage
- Rate limiting (10/min) and MAX_SYMBOLS (100) caps in place

---

## Performance — Blocking Request Thread (High Priority)

`_fetch_symbol_history_batch` in `timeseries_service.py` fetches symbols **sequentially** with `time.sleep()` delays:

```python
INDIVIDUAL_DELAY = 0.5   # 500ms between each fetch
BATCH_DELAY = 1.0        # 1s between batches of 5
MAX_RETRIES = 2           # up to 3s retry delay
```

For 100 symbols (the max): ~100 × 0.5s + 19 × 1.0s = **~69 seconds blocking the request thread**. With Gunicorn `-w 1` (required for WebSocket), this blocks **all** other requests for over a minute.

Options to consider:
- Use `concurrent.futures.ThreadPoolExecutor` for parallel fetches (respect broker rate limits with a semaphore)
- Or reduce MAX_SYMBOLS significantly (e.g., 30)
- Or add a progress/streaming mechanism

---

## Security Review

**No critical issues found.** The PR has good input validation:
- Chain endpoint validates `underlying`, `exchange`, `expiry_date` with regex
- Symbol count capped at 100, interval whitelist, date format validation
- Extra fields stripped from symbol payloads before forwarding

**One minor gap:** The `/timeseries/api/data` endpoint sanitizes symbol dicts (strips extra fields) but doesn't validate the `symbol` and `exchange` string *values* with regex like the chain endpoint does. Recommend adding:
```python
if not re.match(r"^[A-Z0-9:_-]+$", str(s["symbol"])):
    return jsonify({"status": "error", "message": "Invalid symbol format"}), 400
```

---

## Duplicate Constants

`timeseries_service.py` defines its own `NSE_INDEX_SYMBOLS` and `BSE_INDEX_SYMBOLS` sets. These are already defined in `services/vol_surface_service.py`, `services/option_chain_service.py`, `services/gex_service.py`, etc. Consider importing from a shared location to keep them in sync.

---

## Angel Broker Changes (`broker/angel/api/data.py`)

The `get_ist_now()` + `tzinfo=None` changes are correct for making naive datetimes for the API payload. Just flag that this modifies a shared broker module — other tools (history API, OI tracker) depend on these functions. Worth verifying existing history/OI endpoints still work after these changes.

---

## Frontend

- `TrendingOI.tsx` at 1142 lines is quite large for a single component. Consider extracting the table rendering, the controls row, and the data processing utilities into separate files for maintainability.
- Good patterns: proper cancellation with `let cancelled = false`, IST date default, `requestIdRef` for stale-request protection.

---

## Summary

The PR is well-structured overall. The main concern is the **blocking sequential fetch** for large symbol lists — that's the one thing I'd want addressed before merging. The rest are minor improvements.

marketcalls:
@Azhagesan-dev Thanks for the solid work on this — the architecture is clean and follows the existing patterns well. A couple of changes needed before we can merge:

## 1. Reduce default history load with a strike count dropdown

The current approach fetches `strike_count: 25` for the chain (which is fine — it uses multiquotes), but then the data endpoint fetches history **per-symbol sequentially** with `time.sleep(0.5s)` between each call. At ±25 strikes, that's 100+ history calls taking ~69 seconds and blocking the single Gunicorn worker.

**Requested change:** Add a dropdown to control how many strikes are fetched for history data:

- **Options:** 10, 20, 30, 40, 50 (total CE+PE symbols, so 10 = ±5 strikes)
- **Default:** 20 (i.e., ±10 strikes = 20 CE+PE + 1 FUT = 21 history calls ≈ ~11s)
- Keep `strike_count: 25` for the chain call (multiquotes is fast), but only send the selected number of strikes to `/timeseries/api/data`
- Update `MAX_SYMBOLS` from 100 to 50 to match the max dropdown value + futures

This gives users control over the speed/coverage tradeoff while keeping the default load reasonable.

## 2. Minor items (can be done now or in a follow-up)

- **Deduplicate `NSE_INDEX_SYMBOLS` / `BSE_INDEX_SYMBOLS`** — These are repeated across vol_surface, option_chain, gex, oi_profile, and now timeseries services. Consider importing from a shared location.
- **Add symbol regex validation on `/timeseries/api/data`** — The chain endpoint validates inputs with regex but the data endpoint only strips whitespace on symbol strings.

Everything else (exchange handling, IST centralization, security, rate limiting) looks good. Let us know if you have questions!

PR #937, opened 2026-02-18
fix: prevent lower-mode WebSocket subscription from overwriting higher-mode

ronniedreams:
## Summary                                                                                                                                                                                       

  When the sandbox `WebSocketExecutionEngine` subscribes a symbol in LTP mode (mode=1) for paper trading fill detection, it overwrites an existing Quote (mode=2) or Depth (mode=3) subscription at
   the broker adapter level. This causes higher-mode subscribers to stop receiving data for that symbol.

  **Root cause:** `ConnectionPool.subscribe()` treats `(symbol, exchange, mode)` as independent keys, but broker adapters (e.g. `zerodha_adapter.py`) use `exchange:symbol` without mode as the    
  dict key — so the last-subscribed mode wins and silently downgrades the existing subscription.

  **Why it's not widely reported:** The React dashboard defaults to LTP mode (`useMarketData` hook), which is the same mode the sandbox engine uses. Same mode = `ConnectionPool` returns "Already 
  subscribed" = no conflict. The bug only triggers when an external client subscribes in Quote or Depth mode.

  ## Changes

  Two targeted changes in `websocket_proxy/connection_manager.py`:

  - **`subscribe()`**: Before subscribing, check if a higher mode (Depth > Quote > LTP) already exists for the same symbol. If so, skip the lower-mode subscription since higher modes include all 
  lower-mode data.
  - **`unsubscribe()`**: Before calling `adapter.unsubscribe()`, check if other mode subscriptions still exist for the same symbol. If they do, only remove the pool tracking entry without calling
   the adapter, preserving the broker-level subscription for remaining modes.

  ## Steps to Reproduce

  1. Start OpenAlgo with any broker (e.g. Zerodha)
  2. Subscribe a symbol in **Quote mode** (mode=2) via the Python SDK (`subscribe_quote()`)
  3. Verify Quote data is flowing
  4. Place a **paper trading** (sandbox) order for the same symbol
  5. **Result:** Quote data stops for that symbol immediately

  ## Workaround

  Set `SANDBOX_ENGINE_TYPE=polling` in `.env` to use the polling-based execution engine which doesn't make WebSocket subscriptions.

  ## Checklist

  - [x] Code adheres to PEP 8 guidelines
  - [x] Functions include docstrings (existing docstrings preserved, inline comments added for new logic)
  - [x] Local testing confirms functionality
  - [x] No breaking changes to existing functionality
  - [x] Only one file modified: `websocket_proxy/connection_manager.py

---
## Summary by cubic
Fixes a bug where LTP subscriptions could downgrade existing Quote/Depth subscriptions because broker adapters ignore mode in their keys. Ensures higher-mode subscriptions keep receiving data by skipping redundant lower-mode subscribes and preserving adapter-level subscriptions on partial unsubs.

- **Bug Fixes**
  - subscribe(): If a higher mode (Quote/Depth) is already active for a symbol, skip the lower-mode (LTP) subscription to avoid downgrading the adapter.
  - unsubscribe(): When other modes remain for a symbol, remove only the pool entry and do not call adapter.unsubscribe(); fully unsubscribe only when no modes are left.

<sup>Written for commit c1ea2514178cc7ebfe6fe17dede8be103739cb13. Summary will update on new commits.</sup>

### DIFF
--- a/websocket_proxy/connection_manager.py
+++ b/websocket_proxy/connection_manager.py
@@ -456,13 +456,30 @@ class ConnectionPool:
         sub_key = (symbol, exchange, mode)
 
         with self.lock:
-            # Check if already subscribed
+            # Check if already subscribed in this exact mode
             if sub_key in self.subscription_map:
                 return {
                     "status": "success",
                     "message": f"Already subscribed to {symbol}.{exchange}",
                     "connection": self.subscription_map[sub_key] + 1,
                 }
+
+            # Mode hierarchy: Depth (3) > Quote (2) > LTP (1)
+            # Higher modes include all data from lower modes, so subscribing
+            # in a lower mode when a higher mode is already active is redundant
+            # and can cause the broker adapter to downgrade the existing subscription.
+            for higher_mode in range(mode + 1, 4):
+                higher_key = (symbol, exchange, higher_mode)
+                if higher_key in self.subscription_map:
+                    self.logger.debug(
+                        f"Skipping {symbol}.{exchange} mode {mode} subscription - "
+                        f"already subscribed in higher mode {higher_mode}"
+                    )
+                    return {
+                        "status": "success",
+                        "message": f"Already subscribed to {symbol}.{exchange} in mode {higher_mode}",
+                        "connection": self.subscription_map[higher_key] + 1,
+                    }
 
             try:
                 # Get adapter with capacity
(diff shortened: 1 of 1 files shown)


Kalaiviswa:
Thanks @ronniedreams for identifying this issue — the mode overwrite problem is real and your analysis   is appreciated.

  After reviewing the PR against our codebase, we found that the current approach has a gap: lower-mode 
  subscriptions are skipped but not tracked, which can cause silent data loss if the higher-mode        
  subscriber leaves first. There's also no downgrade path when the highest mode is removed while lower  
  modes remain.
  
   We're planning to implement the complete solution for this on Monday — covering upgrade, downgrade,   
  and full tracking of all requested modes at the ConnectionPool level.
  
  Thanks for raising this!

PR #950, opened 2026-02-22
fix: replace volatile in-memory queue with SQLite-backed persistent order queue

### LINKED ISSUE #949: Architecture: Python Strategy Manager's in-memory order queue silently loses orders on crash/restart — no persistence, dead-letter queue, or retry logic
## Summary

The Python Strategy Manager (`blueprints/strategy.py`) uses Python's built-in `queue.Queue` for both regular and smart order queues:

```python
# blueprints/strategy.py
regular_order_queue = queue.Queue()   # For placeorder (up to 10/sec)
smart_order_queue   = queue.Queue()   # For placesmartorder (1/sec)
```

These are **ephemeral in-memory structures**. If the OpenAlgo process crashes, is restarted (`uv run app.py`), receives a SIGTERM (e.g. Docker restart, systemd service restart, server reboot), or if the background `process_orders()` thread dies silently — **all queued orders are permanently lost with zero notification to the user**.

## Scenario (Reproducible in Live Trading)

1. Strategy generates a burst of 50 orders (e.g. from a ChartInk scanner or a Python strategy on market open).
2. `process_orders()` processes regular orders at ≤10/sec → queue drains over ~5 seconds.
3. At second 2, the server admin runs `systemctl restart openalgo` or the process OOMs.
4. The remaining ~30 queued orders are **silently dropped**.
5. The strategy believes orders were sent. The broker has no record. The user's position is unintentionally flat.

## Why This Is Dangerous

In algorithmic trading, this results in:
- **Silent position mismatch**: strategy thinks it's long, broker shows flat.
- **No stop-loss coverage**: hedges never got placed.
- **No audit trail**: the lost orders don't appear in the orderbook, tradebook, or logs.
- **No alert**: the user has no way to know orders were lost.

This is categorically different from bugs #908 and #933 — those are *wrong behavior*. This is **invisible behavior**: the system appears to succeed while silently discarding orders.

## Root Cause

```python
# blueprints/strategy.py
regular_order_queue = queue.Queue()  # lives in process memory only
smart_order_queue   = queue.Queue()  # lives in process memory only

def process_orders():
    while True:
        try:
            smart_order = smart_order_queue.get_nowait()
            ...
            time_module.sleep(1)   # ← if process killed here = order lost forever
        except Exception as e:
            time_module.sleep(1)   # error swallowed, thread keeps running but queue may stall
```

Additionally, `ensure_order_processor()` starts a **daemon thread** (`daemon=True`). Daemon threads are killed **instantly** when the main process exits — they do not flush their queue before dying.

## Impact Assessment

| Scenario | Result |
|---|---|
| App restart mid-queue | All remaining queued orders lost |
| `process_orders` thread exception | Thread exits silently; queue freezes indefinitely |
| OOM kill | All queued orders lost |
| 100 signals fired at market open | First ~10 placed, rest silently dropped |

### Option 1: SQLite-backed persistent queue (fits OpenAlgo's existing stack)

Replace `queue.Queue` with a lightweight SQLite table acting as a durable queue:

```python
# On order submission:
INSERT INTO order_queue (payload, status, created_at) VALUES (?, 'pending', NOW())

# On processing:
SELECT id, payload FROM order_queue WHERE status='pending' ORDER BY created_at LIMIT 1

# On success:
UPDATE order_queue SET status='sent', sent_at=NOW() WHERE id=?

# On failure / max retries:
UPDATE order_queue SET status='failed', retry_count=retry_count+1 WHERE id=?
```

OpenAlgo already uses SQLite via SQLAlchemy — **zero new dependencies required**.

### Option 2: SIGTERM handler + queue drain

Register a `signal.signal(signal.SIGTERM, handler)` that flushes the queue to a JSON file on disk and re-loads it on next startup.

### Option 3: At minimum — dead-letter notification

Log every unsent order to the Traffic Monitor / logs with status `DROPPED` on shutdown, so users at least know what was lost.

## Files Affected

- `blueprints/strategy.py` — queue definition, `process_orders()`, `ensure_order_processor()`
- `database/` — new `order_queue_db.py` if Option 1 is chosen
- `services/` — new `order_queue_service.py` for persistence logic

## Why This Matters

OpenAlgo markets itself as **enterprise-grade** with "Advanced Monitoring Tools." Financial-grade systems require **at-least-once delivery guarantees** for order execution. The current implementation provides **at-most-once** (and often zero-times) delivery under real-world conditions (server updates, OOM, Docker restarts). This is a silent reliability gap in a system that handles real money.

---

**Environment:** All versions (not broker-specific — this is a core architectural issue)  
**Difficulty:** Medium-Hard | **Impact:** 🔴 Critical (live trading) | **Type:** Architecture / Reliability

LuckyAnsari22:
…rder queue

The Python Strategy Manager previously used queue.Queue (ephemeral RAM) to buffer orders. Any orders pending at process exit were silently lost with no error, no alert, and no recovery path.

This commit introduces a durable SQLite-backed order queue (database/order_queue_db.py) with full order lifecycle tracking:

  pending  processing  sent
                        failed (dead-letter after MAX_RETRIES)

Key changes:
- database/order_queue_db.py (new): persistent queue with enqueue, fetch, mark_processing, mark_sent, mark_failed, startup recovery, and queue_depth monitoring helpers
- blueprints/strategy.py: process_orders() now reads from SQLite; worker thread changed from daemon=True to daemon=False for graceful shutdown; SIGTERM/SIGINT handlers added to signal clean exit
- app.py: order_queue DB initialised in parallel with all other DBs

On restart, any order stuck in 'processing' state is automatically recovered and retried — providing crash recovery with zero manual intervention. Failed orders after MAX_RETRIES are preserved in a dead-letter state for inspection.

Zero new dependencies: uses existing SQLite/SQLAlchemy stack.
Configurable via ORDER_QUEUE_MAX_RETRIES env var (default: 3).

Fixes #949

 ## Summary

Fixes #949 

Replaces the ephemeral `queue.Queue` in `blueprints/strategy.py` with a **SQLite-backed persistent order queue**, giving the Strategy Manager at-least-once delivery guarantees for all orders.

Before this fix, any orders buffered in RAM at the moment of an app restart, OOM kill, or SIGTERM were silently discarded — no error, no log entry, no recovery.

### `database/order_queue_db.py` *(new file)*

Self-contained SQLite-backed queue with full order lifecycle:

| Function | Purpose |
|---|---|
| `init_db()` | Creates `order_queue` table on first run |
| `enqueue_order(endpoint, payload)` | Persists order with `status='pending'` |
| `fetch_next_pending(endpoint)` | Fetches oldest pending order |
| `mark_processing(id)` | Marks order as in-flight (prevents double-pickup) |
| `mark_sent(id)` | Records successful delivery |
| `mark_failed(id, error)` | Increments retry count; dead-letters after `MAX_RETRIES` |
| `recover_stale_processing_orders()` | On startup: resets stuck `processing` orders to `pending` |
| `get_failed_orders()` | Returns dead-letter orders for UI/logging |
| `queue_depth()` | Per-status counts for monitoring |

### `blueprints/strategy.py` *(modified)*

| Change | Detail |
|---|---|
| Removed `queue.Queue()` | Replaced with `order_queue_db` calls |
| `process_orders()` rewritten | Uses persistent queue; marks each order through its lifecycle |
| `queue_order()` updated | Calls `enqueue_order()` — public signature unchanged |
| Worker thread: `daemon=False` | Allows graceful flush on shutdown (was `daemon=True`) |
| SIGTERM/SIGINT handlers added | `_shutdown_event.set()` lets worker exit cleanly |
| Startup recovery | `recover_stale_processing_orders()` called on first start |

### `app.py` *(modified — 2 lines)*

`ensure_order_queue_tables_exists` added to the parallel DB init list.

## Order Lifecycle

```
enqueue_order()
      │
      ▼
  [pending]  ──fetch_next_pending()──►  [processing]
                                              │
                       ┌──────────────────────┤
                       │                      │
                  API call OK           API call fails
                       │                      │
                       ▼                      ▼
                    [sent]         retry_count < MAX_RETRIES?
                                        │            │
                                       yes           no
                                        │            │
                                        ▼            ▼
                                    [pending]     [failed]
                                    (retry)     (dead-letter)
```

On app **restart**, orders stuck in `[processing]` are automatically reset to `[pending]` — crash recovery with zero manual intervention.

## Backward Compatibility

- `queue_order(endpoint, payload)` signature is **unchanged** — all callers require zero modification
- Zero new dependencies — uses existing SQLite/SQLAlchemy stack
- Configurable via `.env`: `ORDER_QUEUE_MAX_RETRIES` (default: `3`)

## Testing

```bash
# Start OpenAlgo
uv run app.py

# Trigger a strategy webhook that generates several orders

# Kill mid-queue (simulate OOM / force restart)
# Windows: taskkill /F /PID <pid>
# Linux:   kill -9 <pid>

# Restart OpenAlgo — watch the startup logs for:
# "Recovered N order(s) that were in-flight during last shutdown."

# Verify orders were retried and sent:
sqlite3 openalgo.db "SELECT id, status, retry_count FROM order_queue;"
```

## Files Changed

```
database/order_queue_db.py     (new — 200 lines)
blueprints/strategy.py         (modified — core queue logic replaced)
app.py                         (modified — 2 lines added)
```

---
## Summary by cubic
Replaces the in-memory order queues with a SQLite-backed persistent queue to stop losing orders on shutdown and add crash-safe retries. Fixes #949.

- **New Features**
  - Durable queue with lifecycle: pending → processing → sent/failed; retries with dead-letter after ORDER_QUEUE_MAX_RETRIES.
  - Processor reads from SQLite, recovers in-flight orders on restart; rate limits unchanged (1/s smart, 10/s regular).
  - Graceful shutdown via non-daemon worker and SIGTERM/SIGINT; queue_depth and failed-order helpers; DB init added in app.py; no new dependencies.

- **Migration**
  - No caller changes; queue_order(endpoint, payload) is unchanged.
  - Tables auto-create on start; set ORDER_QUEUE_MAX_RETRIES if needed (default 3).

<sup>Written for commit eff325d4d2aed74613541ff6d1eb939afbcfdc87. Summary will update on new commits.</sup>

### DIFF
--- a/app.py
+++ b/app.py
@@ -82,6 +82,7 @@ from database.latency_db import init_latency_db as ensure_latency_tables_exists
 from database.sandbox_db import init_db as ensure_sandbox_tables_exists
 from database.settings_db import init_db as ensure_settings_tables_exists
 from database.strategy_db import init_db as ensure_strategy_tables_exists
+from database.order_queue_db import init_db as ensure_order_queue_tables_exists
 from database.symbol import init_db as ensure_master_contract_tables_exists
 from database.telegram_db import get_bot_config
 from database.traffic_db import init_logs_db as ensure_traffic_logs_exists
@@ -507,6 +508,7 @@ def setup_environment(app):
             ("Qty Freeze DB", ensure_qty_freeze_tables_exists),
             ("Historify DB", ensure_historify_tables_exists),
             ("Flow DB", ensure_flow_tables_exists),
+            ("Order Queue DB", ensure_order_queue_tables_exists),
         ]
 
         db_init_start = time.time()
--- a/blueprints/strategy.py
+++ b/blueprints/strategy.py
@@ -1,13 +1,25 @@
 import json
 import os
-import queue
 import re
+import signal
 import threading
 import time as time_module
 import uuid
 from collections import deque
 from datetime import datetime, time
 from time import time
+
+from database.order_queue_db import (
+    enqueue_order,
+    fetch_next_pending,
+    get_failed_orders,
+    mark_failed,
+    mark_processing,
+    mark_sent,
+    queue_depth,
+    recover_stale_processing_orders,
+    init_db as init_order_queue_db,
+)
 
 import pytz
 import requests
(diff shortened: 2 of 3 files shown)


LuckyAnsari22:
@marketcalls — This fixes a silent data-loss issue in live trading.
Happy to make any adjustments you'd like before merging.
The fix uses your existing SQLAlchemy stack with zero new dependencies.

Kalaiviswa:
Hi, thanks for the detailed PR and for identifying this issue — you're absolutely right that 
  the in-memory queue.Queue with daemon=True means pending orders could be lost during a       
  graceful restart or deploy.

  We've addressed this in PR #971 with a lighter-weight approach. Since the queue is
  essentially a rate-limiting buffer where orders pass through in milliseconds, we added an    
  atexit handler that drains any remaining orders before process exit, rather than introducing 
  a full SQLite-backed persistent queue.

  Here's what the fix does:
  - Registers an atexit shutdown handler that sends a poison pill and waits up to 30 seconds   
  for the worker to flush remaining orders
  - Keeps a reference to the worker thread so we can join() it on shutdown
  - Only poisons the regular queue so smart orders drain first naturally

  We felt the full persistent queue approach, while well-engineered, could introduce some      
  concerns in practice:

  - SQLite write contention — Every incoming webhook writes to the queue, and the worker thread   continuously reads and updates status. This constant read-write contention on SQLite could  
  lead to database is locked errors under load, especially since OpenAlgo already uses multiple   SQLite databases concurrently.
  - Signal handler conflicts — Registering SIGTERM/SIGINT handlers at blueprint import time can   interfere with Flask/Gunicorn/eventlet's own signal handling, potentially causing unexpected   shutdown behavior in production.
  - Added latency per order — Every order would go through enqueue → mark_processing →
  mark_sent/mark_failed (3 DB writes per order), adding disk I/O overhead to what is currently 
  a sub-millisecond in-memory operation.
  - New database to maintain — Introduces another SQLite database with its own schema,
  initialization path, and potential migration needs, increasing the overall maintenance       
  surface.

  The queue's purpose is rate limiting, not durable message delivery — orders are in-flight for   milliseconds, strategies resend signals on every candle/tick, and failed API calls are      
  already logged for visibility. The simpler atexit approach covers the one realistic scenario 
  (graceful restart) without these trade-offs.

  Really appreciate you taking the time to dig into this and put together such a thorough      
  solution. Thanks for contributing!

LuckyAnsari22:
Thanks for the detailed explanation — understood the trade-offs. I'll study PR #971 and look for other areas to contribute.

PR #968, opened 2026-02-26
fix: add session cleanup to prevent connection leak in traffic_db

Sadhanandhann:
## Summary
- `scoped_session` with `NullPool` in `traffic_db.py` never calls `remove()`, leaking a SQLite connection on every `log_request()` call
- Over hours of operation, leaked connections exhaust OS file descriptors (`Errno 24: Too many open files`), crashing the app with `sqlite3.OperationalError: unable to open database file`
- Added `finally: logs_session.remove()` to all 12 methods that use `logs_session`

## Root Cause
`logs_session` is a `scoped_session` — it binds a session to each thread and reuses it. With `NullPool`, each session opens a fresh connection but the scoped session registry never releases them. On a busy instance, `logs.db` accumulates 200+ open file handles until the OS limit is hit.

## Test plan
- [ ] Run OpenAlgo instance and monitor open file handles on `logs.db` (`lsof -p <pid> | grep logs.db | wc -l`)
- [ ] Verify handle count stays stable (should be 0-1 instead of growing unbounded)
- [ ] Confirm traffic logging, IP ban checks, and security dashboard still function correctly

---
## Summary by cubic
Fixes a SQLite connection leak in traffic_db by removing scoped sessions after each call. Prevents file descriptor growth and avoids “Too many open files” crashes.

- **Bug Fixes**
  - Added logs_session.remove() in finally blocks across 12 methods.
  - Covers logging, IP bans, 404 tracking, and invalid API key tracking.
  - Ensures scoped_session + NullPool connections are cleaned up after each operation.

<sup>Written for commit a2aa3510601d3353a37e3b871cb745b45141e12a. Summary will update on new commits.</sup>

### DIFF
--- a/database/traffic_db.py
+++ b/database/traffic_db.py
@@ -94,6 +94,8 @@ class TrafficLog(LogBase):
             logger.exception(f"Error logging traffic: {str(e)}")
             logs_session.rollback()
             return False
+        finally:
+            logs_session.remove()
 
     @staticmethod
     def get_recent_logs(limit=100):
@@ -103,6 +105,8 @@ class TrafficLog(LogBase):
         except Exception as e:
             logger.exception(f"Error getting recent logs: {str(e)}")
             return []
+        finally:
+            logs_session.remove()
 
     @staticmethod
     def get_stats():
@@ -122,6 +126,8 @@ class TrafficLog(LogBase):
         except Exception as e:
             logger.exception(f"Error getting traffic stats: {str(e)}")
             return {"total_requests": 0, "error_requests": 0, "avg_duration": 0}
+        finally:
+            logs_session.remove()
 
 
 class IPBan(LogBase):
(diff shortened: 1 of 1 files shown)


Kalaiviswa:
Hi, thanks for taking the time to look into this and raise the PR — connection leak
  prevention is definitely an important concern!

  However, after reviewing our codebase, it turns out this is already handled at the
  application level:

  - app.py has a @app.teardown_appcontext handler that calls logs_session.remove() after every 
  Flask request — this is the recommended SQLAlchemy pattern for scoped sessions in web apps   
  - Blueprint-level teardowns in traffic.py and security.py also call logs_session.remove() as 
  an additional safety net
  - traffic_logger.py has an explicit finally: logs_session.remove() in the WSGI middleware    

  So the cleanup is already covered across all request paths.

  One concern with the approach in this PR — adding remove() inside each individual method     
  could actually cause issues. For example, track_404() internally calls IPBan.is_ip_banned(). 
  If is_ip_banned() removes the session in its finally block, the outer method would lose its  
  active session mid-operation, which could lead to unexpected behavior.

  Really appreciate you looking into this though — thanks for contributing!

PR #1054, opened 2026-03-05
feat: Delta candlestick mode2 streaming

tradesbybashab:
Replace v2/ticker with candlestick_1m as the primary channel for
mode 2 (Quote) subscriptions. v2/ticker's LTP field returns
mark_price (a derived/smoothed price for margin calculations),
not the actual last traded price — causing divergence from
TradingView and incorrect data for LTP-dependent consumers.

candlestick_1m provides server-authoritative OHLCV with the close
field reflecting the actual last trade price, updated every ~500ms.

Design decisions:
- candlestick_1m chosen over all_trades to avoid 100-500+ msgs/sec
  bandwidth on volatile instruments (BTC), and to get server-side
  OHLCV without client-side candle building
- Companion v2/ticker channel runs silently alongside candlestick_1m
  to provide OI, bid/ask, and daily open (fields candlestick_1m
  lacks). Reference counting ensures cleanup when multiple symbols
  share the companion channel
- Idempotency guard prevents ref count over-increment on repeated
  subscribe() calls (e.g. reconnection replays)

Files changed:
- delta_adapter.py: companion channel logic, normalise_candlestick,
  cache_ticker_companion, subscribe/unsubscribe ref counting
- delta_mapping.py: MODE_CHANNELS, COMPANION_CHANNELS mappings
- delta_websocket.py: subscribe/unsubscribe candlestick_1m methods
- base_adapter.py: minor adapter compatibility enhancement

---
## Summary by cubic
Switch Quote mode to candlestick_1m (true LTP) with a silent v2/ticker companion for OI/bid/ask/daily open. Improves resiliency and scaling, and prevents duplicate orders by retrying only idempotent GET requests on transport errors.

- **New Features**
  - Quote mode streams candlestick_1m; companion v2/ticker caches OI, bid/ask, and daily open (cached only, never published).
  - Candlestick normalization merges cached fields and computes change/change_percent; added channel dispatch helpers; DeltaWebSocket supports subscribe/unsubscribe for candlestick_1m.

- **Bug Fixes**
  - Idempotent companion ref counting prevents over-increment; correct upstream subscribe/unsubscribe and independent companion cleanup when the last reference is removed.
  - Resilience: retry transient socket/transport errors with backoff and HMAC re-sign for GET only; non-GET fail fast to avoid duplicate orders.
  - HTTPX pool increased (200 max connections) to prevent WinError 10035; OS-aware FD thresholds avoid false alerts on Windows.

<sup>Written for commit 7f7a0299785b0619da9b568a4b02c5132a25e4be. Summary will update on new commits.</sup>

### DIFF
--- a/broker/deltaexchange/api/order_api.py
+++ b/broker/deltaexchange/api/order_api.py
@@ -85,8 +85,11 @@ def get_api_response(endpoint, auth, method="GET", payload="", params=None):
                 response = client.request(m, url, headers=headers, content=body)
         except (OSError, ConnectionError, httpx.TransportError) as e:
             # Transient socket / connection errors (e.g. WinError 10035,
-            # ConnectionResetError, httpx pool timeouts).  Retry with backoff.
-            if _attempt < _MAX_RETRIES:
+            # ConnectionResetError, httpx pool timeouts).
+            # ONLY retry GET requests (idempotent). For POST/PUT/DELETE (mutating
+            # operations), a retry after transport error could duplicate the side
+            # effect if the first request succeeded but the response was lost.
+            if method.upper() == "GET" and _attempt < _MAX_RETRIES:
                 wait = (_RETRY_BASE * (2 ** _attempt)) + random.uniform(0.0, 0.5)
                 logger.warning(
                     f"[DeltaExchange] Transient error on {endpoint} "
@@ -104,7 +107,8 @@ def get_api_response(endpoint, auth, method="GET", payload="", params=None):
                     api_secret=api_secret,
                 )
                 continue
-            logger.error(f"[DeltaExchange] Request error after {_MAX_RETRIES} retries: {e}")
+            # For non-GET or retries exhausted, fail immediately
+            logger.error(f"[DeltaExchange] Request error on {method.upper()} {endpoint}: {e}")
             return {"success": False, "error": {"code": "request_error", "message": str(e)}}
         except Exception as e:
             logger.error(f"[DeltaExchange] Request error: {e}")


Kalaiviswa:
Hi @tradesbybashab, thanks for this contribution! Had a few thoughts:

  Could you split this into separate PRs?

  The Greeks scipy fallback and the streaming changes are independent — splitting them would    
  make review easier and we can merge the Greeks change faster.

  On the candlestick_1m change:

  I'd like to understand the motivation better. Is this because v2/ticker returns mark_price    
  instead of the actual last traded price? If so, that makes sense — but could you clarify this 
  in the PR description?

  Also, the companion channel + reference counting adds quite a bit of complexity. Would a      
  simpler approach work — like using all_trades channel for LTP alongside v2/ticker for
  OI/bid/ask? Or is there a specific reason candlestick_1m is preferred?

  cubic-dev also flagged a ref-counting issue worth looking at — repeated subscribes could      
  over-increment the count.

  Happy to discuss further!

tradesbybashab:
Hi @Kalaiviswa, thanks for the thoughtful review! Happy to address each point.

`PR Split — absolutely. I'll split this into two separate PRs:`

PR 1: Pure-scipy Black-76 fallback for option Greeks (option_greeks_service.py only)
PR 2: Delta Exchange streaming changes (delta_adapter.py, delta_mapping.py, delta_websocket.py, base_adapter.py)
Will get those up shortly.

`On v2/ticker and the mark_price issue:`

Yes, exactly — v2/ticker's LTP field maps to mark_price, which is a derived/smoothed price Delta uses for margin calculations and liquidations. It diverges from the actual last traded price, especially during volatile moves. I confirmed this by comparing the WebSocket output against TradingView's live chart for BTCUSD — the mark_price consistently lagged behind or overshot the real traded price. candlestick_1m's close field reflects the actual last trade price. I'll clarify this in the PR description.

`On candlestick_1m vs all_trades:`

I evaluated all_trades as an alternative before settling on candlestick_1m. Here's why I chose the current approach:

Bandwidth: all_trades on BTCUSD generates 100–500+ messages/second during volatile periods (every single fill = one message). With 4 symbols subscribed that's potentially 2000+ msgs/sec through the proxy. candlestick_1m sends ~2 msgs/sec per symbol regardless of volume — the exchange server does the aggregation.

Server-authoritative OHLCV: candlestick_1m delivers O, H, L, C, V in a single message — Delta's canonical candle. With all_trades we'd need a full client-side candle builder (tracking candle boundaries, min/max, volume sums) in the adapter, which is arguably more complex than the companion pattern and more error-prone.

Reconnection resilience: After a WebSocket drop, candlestick_1m immediately sends the current candle state with correct OHLCV — zero data loss. With all_trades, every trade missed during disconnection is permanently lost. The client-built candle would have wrong High/Low/Volume until the next candle boundary, with no way to backfill.

LTP accuracy is sufficient for Quote mode: candlestick_1m's close updates every ~500ms — more than adequate for mode 2 (Quote/OHLCV consumers). Users who need true tick-level LTP use mode 1, which already uses v2/ticker.

Bounded complexity: The companion pattern is ~60 lines of one-time implementation. all_trades would simplify subscribe/unsubscribe but add an entire candle builder with reconnection gap handling — net complexity is similar or higher.

That said, I'm open to revisiting if there are concerns I haven't considered. The companion pattern is well-contained in delta_adapter.py and doesn't affect any other broker's code path.

`On the ref-counting issue:`

Good catch by cubic-dev — the over-increment on repeated subscribe() calls was a valid bug. It's already fixed in commit 3c83ed6d: added an idempotency check (is_new_subscription = corr_id not in self.subscriptions) before modifying the companion ref count.

Let me get the split PRs up!

tradesbybashab:
@Kalaiviswa Split done!

PR #1059, opened 2026-03-06
feat: pure-scipy Black-76 fallback for option Greeks

tradesbybashab:
Add automatic fallback to pure math+scipy implementation when py_vollib is unavailable (e.g. Python 3.13 where numba cannot write JIT cache to site-packages on Windows Store installs).

Implementation:
- Black-76 IV via scipy.optimize.brentq with proper error normalization to match py_vollib error patterns
- Delta, gamma, theta, vega, rho via scipy.stats.norm
- All functions return float() casts for type safety
- Transparent fallback: import py_vollib at module load, use scipy functions if import fails

No behavioral change when py_vollib is available.

---
## Summary by cubic
Add a pure-scipy Black-76 fallback for IV and Greeks when py_vollib isn’t usable, plus better IV error diagnosis so deep-ITM prices return valid Greeks instead of errors. Behavior is unchanged when py_vollib loads.

- **New Features**
  - Auto-detect py_vollib at module load and fallback to scipy if needed.
  - IV via brentq with error messages aligned to py_vollib.
  - Greeks (delta, gamma, theta, vega, rho) via scipy.stats.norm with float casts.
  - check_pyvollib_availability now always succeeds due to the fallback.

- **Bug Fixes**
  - IV fallback now checks residual signs to distinguish below intrinsic vs above theoretical max.
  - Deep-ITM cases return valid Greeks instead of a 500 error; removed broad keyword checks that masked bad data.

<sup>Written for commit 58c30bef077b24c736b165b97cc924cda0922eca. Summary will update on new commits.</sup>

### DIFF
--- a/services/option_greeks_service.py
+++ b/services/option_greeks_service.py
@@ -66,19 +66,39 @@ def _black76_iv(
     price: float, F: float, K: float, r: float, T: float, flag: str
 ) -> float:
     """Implied volatility via Brent's method.  Same arg order as py_vollib.
-    
-    Raises ValueError with 'convergence' keyword on failure to match 
-    py_vollib error handling expectations.
+
+    Raises ValueError whose message lets the caller distinguish:
+      - 'below intrinsic'  → price is below theoretical minimum (deep ITM)
+      - 'exceeds theoretical maximum' → price above max Black-76 value (bad data)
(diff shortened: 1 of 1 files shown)


tradesbybashab:
All issues resolved, ready for final review and merge.

PR #1061, opened 2026-03-06
refactor: centralized API_RATE_LIMIT configuration to reduce duplication across endpoints

### LINKED ISSUE #1020: Centralize API_RATE_LIMIT configuration to reduce duplication across endpoints
## Description

Every API endpoint file in \`restx_api/\` independently reads the same environment variable:

\`\`\`python
API_RATE_LIMIT = os.getenv("API_RATE_LIMIT", "10 per second")
\`\`\`

This is duplicated 40+ times across all endpoint files. If the default needs to change, every file must be updated.

## Affected Files

All files in \`restx_api/\`:
- \`funds.py\` (line 15)
- \`margin.py\` (line 14)
- \`market_holidays.py\` (line 13)
- \`symbol.py\`, \`intervals.py\`, \`depth.py\`, \`ping.py\`, etc.

## What to Do

1. Create a shared config constant (e.g., in \`utils/config.py\` or a new \`restx_api/config.py\`):

\`\`\`python
# utils/config.py or restx_api/config.py
import os

API_RATE_LIMIT = os.getenv("API_RATE_LIMIT", "10 per second")
\`\`\`

2. Update all endpoint files to import from the shared location:

\`\`\`python
# Before (in each file)
API_RATE_LIMIT = os.getenv("API_RATE_LIMIT", "10 per second")

# After (in each file)
from utils.config import API_RATE_LIMIT
\`\`\`

## Acceptance Criteria

- [ ] Single source of truth for API_RATE_LIMIT
- [ ] All endpoint files import from shared location
- [ ] Rate limiting still works correctly
- [ ] No duplicate \`os.getenv\` calls for this variable

## Skills You'll Learn
- DRY principle (Don't Repeat Yourself)
- Python module organization
- Configuration management patterns

vbhavh:
Closes https://github.com/marketcalls/openalgo/issues/1020

Currently each endpoint inside restx_api/ reads the API_RATE_LIMIT
environment variable individually using:

API_RATE_LIMIT = os.getenv("API_RATE_LIMIT", "10 per second")

This results in the same configuration being duplicated across many
endpoint files.

This PR centralizes the configuration by introducing a shared
restx_api/config.py module and updating all endpoints to import
API_RATE_LIMIT from there.

Changes:

Added restx_api/config.py
Defined API_RATE_LIMIT once
Updated all REST endpoint files to import the shared constant
Removed duplicated os.getenv calls
This makes the configuration easier to maintain and ensures a single
source of truth for API rate limiting.

---
## Summary by cubic
Centralized API rate limit in restx_api/config.py and standardized imports across endpoints. Added validation for API_RATE_LIMIT and fixed a margin behavior regression.

- **Refactors**
  - Added restx_api/config.py: reads API_RATE_LIMIT (default "10 per second"), validates format, and falls back to default if invalid.
  - Replaced per-file env reads with a shared import; corrected import paths across endpoints.
  - Removed hard-coded "50 per second" in margin.

- **Migration**
  - To keep the previous margin behavior, set API_RATE_LIMIT="50 per second" in the environment (applies globally).

<sup>Written for commit 35cd77ecf99db70ef9832620625a96c186f2bfb8. Summary will update on new commits.</sup>

### DIFF
--- a/restx_api/analyzer.py
+++ b/restx_api/analyzer.py
@@ -11,8 +11,9 @@ from limiter import limiter
 from restx_api.account_schema import AnalyzerSchema, AnalyzerToggleSchema
 from services.analyzer_service import get_analyzer_status, toggle_analyzer_mode
 from utils.logging import get_logger
+from config import API_RATE_LIMIT
 
-API_RATE_LIMIT = os.getenv("API_RATE_LIMIT", "10 per second")
+
 api = Namespace("analyzer", description="Analyzer Mode API")
 
 # Initialize logger
--- a/restx_api/analyzer.py
+++ b/restx_api/analyzer.py
@@ -11,7 +11,7 @@ from limiter import limiter
 from restx_api.account_schema import AnalyzerSchema, AnalyzerToggleSchema
 from services.analyzer_service import get_analyzer_status, toggle_analyzer_mode
 from utils.logging import get_logger
-from config import API_RATE_LIMIT
+from restx_api.config import API_RATE_LIMIT
 
 
 api = Namespace("analyzer", description="Analyzer Mode API")
--- a/restx_api/basket_order.py
+++ b/restx_api/basket_order.py
@@ -13,8 +13,8 @@ from limiter import limiter
 from restx_api.schemas import BasketOrderSchema
 from services.basket_order_service import emit_analyzer_error, place_basket_order
 from utils.logging import get_logger
+from config import API_RATE_LIMIT
 
-API_RATE_LIMIT = os.getenv("API_RATE_LIMIT", "10 per second")
 api = Namespace("basket_order", description="Basket Order API")
 
 # Initialize logger
--- a/restx_api/basket_order.py
+++ b/restx_api/basket_order.py
@@ -13,7 +13,7 @@ from limiter import limiter
 from restx_api.schemas import BasketOrderSchema
 from services.basket_order_service import emit_analyzer_error, place_basket_order
 from utils.logging import get_logger
-from config import API_RATE_LIMIT
+from restx_api.config import API_RATE_LIMIT
 
 api = Namespace("basket_order", description="Basket Order API")
 
--- a/restx_api/cancel_all_order.py
+++ b/restx_api/cancel_all_order.py
@@ -11,7 +11,7 @@ from limiter import limiter
 from restx_api.schemas import CancelAllOrderSchema
 from services.cancel_all_order_service import cancel_all_orders, emit_analyzer_error
 from utils.logging import get_logger
-from config import API_RATE_LIMIT
+from restx_api.config import API_RATE_LIMIT
 
 api = Namespace("cancel_all_order", description="Cancel All Orders API")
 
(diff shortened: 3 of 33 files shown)


vbhavh:
@marketcalls

marketcalls (changes requested):
Thanks for the effort on centralizing the rate limit config — the intent is good, but there are a few issues that need to be addressed before this can be merged.

## 1. Broken imports — `from config import API_RATE_LIMIT` will crash at runtime

The new config module lives at `restx_api/config.py`, but ~22 of the 33 modified files use the bare import:

```python
from config import API_RATE_LIMIT
```

There is no top-level `config.py` in the project root, so this will raise `ModuleNotFoundError` at runtime for most endpoints (basket_order, cancel_all_order, chart_api, close_position, depth, expiry, history, holdings, instruments, intervals, market_holidays, multi_option_greeks, multiquotes, option_chain, option_symbol, orderbook, quotes, search, split_order, symbol, synthetic_future, ticker).

The correct import should be either:
- `from restx_api.config import API_RATE_LIMIT` (absolute), or
- `from .config import API_RATE_LIMIT` (relative)

The PR currently uses **three different import styles** across the files — please pick one and apply it consistently.

## 2. Silent behavioral regression in `margin.py`

The margin endpoint previously had a **higher** rate limit default:

```python
# Before
API_RATE_LIMIT = os.getenv("API_RATE_LIMIT", "50 per second")

# After (centralized)
API_RATE_LIMIT = os.getenv("API_RATE_LIMIT", "10 per second")
```

This silently reduces the margin endpoint's rate limit from 50 to 10 requests/second for anyone not explicitly setting the env var. This is an unintended behavioral change that could break existing users. If the margin endpoint intentionally had a higher limit, the centralization approach needs to account for per-endpoint overrides.

## 3. No input validation on rate limit string

Minor point: the `os.getenv()` value is passed directly to Flask-Limiter decorators. A malformed env var (e.g., `API_RATE_LIMIT="abc"`) would cause a runtime error on the first request. Not a blocker, but worth considering adding basic validation in `config.py` since it's now the single source of truth.

## Security Review

No security concerns introduced by this PR:
- No new attack surfaces or endpoints
- The rate limit value comes from an env var (trusted source), unchanged from the original pattern
- No user-controlled input is involved in the rate limit configuration
- The only security-relevant note is item #2 above — reducing the margin rate limit is a functional regression, not a security issue

## Summary

The main blockers are the broken imports (#1) and the margin rate limit regression (#2). Please fix those and this will be in good shape.

vbhavh:
hey @marketcalls, everything's fixed now.

marketcalls:
Thanks for the contribution. We're temporarily closing this PR along with parent issue #1020 — we plan to re-raise the issue later with refreshed scope. Your work here will be a useful reference when we reopen. Appreciate your effort.

marketcalls:
Additional context: a lot has changed in the codebase since this was filed — several of the affected files have been refactored, so merging against the current baseline would need significant rework. We'll re-raise a fresh issue with updated scope; you're welcome to resubmit then. Thanks again.

PR #1102, opened 2026-03-11
fix: order quantity sent as float instead of int causing broker rejections

Profesooor:
## Summary

- **Bug:** Order quantity (e.g. `65`) was being deserialized as a float (`65.0`) by the marshmallow schema, causing broker APIs like Zerodha to reject orders with: `"Invalid request (failed to decode 'quantity', got: '65.0' (expected int))"`
- **Root cause:** The `quantity` field in `OrderSchema`, `SmartOrderSchema`, `ModifyOrderSchema`, `BasketOrderItemSchema`, and `SplitOrderSchema` used `fields.Float(...)`, which converts integer input like `"65"` or `65` to `65.0`
- **Fix:** Changed `quantity` from `fields.Float` to `fields.Int` in all 5 affected schemas, and updated validation error messages from "number" to "integer"
- Note: `OptionsOrderSchema` and `OptionsMultiOrderLegSchema` already correctly used `fields.Int` for quantity — this PR brings the remaining schemas in line

## Affected schemas in `restx_api/schemas.py`

- `OrderSchema`
- `SmartOrderSchema`
- `ModifyOrderSchema`
- `BasketOrderItemSchema`
- `SplitOrderSchema`

## Test plan

- [ ] Send a placeorder API request with integer quantity and verify it reaches the broker as an int (not float)
- [ ] Send a smartorder API request and verify quantity is passed as int
- [ ] Send a modify order request and verify quantity is int
- [ ] Send a basket order and verify each leg's quantity is int
- [ ] Send a split order and verify quantity is int
- [ ] Verify that non-integer quantity values (e.g. `65.5`) are properly rejected with a validation error

---
## Summary by cubic
Fixes broker order rejections by sending order quantity as an integer instead of a float. Quantity now uses `fields.Int` across affected schemas, and validation messages reflect "integer".

- **Bug Fixes**
  - Switched `quantity` to `fields.Int` in: `OrderSchema`, `SmartOrderSchema`, `ModifyOrderSchema`, `BasketOrderItemSchema`, `SplitOrderSchema`.
  - Updated validation text from "number" to "integer".
  - Brings these schemas in line with existing options schemas that already use integers.

<sup>Written for commit c78bb1b92526c8558f152025f50a3a6a9824e371. Summary will update on new commits.</sup>

### DIFF
--- a/restx_api/schemas.py
+++ b/restx_api/schemas.py
@@ -9,8 +9,8 @@ class OrderSchema(Schema):
     exchange = fields.Str(required=True, validate=validate.OneOf(VALID_EXCHANGES))
     symbol = fields.Str(required=True)
     action = fields.Str(required=True, validate=validate.OneOf(["BUY", "SELL", "buy", "sell"]))
-    quantity = fields.Float(
-        required=True, validate=validate.Range(min=0, min_inclusive=False, error="Quantity must be a positive number.")
+    quantity = fields.Int(
+        required=True, validate=validate.Range(min=0, min_inclusive=False, error="Quantity must be a positive integer.")
     )
     pricetype = fields.Str(
         missing="MARKET", validate=validate.OneOf(["MARKET", "LIMIT", "SL", "SL-M"])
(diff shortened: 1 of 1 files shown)


Kalaiviswa:
Thanks for pointing out this issue! The float-to-int quantity problem was indeed causing      
  broker rejections (e.g., Zerodha rejecting 1.0 expecting 1).

  However, we can't simply change fields.Float to fields.Int in the schemas because OpenAlgo    
  supports Delta Exchange which requires fractional quantities for crypto spot orders (e.g.,    
  0.0001 BTC, 0.05 SOL). Changing to fields.Int would break crypto trading.

  Instead, we've implemented a @post_load hook that:
  - Non-crypto exchanges (Zerodha, Angel, etc.) → converts quantity to int (e.g., 1.0 → 1)      
  - CRYPTO exchange (Delta Exchange) → keeps quantity as float (e.g., 0.0001 stays 0.0001)      
  - Rejects fractional quantities for non-crypto exchanges with a clear error (e.g., 1.9 is     
  rejected instead of silently truncating to 1)

  Please pull the latest commits and give it a test. Appreciate you flagging this!

PR #1172, opened 2026-03-25
fix: disable gunicorn control socket to prevent Permission denied in Docker

### LINKED ISSUE #1171: Gunicorn 25.x control socket fails with Permission denied in Docker
## Bug Description

After upgrading to Gunicorn 24+/25.x, the new **control socket** feature is enabled by default. It tries to create a Unix socket file (`gunicorn.ctl`) in the working directory (`/app/`).

In Docker, `/app/` is owned by `root` but Gunicorn runs as `appuser` (uid=1000), causing:

```
[ERROR] Control server error: [Errno 13] Permission denied
```

This error repeats every time Gunicorn's arbiter restarts or the control server thread retries (observed 4-8 times per day).

## Root Cause

- Gunicorn 24.0+ added a control socket server (`gunicorn.ctl`) — enabled by default
- Default socket path is relative: `gunicorn.ctl` → resolves to `/app/gunicorn.ctl`
- `/app/` is `drwxr-xr-x root root` — `appuser` cannot create files there
- The `worker-tmp-dir` was already moved to `/tmp/gunicorn_workers` for the same reason, but the control socket wasn't addressed

## Fix

Add `--control-socket-disable` to the gunicorn launch in `start.sh`. OpenAlgo doesn't use Gunicorn's control socket feature (no `gunicornctl` commands are invoked anywhere).

Alternative: `--control-socket /tmp/gunicorn.ctl` if control socket functionality is desired in the future.

## Environment

- Gunicorn: 25.1.0
- Docker: non-root user (`appuser`, uid=1000)
- `/app/` owned by root (standard Docker image pattern)

jagatsingh:
## Summary

Fixes #1171 — Gunicorn 25.x control socket fails with `Permission denied` when running as non-root user in Docker.

## Change

One line added to `start.sh`:

```diff
 exec /app/.venv/bin/gunicorn \
     --worker-class eventlet \
     --workers 1 \
     --bind 0.0.0.0:${APP_PORT} \
     --timeout 300 \
     --graceful-timeout 30 \
     --worker-tmp-dir /tmp/gunicorn_workers \
+    --control-socket-disable \
     --log-level warning \
     app:app
```

## Why

Gunicorn 24+ added a control socket (`gunicorn.ctl`) enabled by default. It tries to create the socket in `/app/`, which is owned by `root`. The process runs as `appuser` → `EACCES`. This produces 4-8 error log lines per day.

The control socket is not used by OpenAlgo — no `gunicornctl` commands are invoked anywhere in the codebase.

## Test

Before: `[ERROR] Control server error: [Errno 13] Permission denied` in logs
After: No control socket errors, gunicorn starts cleanly

Closes #1171

---
## Summary by cubic
Disable `gunicorn` control socket in Docker to stop “Permission denied” errors when running as a non-root user (fixes #1171). Adds `--control-socket-disable` to `start.sh`; no behavior change since we don’t use `gunicornctl`.

- **Bug Fixes**
  - `gunicorn` 24+ enables a control socket by default, which tries to create `gunicorn.ctl` under `/app/` (owned by root), causing EACCES. Disabling it removes noisy error logs and keeps startup clean.

<sup>Written for commit a9bcfc42af91a77d3c8f010a5630e3d778422d77. Summary will update on new commits.</sup>

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -241,5 +241,6 @@ exec /app/.venv/bin/gunicorn \
     --timeout 300 \
     --graceful-timeout 30 \
     --worker-tmp-dir /tmp/gunicorn_workers \
+    --control-socket-disable \
     --log-level warning \
     app:app


jagatsingh:
Closing — `--control-socket-disable` is not recognized by gunicorn 25.1.0. The flag may be from a newer unreleased version or named differently. Will investigate the correct approach.

PR #1225, opened 2026-04-08
Feature/arm64 support

### LINKED ISSUE #1195: [ Feature Request ] Docker images for ARM architecture
There is no images for ARM architecture. offically 

we can build from code bit it adding addition things . I want to run behind reverse proxy

Icamtu:
This PR adds official ARM64 architecture support to the Docker build process.

- Integrated `docker/setup-qemu-action@v3` for multi-platform builds.
- Updated `docker/build-push-action@v6` to include `linux/arm64`.
- Verified Dockerfile compatibility with multi-arch base images.

This resolves the ARM64 support request mentioned in issue #1195.

---
## Summary by cubic
Add official ARM64 Docker build support across CI and a dedicated ARM64 workflow, enabling multi-arch images and automated ARM64 pushes. Resolves #1195.

- **New Features**
  - Added `build-arm.yml` to sync upstream and build/push `linux/arm64` images on schedule or manual trigger.
  
- **Refactors**
  - Updated `ci.yml` to use `docker/setup-qemu-action@v3` and build multi-arch (`linux/amd64`, `linux/arm64`) with `docker/build-push-action@v6`.
  - Removed `security.yml` to prevent upstream workflows from running in the builder.

<sup>Written for commit 7039ec787228d6ec3dacb67666315915ce5d45f5. Summary will update on new commits.</sup>


marketcalls:
Thanks a lot for jumping on this @Icamtu! Genuinely appreciated — especially since you also filed the original issue (#1205) and then put the effort in to follow through with a PR. That's the best kind of contribution.

Closing this because arm64 support actually landed on main today via a different approach in 7a8a5c34f — `ci: build multi-arch Docker images (amd64 + arm64) on native runners`. That commit closes both #1195 and #1205. We basically raced each other to the same goal from opposite directions, which is a good problem to have.

Let me walk through what 7a8a5c34f does differently from this PR and why we went that route, so you can compare and decide if you'd rather switch your fork over to pulling the official multi-arch image from `marketcalls/openalgo:latest` going forward.

## Native runners instead of QEMU

The main architectural difference: instead of `docker/setup-qemu-action@v3` + `platforms: linux/amd64,linux/arm64` on a single amd64 runner, `docker-build` is now a matrix job that runs once per arch on its **native** GitHub runner:

```yaml
strategy:
  matrix:
    include:
      - platform: linux/amd64
        runner: ubuntu-latest
      - platform: linux/arm64
        runner: ubuntu-24.04-arm   # GitHub's native arm64 runner, free for public repos
```

A new `docker-manifest` job then downloads both per-arch digests and runs `docker buildx imagetools create` to assemble the final manifest list.

**Why not QEMU?** The image now installs Chromium for Kaleido/Plotly chart rendering (the Telegram `/chart` command was broken in Docker for exactly this reason — see 3e454a079). Chromium's apt post-install scripts are painfully slow under QEMU emulation — roughly 15–25 min per arm64 build vs ~2 min native. On every merge to main we'd be paying that cost. Native arm64 runners via `ubuntu-24.04-arm` keep wall-clock time flat and eliminate a whole class of cross-compile flakiness.

## Why `load:` can't stay if `platforms:` has two entries

One small technical note on the `ci.yml` change in this PR: removing `load: ${{ github.event_name == 'pull_request' }}` breaks the Kaleido + Chromium smoke test added in 3e454a079, because that step runs `docker run <image>` against the locally-loaded PR image. And even keeping `load: true` alongside `platforms: linux/amd64,linux/arm64` doesn't work — buildx refuses to load a multi-platform image into the single-platform local Docker daemon, it's a fundamental limitation of the container driver.

The native-matrix approach in 7a8a5c34f sidesteps both problems: each matrix leg builds a single-platform image, loads it locally with `load: true`, and runs the smoke test natively on its own arch. So arm64 gets its own native smoke test coverage — arch-specific regressions (missing arm64 libs, wheel compatibility, Chromium launch issues) now surface on PR builds, not in production. This is a bonus we couldn't have gotten with the QEMU approach no matter how it was wired.

## About `security.yml`

I saw the deletion of `.github/workflows/security.yml` in this PR. I understand the reasoning from cubic's auto-summary — "Removed security.yml to prevent upstream workflows from running in the builder" — and it makes total sense in the context of your own fork where you don't want it running. But upstream we want to keep it: it runs Bandit with SARIF upload to GitHub Code Scanning plus weekly pip-audit, and both feed into the Security tab. If you'd like to stop it from running in your fork, you can disable it without deleting the file via **Settings → Actions → General → Workflow permissions → Disable actions for this fork**, or by adding a branch filter to skip your fork's refs. That way upstream keeps its security coverage and your fork stays quiet.

## About `build-arm.yml`

The new scheduled workflow looks like it was authored for your personal fork — it pushes to `icamtu/openalgo-arm64:latest`, uses your fork's Docker Hub secrets, and cleans up upstream workflow files after syncing. Totally reasonable for a fork-side sync-and-build pattern. But upstream now publishes the arm64 image directly to `marketcalls/openalgo:latest` via the matrix in 7a8a5c34f, so you don't need the sync-and-build dance anymore — you can just `docker pull marketcalls/openalgo:latest` on any arm64 host.

If there's a reason you still want to ship a separate `icamtu/openalgo-arm64` image (e.g. you want a different base, different default env, or a different release cadence), that can live in your fork without needing to be merged into upstream.

## How to verify the fix works for you

After the first CI run on main completes (triggered by my commit 7a8a5c34f), both arches should appear in the Docker Hub manifest:

```bash
docker manifest inspect marketcalls/openalgo:latest
```

You should see two entries under `manifests`: one with `"architecture": "amd64"` and one with `"architecture": "arm64"`, both `"os": "linux"`. The third `"architecture": "unknown"` entry is the OCI attestation manifest from buildx and is normal.

Then on any arm64 host (Apple Silicon, Raspberry Pi, AWS Graviton, Oracle Ampere):

```bash
docker pull marketcalls/openalgo:latest
docker image inspect marketcalls/openalgo:latest --format '{{.Architecture}}'
# → arm64
```

No `--platform` flag needed — Docker's manifest list resolution picks the right one for your host automatically. And for your original reverse-proxy use case (mentioned in #1195's context, by chaitanyayeleti), the openalgo image has first-class reverse-proxy support via `HOST_SERVER`, `WEBSOCKET_URL`, and `CORS_ALLOWED_ORIGINS` env vars — so once the arm64 image is available, `install/install-docker.sh` will pull it natively and set up nginx + Let's Encrypt in front of it automatically.

Thanks again for the PR and the original issue — you pushed us to actually fix this, and the arm64 smoke test coverage we got as a bonus is directly because you escalated it. If the arm64 manifest doesn't appear after the next CI run, please reopen #1205 and we'll dig in.

PR #1247, opened 2026-04-13
fix: Inject documented env vars (OPENALGO_API_KEY, STRATEGY_ID, etc.) into strategy subprocess

### LINKED ISSUE #1246: Bug: Strategy subprocess does not receive documented environment variables (OPENALGO_API_KEY, STRATEGY_ID, etc.)
## Bug Description

The `strategies/README.md` documents that the following environment variables are **"automatically set"** for hosted Python strategies:

- `STRATEGY_ID`
- `STRATEGY_NAME`
- `OPENALGO_API_KEY`
- `OPENALGO_HOST`

However, `start_strategy_process()` in `blueprints/python_strategy.py` never actually passes these variables to the subprocess. The `subprocess.Popen()` call inherits the parent environment, but none of these strategy-specific variables exist there.

## Impact

Any strategy that uses `os.getenv("OPENALGO_API_KEY")` — as recommended by the README — crashes immediately on startup with an error like:

```
Error: OPENALGO_APIKEY not set
```

This affects **all** hosted Python strategies that follow the documented approach.

## Steps to Reproduce

1. Upload a Python strategy that reads the API key via `os.getenv("OPENALGO_API_KEY")`
2. Start the strategy via the Python Strategy dashboard
3. Check the strategy log — it crashes immediately because the env var is not set

## Root Cause

In `start_strategy_process()` (around line 458), the subprocess arguments are built but no `env` dict is constructed or passed:

```python
subprocess_args["cwd"] = str(Path.cwd())
# No env injection here — STRATEGY_ID, OPENALGO_API_KEY etc. are never set
# Start the process
cmd = [get_python_executable(), "-u", str(file_path.absolute())]
process = subprocess.Popen(cmd, **subprocess_args)
```

## Expected Behavior

The documented environment variables should be available to strategy subprocesses, as stated in `strategies/README.md`.

## Suggested Fix

After setting `subprocess_args["cwd"]`, build a custom env dict and inject the documented variables:

```python
strategy_env = os.environ.copy()
strategy_env["STRATEGY_ID"] = strategy_id
strategy_env["STRATEGY_NAME"] = config.get("name", strategy_id)
strategy_env.setdefault("OPENALGO_HOST", "http://127.0.0.1:5000")
try:
    from database.auth_db import get_api_key_for_tradingview
    user_id = config.get("user_id")
    if user_id:
        _api_key = get_api_key_for_tradingview(user_id)
        if _api_key:
            strategy_env["OPENALGO_API_KEY"] = _api_key
except Exception as e:
    logger.warning(f"Could not inject API key for strategy {strategy_id}: {e}")
subprocess_args["env"] = strategy_env
```

Note: `OPENALGO_HOST` uses `setdefault` so deployments with a custom host/port (set via environment) are not overridden.

## Environment

- OpenAlgo v2.0.0.2 (also confirmed present in v2.0.0.3 source)
- Docker deployment on Ubuntu
- Strategy file uses `os.getenv("OPENALGO_API_KEY")` as documented

RobinAThomas:
## Summary

Fixes #1246 — Strategy subprocess does not receive documented environment variables.

`strategies/README.md` documents that `STRATEGY_ID`, `STRATEGY_NAME`, `OPENALGO_API_KEY`, and `OPENALGO_HOST` are "automatically set" for hosted Python strategies. However, `start_strategy_process()` in `blueprints/python_strategy.py` never actually injects them into the subprocess environment.

## Changes

In `start_strategy_process()`, after setting `subprocess_args["cwd"]` and before spawning the process:

1. Copy the parent environment (`os.environ.copy()`)
2. Inject `STRATEGY_ID` and `STRATEGY_NAME` from the strategy config
3. Set `OPENALGO_HOST` as a fallback default using `setdefault()` — respects deployment-specific config if already set in the environment
4. Decrypt and inject `OPENALGO_API_KEY` using the existing `get_api_key_for_tradingview()` helper
5. Pass the enriched env dict to `subprocess.Popen()`

## Key Design Decisions

- **`setdefault` for `OPENALGO_HOST`**: Uses `strategy_env.setdefault("OPENALGO_HOST", "http://127.0.0.1:5000")` so that deployments with a custom host/port (set via environment) are not overridden. Localhost is only a fallback default.
- **API key from database**: Uses the existing `get_api_key_for_tradingview(user_id)` function to decrypt the API key, keeping it consistent with how TradingView webhooks resolve it.
- **Wrapped in try/except**: API key injection is wrapped to avoid breaking strategy startup if the database lookup fails.

## Testing

- Verified on OpenAlgo v2.0.0.2 Docker deployment
- Strategy using `os.getenv("OPENALGO_API_KEY")` now starts successfully
- Previously crashed immediately with `Error: OPENALGO_APIKEY not set`

### DIFF
--- a/blueprints/python_strategy.py
+++ b/blueprints/python_strategy.py
@@ -462,7 +462,7 @@ def start_strategy_process(strategy_id):
             strategy_env = os.environ.copy()
             strategy_env["STRATEGY_ID"] = strategy_id
             strategy_env["STRATEGY_NAME"] = config.get("name", strategy_id)
-            strategy_env["OPENALGO_HOST"] = "http://127.0.0.1:5000"
+            strategy_env.setdefault("OPENALGO_HOST", "http://127.0.0.1:5000")
             try:
                 from database.auth_db import get_api_key_for_tradingview
                 user_id = config.get("user_id")
--- a/blueprints/python_strategy.py
+++ b/blueprints/python_strategy.py
@@ -457,6 +457,23 @@ def start_strategy_process(strategy_id):
             subprocess_args["stderr"] = subprocess.STDOUT
             subprocess_args["cwd"] = str(Path.cwd())
 
+            # Inject documented strategy environment variables
+            # (per strategies/README.md: STRATEGY_ID, STRATEGY_NAME, OPENALGO_API_KEY, OPENALGO_HOST)
+            strategy_env = os.environ.copy()
+            strategy_env["STRATEGY_ID"] = strategy_id
+            strategy_env["STRATEGY_NAME"] = config.get("name", strategy_id)
+            strategy_env["OPENALGO_HOST"] = "http://127.0.0.1:5000"
+            try:
+                from database.auth_db import get_api_key_for_tradingview
+                user_id = config.get("user_id")
+                if user_id:
+                    _api_key = get_api_key_for_tradingview(user_id)
+                    if _api_key:
+                        strategy_env["OPENALGO_API_KEY"] = _api_key
+            except Exception as e:
+                logger.warning(f"Could not inject API key for strategy {strategy_id}: {e}")
+            subprocess_args["env"] = strategy_env
+
             # Start the process
             # Use Python unbuffered mode for real-time output
             cmd = [get_python_executable(), "-u", str(file_path.absolute())]


RobinAThomas:
Addressed the review feedback from Cubic AI:

**P2 fix**: Changed `strategy_env["OPENALGO_HOST"] = "http://127.0.0.1:5000"` to `strategy_env.setdefault("OPENALGO_HOST", "http://127.0.0.1:5000")` so that deployments with a custom host/port already set in the environment are not overridden. Localhost `127.0.0.1:5000` is now only a fallback default.

Updated in commit a628f9d.

marketcalls (approved):
## Validation ✅

**Fixes a documented-but-missing feature.** `strategies/README.md:97-100` promises hosted Python strategies receive `STRATEGY_ID`, `STRATEGY_NAME`, `OPENALGO_API_KEY`, and `OPENALGO_HOST` automatically (with sample code at README:65 calling `os.getenv("OPENALGO_API_KEY")`), but `start_strategy_process()` never injected them. Strategies crashed on startup with `OPENALGO_APIKEY not set`.

**Fix is correct and minimal (17/0 at `blueprints/python_strategy.py:457-473`):**
- `os.environ.copy()` preserves parent env (so child still gets PATH, PYTHONPATH, etc.)
- `STRATEGY_ID` / `STRATEGY_NAME` injected from config
- `setdefault("OPENALGO_HOST", …)` respects custom deployments — only fills in localhost if unset
- API key decrypted via the existing `get_api_key_for_tradingview(user_id)` helper — same path TradingView webhooks use, so no duplicated crypto logic
- API key lookup wrapped in try/except so a DB hiccup doesn't prevent the strategy from starting

**Security review (single-user, self-hosted threat model per CLAUDE.md):**
- The subprocess is the user's own uploaded code; passing the user's own API key to it is equivalent to — and safer than — the current workaround of hardcoding it in the script.
- Host-level access already implies full DB read access (`db/openalgo.db` with the pepper-hashed key plus `.env`), so a subprocess env var is not a new exposure surface.
- `utils/logging.py`'s `SensitiveDataFilter` already scrubs API keys from logs.

**CI:** cubic + snyk passing.

**Follow-up nits (non-blocking):**
- Narrow `except Exception` on line 471 to DB-specific exceptions so unrelated bugs surface.
- Consider stripping `APP_KEY` / `API_KEY_PEPPER` from the copied env before passing to the child — user strategies have no need for them and it reduces blast radius if a strategy is compromised.

LGTM.

marketcalls:
## Follow-up: docs + examples alignment (commit 4742ece on main)

Merged alongside this PR. After shipping #1247 I realized two documentation gaps plus a latent bug in the example strategy, so I pushed a follow-up directly to \`main\`:

**Changes (`4742ece`):**

- **`strategies/README.md`** — Split the env var documentation into two clearly labeled sections:
  - *Injected by the platform*: `STRATEGY_ID`, `STRATEGY_NAME`, `OPENALGO_API_KEY`, `OPENALGO_HOST` (what this PR added).
  - *Inherited from `.env`*: `HOST_SERVER`, `WEBSOCKET_URL`, `WEBSOCKET_HOST`, `WEBSOCKET_PORT`, and any other key in `.env`. This works because `utils/env_check.py:234` calls `load_dotenv(override=True)` at startup and this PR's `os.environ.copy()` passes them through to the subprocess.
  - Recommends `HOST_SERVER` / `WEBSOCKET_URL` for new strategies (same names as OpenAlgo's own `.env`), with `OPENALGO_HOST` kept as a documented alias for backward compat.

- **`strategies/examples/simple_ema_strategy.py`** — Fixed a latent bug: the file was reading `OPENALGO_APIKEY` (no underscore), which the platform never injected — not before this PR, not after. Now reads `OPENALGO_API_KEY` (matches what `start_strategy_process` injects) and pulls `host` / `ws_url` from `HOST_SERVER` / `WEBSOCKET_URL` with safe fallbacks.

- **`examples/python/emacrossover_strategy_python.py`** — New WebSocket-driven EMA crossover sample. All config (API key, host, ws URL, symbol, EMA periods, risk, direction, interval) is overridable via env vars so the same file works standalone (`export OPENALGO_API_KEY=...`) and under the `/python` runner without edits.

**Why direct-to-main:** docs + additive examples, no code-path impact on live trading, and the tests/lint would have been no-ops on these paths.

Closes the docs drift for issue #1246.

PR #1256, opened 2026-04-15
fix: clean up logs_session for non-banned IPs in SecurityMiddleware

Sadhanandhann:
## Summary
- `IPBan.is_ip_banned()` in `SecurityMiddleware.__call__()` opens a `logs_session` connection on every request at the WSGI level
- Previously, `logs_session.remove()` was only called for **banned IPs** (line 62)
- For **non-banned IPs** (99.9% of requests), the session was never cleaned up — the comment said "cleanup handled by Flask's teardown" but `teardown_appcontext` doesn't reliably fire for WSGI middleware running outside the Flask app context
- This causes file descriptor leaks on `logs.db` (~1 leaked FD per request, accumulating to 200+ over hours and eventually crashing the server with "unable to open database file")

## Fix
Wrap the `is_ip_banned()` call in `try/finally` so `logs_session.remove()` runs for **all** code paths (banned and non-banned).

## Test plan
- [ ] Verify banned IP path still returns 403
- [ ] Monitor `lsof -p <PID> | grep logs.db` under sustained load — FD count should stay stable instead of growing unbounded
- [ ] Run 1000+ concurrent API requests and confirm no "unable to open database file" errors

---
## Summary by cubic
Fixes a resource leak in `SecurityMiddleware` by ensuring `logs_session.remove()` runs for every request, preventing `logs.db` file-descriptor buildup and related crashes.

- **Bug Fixes**
  - Wrap `IPBan.is_ip_banned()` in try/finally to guarantee session cleanup for both banned and non-banned IPs (runs outside Flask teardown scope).
  - Preserve 403 response behavior for banned IPs; no other behavior changes.

<sup>Written for commit 9e6bd58d2c909140f28cdda29b88e59ed9417400. Summary will update on new commits.</sup>

### DIFF
--- a/utils/security_middleware.py
+++ b/utils/security_middleware.py
@@ -55,12 +55,15 @@ class SecurityMiddleware:
         # Get real client IP (handles proxies)
         client_ip = get_real_ip_from_environ(environ)
 
-        # Check if IP is banned
-        if IPBan.is_ip_banned(client_ip):
-            # Clean up scoped session — this runs at WSGI level, outside Flask
-            # request context, so blueprint/app teardown handlers won't fire.
+        # Check if IP is banned — this opens a logs_session connection.
+        # Must clean up in ALL paths (banned and non-banned) because this
+        # runs at WSGI level, outside Flask's teardown_appcontext scope.
+        try:
+            is_banned = IPBan.is_ip_banned(client_ip)
+        finally:
             logs_session.remove()
 
+        if is_banned:
             # Return 403 Forbidden for banned IPs
             status = "403 Forbidden"
             headers = [("Content-Type", "text/plain")]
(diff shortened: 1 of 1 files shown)


marketcalls (approved):
## Validation ✅

**Change:** Wraps `IPBan.is_ip_banned(client_ip)` in `try/finally` with `logs_session.remove()` so the scoped session is released on every request path (banned and non-banned).

**Why this is correct:**
- `IPBan.is_ip_banned()` runs at the WSGI layer before Flask's app context is pushed, so relying on `teardown_appcontext` to clean up the `logs_session` connection opened here is fragile. Under sustained load this leaked one FD per request against `logs.db`, eventually causing "unable to open database file" crashes.
- The fix matches the existing defensive pattern in `traffic_logger.py` (finally-block `logs_session.remove()`) that is documented in CLAUDE.md's "5 layers of session cleanup."
- If `is_ip_banned` raises, `finally` runs and the exception propagates; `is_banned` is never dereferenced, so no `UnboundLocalError` risk.
- Double-cleanup (middleware + `teardown_appcontext`) is harmless — `scoped_session.remove()` is idempotent.

**CI:** All checks passing (backend-lint/test, frontend-build/lint/test/e2e, security-scan, cubic, snyk).

**Nit (non-blocking):** Initializing `is_banned = False` before the `try` would make control flow explicit, but current code is correct.

LGTM.

PR #1263, opened 2026-04-16
fix: use real OS threads for asyncio event loops under eventlet

Kalaiviswa:
fix: use real OS threads for asyncio event loops under eventlet

---
## Summary by cubic
Run `asyncio` event loops in real OS threads when running under `eventlet`, preventing green-thread conflicts and fixing the “Cannot run the event loop while another loop is running” error under `Gunicorn` + `eventlet`.

- **Bug Fixes**
  - Detect `eventlet` at runtime and use `eventlet.patcher.original("threading")`; otherwise use standard `threading`.
  - Start event loop threads with `_original_threading.Thread` in `services/websocket_client.py` and `websocket_proxy/app_integration.py` to ensure real OS threads.

<sup>Written for commit 99ace9fd50b81dea14227fbc65e3830cc53610c7. Summary will update on new commits.</sup>

### DIFF
--- a/services/websocket_client.py
+++ b/services/websocket_client.py
@@ -6,6 +6,7 @@ This client handles authentication and provides a simple interface for services.
 import asyncio
 import json
 import os
+import sys
 import threading
 import time
 from collections.abc import Callable
@@ -16,6 +17,16 @@ import websockets
 from dotenv import load_dotenv
 
 from utils.logging import get_logger
+
+# Import the original threading module to run the asyncio event loop in a real
+# OS thread, bypassing eventlet's monkey-patching which turns threading.Thread
+# into green threads where asyncio.new_event_loop() cannot work.
+if "eventlet" in sys.modules:
+    import eventlet
+
+    _original_threading = eventlet.patcher.original("threading")
+else:
+    _original_threading = threading
 
 # Initialize logger
 logger = get_logger(__name__)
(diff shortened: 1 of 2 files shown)


marketcalls (approved):
## Review: Approve

**Correctness:** The approach exactly matches the established pattern in `services/telegram_bot_service.py:12-17`, which is documented in `CLAUDE.md:182`:

> Any code that needs async behavior must use eventlet green threads or run async work on a separate real OS thread (see `telegram_bot_service.py:_render_plotly_png` for the pattern).

Both `websocket_client.py` and `app_integration.py` create `threading.Thread` to run `asyncio.new_event_loop()`. Under Gunicorn+eventlet, `threading.Thread` is monkey-patched to green threads where asyncio event loops can't run — causing silent failures of WebSocket market data and the proxy server (port 8765).

**No other call sites need this fix.** All other `asyncio.new_event_loop()` / `asyncio.run()` callers (app.py, blueprints/telegram.py, restx_api/telegram_bot.py, broker data.py files) either already guard against eventlet with `initialize_bot_sync()` or run inline on request threads without spawning a separate `threading.Thread`.

**Minor note:** Naming uses `_original_threading` (leading underscore) vs the existing `original_threading` in telegram_bot_service.py — both are fine, the underscore is arguably better for signaling module-private scope.

Clean, focused, correct fix.

PR #1297, opened 2026-04-24
fix(python_strategy): clean up stale RUNNING_STRATEGIES and surface scheduled-start failures

RobinAThomas:
## Summary

Two related bugs cause scheduled Python strategies to silently fail to start the day after a successful run.

### Bug 1 — `start_strategy_process` does not validate stale in-memory entries

```python
if strategy_id in RUNNING_STRATEGIES:
    return False, "Strategy already running"
```

The `RUNNING_STRATEGIES` dict is only cleaned by `stop_strategy_process()`. Any code path that bypasses that — natural exit, internal squareoff, scheduled stop that did not fire, container hiccup — leaves a stale entry that blocks **every** future start until the worker is restarted.

### Bug 2 — `scheduled_start_strategy` discards the failure

```python
logger.info(f"... all checks passed, starting")
start_strategy_process(strategy_id)   # return value thrown away
```

When Bug 1 fires, there is no log line, no `last_error` on the config, no SSE broadcast. No log file is ever created (`start_strategy_process` bails before opening one), making this very hard to diagnose. From the user side it looks like the scheduler simply did not fire.

## Reproduction

1. Schedule a strategy to start at 09:00 and stop at 16:00.
2. Let the strategy self-terminate before 16:00 (e.g. internal squareoff that calls `sys.exit`) so the child process exits **without** going through `stop_strategy_process()`.
3. Next morning at 09:00 the scheduler ticks — strategy silently does not start, no new log file is created, and the strategy card still shows yesterday's stop time.

## Fix

- **`start_strategy_process`**: when a `RUNNING_STRATEGIES` entry already exists, validate liveness via `Popen.poll()` / `psutil.is_running()` / `psutil.pid_exists()` depending on the stored object type. If the process is dead, log a warning, close the log handle, drop the stale entry, and fall through to start a fresh one. If the process is alive, behaviour is unchanged.
- **`scheduled_start_strategy`**: capture `(success, message)` and on failure log an error, set `last_error` on the config, and broadcast an `error` status update via SSE so the UI reflects reality.

Both changes are surgical (50 + / 2 −) and live entirely inside the existing `PROCESS_LOCK` critical section, so concurrency semantics are unchanged.

## Testing

- Verified hot-patched on a production deployment (v2.0.0.2) that hit this exact bug today; restart was clean and the patch loads without error.
- `python -m py_compile blueprints/python_strategy.py` passes.
- Existing callers of `start_strategy_process` (`stop_strategy_process`, the manual start route at line 1653, the post-login restoration paths at 2589/2654) are unaffected — they all pass through the same gateway and now also benefit from the stale-entry cleanup.

---
## Summary by cubic
Fixes silent no-starts for scheduled Python strategies and prevents stop crashes under `eventlet` workers. Cleans stale `RUNNING_STRATEGIES`, surfaces scheduled-start failures, and switches to eventlet-safe stop logic.

- **Bug Fixes**
  - `start_strategy_process`: validate tracked process (poll/is_running/pid_exists); if dead, warn, close log handle, drop stale entry, then start fresh.
  - `scheduled_start_strategy`: handle `(success, message)`; on failure log an error, set and persist `last_error`, and broadcast an SSE `error`.
  - Stop logic: replace `psutil.wait`/`wait_procs` with `_wait_for_pid_exit` polling (`psutil.pid_exists` + sleep) in `stop_strategy_process` and `terminate_process_cross_platform` to avoid `select.poll` errors under `gunicorn` `eventlet`.

<sup>Written for commit a24a2c3ee767ae286977bbf1d2bba27a56038be6. Summary will update on new commits. <a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1297?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/blueprints/python_strategy.py
+++ b/blueprints/python_strategy.py
@@ -101,6 +101,43 @@ IS_WINDOWS = OS_TYPE == "windows"
 IS_MAC = OS_TYPE == "darwin"
 IS_LINUX = OS_TYPE == "linux"
 
+# Aliased ``time`` module: the top-level ``from datetime import ... time``
+# import shadows the stdlib ``time`` module, so we re-import under a
+# distinct name for sleep/monotonic helpers.
+import time as time_mod  # noqa: E402  (intentional late import for clarity)
+
+
+def _wait_for_pid_exit(pid, timeout=5.0, poll_interval=0.1):
+    """Eventlet-safe wait for a PID to exit.
+
+    ``psutil.Process.wait()`` and ``psutil.wait_procs()`` ultimately call
+    ``select.poll()`` inside ``psutil._psposix.wait_pid_pidfd_open``. Under
+    gunicorn ``--worker-class eventlet`` (OpenAlgo's default), eventlet's
+    monkey-patching strips ``select.poll`` and the call raises
+    ``AttributeError: module 'select' has no attribute 'poll'`` -- which
+    breaks both manual and scheduled strategy stops.
+
+    This helper polls with ``psutil.pid_exists`` (a simple ``os.stat`` call)
+    and ``time.sleep`` (eventlet-aware), and reaps zombie children if we
+    own them. Returns True if the process has exited within the timeout.
+    """
+    deadline = time_mod.monotonic() + max(0.0, timeout)
+    while time_mod.monotonic() < deadline:
+        if not psutil.pid_exists(pid):
+            return True
+        try:
+            p = psutil.Process(pid)
+            if p.status() == psutil.STATUS_ZOMBIE:
+                try:
+                    os.waitpid(pid, os.WNOHANG)
+                except (ChildProcessError, OSError):
+                    pass
+                return True
+        except psutil.NoSuchProcess:
+            return True
+        time_mod.sleep(poll_interval)
+    return not psutil.pid_exists(pid)
+
 
 def init_scheduler():
     """Initialize the APScheduler with IST timezone"""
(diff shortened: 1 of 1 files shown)


marketcalls:
@cubic-dev-ai what is the current status?

marketcalls:
@RobinAThomas — thank you for the careful diagnosis and the surgical fix. The three problems you identified are all real and reproducible:

1. **Stale `RUNNING_STRATEGIES` blocking future starts** — confirmed at `blueprints/python_strategy.py:423-424`. The dict is only ever cleaned by `stop_strategy_process` / `terminate_process_cross_platform`, so any path that bypasses those (natural exit, internal squareoff, OOM, worker hiccup) leaves a permanent zombie entry until the worker restarts.
2. **Silent scheduled-start failure** — confirmed at line 1075 (`start_strategy_process(strategy_id)` with the return value discarded). The manual-stop route at line 1790 already uses the `success, message = ...` pattern correctly; this fix brings the scheduler in line.
3. **Eventlet `select.poll` issue on `psutil.Process.wait()` / `wait_procs`** — real on Linux production (gunicorn `--worker-class eventlet -w 1`). I checked all seven `process.wait(...)` callsites and the three the PR replaces (lines 671, 674, 732) are exactly the ones that hit `psutil` — the other four are `subprocess.Popen.wait()` which is safe under eventlet because it uses `os.waitpid`. Spot-on triage.

We are verifying the PR end-to-end and will test it on Monday on a staging deployment that reproduces the stuck-strategy condition. Will approve and merge once it passes — thank you for your patience.

### One minor cleanup suggestion before merge (optional, but recommended)

OpenAlgo runs cross-platform (Windows / macOS / Linux), and the zombie-reap inside `_wait_for_pid_exit` references `os.WNOHANG`, which is Unix-only:

```python
if p.status() == psutil.STATUS_ZOMBIE:
    try:
        os.waitpid(pid, os.WNOHANG)   # os.WNOHANG raises AttributeError on Windows
    except (ChildProcessError, OSError):
        pass
    return True
```

Today this is fine in practice because Windows processes never enter `STATUS_ZOMBIE`, so the line is unreachable on Windows. But it is a latent footgun if the entry condition ever widens. Could you wrap it in the existing `IS_WINDOWS` guard so it is explicitly cross-platform-safe?

```python
if p.status() == psutil.STATUS_ZOMBIE:
    if not IS_WINDOWS:
        try:
            os.waitpid(pid, os.WNOHANG)
        except (ChildProcessError, OSError):
            pass
    return True
```

`IS_WINDOWS` is already defined near the top of the file, so no new import is needed. Same for `IS_LINUX` / `IS_MAC` if you prefer a more specific guard.

Thanks again for the thorough PR description and the production-deployment validation — exactly the kind of contribution that makes triage fast.

RobinAThomas:
Thanks @marketcalls — pushed the `IS_WINDOWS` guard in [`6789824`](https://github.com/marketcalls/openalgo/pull/1297/commits/67898249). The zombie-reap branch now skips `os.waitpid(..., os.WNOHANG)` on Windows, with a comment noting that the branch is unreachable there today but the guard keeps it safe if the entry condition ever widens.

Ready for your Monday staging run whenever you'd like to proceed.

PR #1298, opened 2026-04-24
Super order 6822451405161363234

radhakrishnan4me-code:
## Summary by cubic
Adds “Super Order” — an OCO-style order with entry, target, stop-loss, and optional trailing SL — with a new REST API, DB storage, and an auto-starting background monitor that trails SL, triggers exits, and cancels on expiry.

- **New Features**
  - `SuperOrder` model and `database/superorder_db` with auto-init, plus an auto-starting monitor that tracks `PENDING/ACTIVE` orders, trails SL, triggers exits, and cancels on expiry.
  - REST API `'/api/v1/superorder'` (place/list/details/modify/cancel), rate-limited; supports `INTRADAY/MIS/NRML/CNC/MTF`; added `'/orders/superorderbook'`.
  - Swagger docs and API Playground requests (`collections/openalgo/**/SuperOrder.bru`); schemas use `product`, `action`, and `price`.

- **Bug Fixes**
  - Forced live order status lookups in the monitor via `{"orderid": ..., "mode": "live"}`; fixed symbol resolution and logging; added startup import checks to fail fast.
  - Fixed monitor crashes by calling `get_order_status` with the correct payload; prevented duplicate WebSocket reconnect threads in `broker/aliceblue/api/alicebluewebsocket.py`; restored `aliceblue_adapter.py` and kept the monitor strictly broker-agnostic (`get_multiquotes`, `get_order_status`, `place_order`, `cancel_order`).
  - Stopped unconditional WebSocket quote unsubscriptions and added in-memory quote cache checks in `broker/aliceblue/api/data.py` to prevent reconnect storms and preserve continuous ticks for monitoring.

<sup>Written for commit 13e2a804801a4caac39aeb88d4c59e52d4f1f367. Summary will update on new commits.</sup>

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,10 +75,16 @@ All brokers share a unified API interface, making it easy to switch between brok
 
 ### Unified REST API Layer (`/api/v1/`)
 A single, standardized API across all brokers with 30+ endpoints:
-- **Order Management**: Place, modify, cancel orders, basket orders, smart orders with position sizing
+- **Order Management**: Place, modify, cancel orders, basket orders, smart orders with position sizing, and **Super Orders** (multi-legged OCO orders with entry, target, trailing stop-loss).
 - **Portfolio**: Get positions, holdings, order book, trade book, funds
 - **Market Data**: Real-time quotes, historical data, market depth (Level 5), symbol search
 - **Advanced**: Option Greeks calculator, margin calculator, synthetic futures, auto-split orders
+
+### Super Order (Multi-Legged OCO)
+- **Single Submission**: Submit entry, target, and stop-loss legs in a single API call.
+- **Auto-Monitoring**: The background monitoring service tracks partial and full fills of the main leg.
+- **Trailing Stop-Loss**: Dynamically trail your stop-loss order by a configurable trailing jump as the market moves in your favor.
+- **OCO Logic**: When either the target or stop-loss condition is met, the other leg is automatically cancelled.
 
 ### Real-Time WebSocket Streaming
 - Unified WebSocket proxy server for all brokers (port 8765)
--- a/app.py
+++ b/app.py
@@ -93,6 +93,7 @@ from database.latency_db import init_latency_db as ensure_latency_tables_exists
 from database.leverage_db import init_db as ensure_leverage_tables_exists
 from database.sandbox_db import init_db as ensure_sandbox_tables_exists
 from database.settings_db import init_db as ensure_settings_tables_exists
+from database.superorder_db import init_db as ensure_superorder_tables_exists
 from database.strategy_db import init_db as ensure_strategy_tables_exists
 from database.symbol import init_db as ensure_master_contract_tables_exists
 from database.telegram_db import get_bot_config
@@ -515,6 +516,7 @@ def setup_environment(app):
                 ("Strategy DB", ensure_strategy_tables_exists),
                 ("Sandbox DB", ensure_sandbox_tables_exists),
                 ("Action Center DB", ensure_action_center_tables_exists),
+                ("Super Order DB", ensure_superorder_tables_exists),
                 ("Chart Prefs DB", ensure_chart_prefs_tables_exists),
                 ("Market Calendar DB", ensure_market_calendar_tables_exists),
                 ("Qty Freeze DB", ensure_qty_freeze_tables_exists),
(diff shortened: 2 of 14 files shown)


marketcalls:
Thanks @radhakrishnan4me-code, appreciate the substantial work here. The OCO-style logic with trailing SL + the auto-starting monitor is non-trivial, and keeping it strictly broker-agnostic via `get_multiquotes` / `get_order_status` / `place_order` / `cancel_order` is exactly the right shape. The Aliceblue WebSocket fixes (no duplicate reconnect threads, in-memory quote cache to prevent reconnect storms) are good catches that go beyond the Super Order scope.

### My plan: merge this as Super Order v1 once the items below are resolved

Before I get to the merge bar, here's how we're thinking about order APIs going forward, because OpenAlgo has grown into something more than a REST wrapper. An order primitive today has to weave through seven pipelines:

1. **Execution surfaces** — live + sandbox/analyzer mode + latency tracking; logs to `apilog_db` (live) and `analyzer_db` (sandbox)
2. **Bruno collection** — `.bru` files for the API playground
3. **React UI**
4. **Action Center** — semi-automatic approval queue, both live and analyze modes
5. **OpenAlgo Flow** — node type for the workflow builder
6. **Telegram alerts** on lifecycle events
7. **Toast notifications** via `socketio.emit`

Canonical worked example: the [GTT orders implementation plan](https://github.com/marketcalls/openalgo/blob/main/docs/plans/2026-04-22-gtt-orders-implementation-plan.md). I'm putting a contributor checklist (`docs/plans/order-api-pipelines-checklist.md`) together so this is in writing for the next PR rather than buried in review comments — none of this was documented anywhere when you started, and the work you've done is solid.

### Where this PR maps today

| Pipeline | Status |
|---|---|
| Bruno collection (`.bru`) | ✅ |
| Live REST + DB + monitor | ✅ |
| `apilog_db` for user-initiated calls | ✅ |
| Sandbox/analyzer routing | ❌ — monitor forces `{"mode": "live"}` |
| `analyzer_db` logs | ❌ |
| Latency tracking for monitor auto-actions | ❌ |
| React UI | ❌ |
| Action Center | ❌ |
| Flow node | ❌ |
| Telegram alerts | ❌ |
| Toast notifications | ❌ |

### ❗ Critical: please address the cubic-dev-ai review comments before merge

cubic-dev-ai surfaced **4 P1** and **10 P2** findings in its review, and after walking through them I'd say almost all are genuine and would otherwise come back as bug reports after release. **Resolving these is a precondition for merge** — please go through each, validate (a few might be false positives once you trace the code paths), and either fix or reply with why it's not actually a problem.

The four P1s in particular need attention:

- `restx_api/schemas.py:452` — **SuperOrder schema field renames are incompatible with service expectations**, causing missing-key `KeyError`s at runtime during order placement. If correct, the feature won't work end-to-end as it stands; this is the single most important one to validate first.
- `services/superorder_service.py:351` — ACTIVE superorder cancellation ignores child-order cancel failures and commits `CANCELLED` regardless, leaving the DB out of sync with the broker state. Real money implication: the user's `/superorderbook` shows the order cancelled, but the broker still has a live SL/target.
- `services/superorder_monitoring_service.py:140` — expired-order handler overwrites status before the branch checks run, so the broker-cancel call is skipped on expiry. Same DB/broker drift problem in the time-out path.
- `blueprints/orders.py:181` — `/superorderbook` route uses a JSON-only condition that fails for normal GET requests, falling back to a missing template and 500-ing on the simplest case.

The P2s are mostly state-hygiene and contract-consistency issues — process-level guard against duplicate monitor instances under the dev reloader, intraday expiry-timestamp-in-the-past edge case, broad `except Exception` swallowing client JSON errors as 500s, RESTX/Marshmallow type drift, the crypto Bruno collection using stock-market sample values, hot-loop INFO logging, batch multiquote not unsubscribing leaving subscription state to grow, and cached quote results skipping normalisation.

Please reply to each cubic-dev comment in-line (either with the commit that fixes it, or with rationale if you think it's a false positive). The two state-divergence bugs (P1 on cancel + P1 on expiry) and the schema-rename P1 are real-money bugs and shouldn't ship.

### Three platform-side asks for v1

In addition to the cubic-dev items, three small additions before merge:

1. **Sandbox routing.** Route the monitor's auto-actions through `services/order_router_service.py` instead of forcing `{"mode": "live"}`. That handles the live/sandbox switch transparently — you should get analyzer-mode parity for free, or close to it.
2. **`socketio.emit("superorder_update", {...})`** on every state transition (`PENDING → ACTIVE`, target hit, SL hit, trailed, expired). Three lines; lets the future UI show real-time updates without polling.
3. **Telegram lifecycle hook.** Half a dozen lines hitting `services/telegram_bot_service.notify(...)` on target/SL/trail/expiry. Users running Super Order unattended will absolutely want this — pairs naturally with the monitor you've already built.

### After merge

Once the cubic-dev items are addressed and the three platform-side asks are in, I'll merge as **Super Order v1** and open a tracking issue listing the remaining four pipelines (React UI, Action Center, Flow node, full latency instrumentation). Happy to do those in follow-up PRs myself or coordinate with you — whatever works for your bandwidth.

Thanks again. The platform-thinking takes a while to internalize; once you've done one feature through all seven pipelines, the next one is much faster.

Issue #882, opened 2026-02-10
Greeks in OptionChain


kumar9:
Kindly add greeks in the option chain of OpenAlgo, it will help many to trade based on greeks

marketcalls:
Thanks @kumar9 — this is actively in progress and the foundation just landed on `main`.

**What just shipped (commit [8d973504](https://github.com/marketcalls/openalgo/commit/8d973504))**

We migrated the Black-76 Greeks/IV backend from `py_vollib` to [`opengreeks`](https://pypi.org/project/opengreeks/), a Rust-core reimplementation:

- Bit-identical Greeks (Δ/Γ/Θ/Vega exact, ρ at float-64 last bit, IV ~13 significant digits)
- Pure-math speedup: implied volatility 17.3 µs → 0.38 µs (**46×**), full 40-strike chain refresh 1.49 ms → 0.12 ms (**12.8×**)
- One runtime dependency (NumPy) instead of seven

The full parity + benchmark report is in [`docs/benchmarks/greeks_opengreeks_parity.md`](https://github.com/marketcalls/openalgo/blob/main/docs/benchmarks/greeks_opengreeks_parity.md).

**Why this matters for option chain with Greeks**

Pre-migration, computing IV + 5 Greeks for ~200 option contracts took several milliseconds of pure CPU per refresh — too heavy to run on every tick. Post-migration, the same chain refresh is ~100 µs of math, which is fast enough to:

1. Enrich the existing REST option-chain endpoint with Greeks per strike (in-flight)
2. Stream Greeks **live over WebSocket** as the underlying ticks, so subscribers see Δ/Γ/Θ/Vega/IV updating in real time alongside LTP and OI

**What's coming next**

- Greeks-enriched option chain in the REST `/api/v1/optionchain` response (additive, backward-compatible)
- WebSocket option-chain stream with per-tick Greeks (new endpoint, opt-in)
- Black-Scholes / Black-Scholes-Merton models exposed as a separate API path for users who want dividend-aware Greeks for indices (tracked in #677)

Keeping this issue open as the tracking ticket for the option-chain Greeks rollout. Will link the PRs here as they ship.

marketcalls:
in the latest version it is included.

PR #1327, opened 2026-05-02
refactor(greeks): replace py_vollib with inline Black-76 module

iamrony777:
#### **py_vollib library not installed** - ( while loading __/ivsmile__ & __/strategybuilder__ (greeks tab) in the dashboard ) 

```
[2026-05-02 16:29:46,022] INFO in data: Requesting OHLC for 2 FNO instruments
[2026-05-02 16:29:46,022] INFO in data: Requesting OHLC with exchange_symbols: NSE_NIFTY2642125250CE,NSE_NIFTY2642125250PE...
[2026-05-02 16:29:46,022] INFO in data: Making direct API request to endpoint: /v1/live-data/ohlc
[2026-05-02 16:29:46,134] INFO in data: Successfully processed 102 quotes in 3 batches
[2026-05-02 16:29:46,135] ERROR in option_greeks_service: py_vollib library not installed.
...
```
Even though it was installed ( installed *vollib* later ) 

```bash
(openalgo) rony@fedora:~/Trade/openalgo$ uv pip list | grep vollib
py-vollib                 1.0.1
vollib                    1.0.7 
```

`py_vollib==1.0.1` is broken on standard CPython 3.12. Its hard dependency `py_lets_be_rational==1.0.1` does:

```python
from _testcapi import DBL_MIN, DBL_MAX
```

`_testcapi` is a CPython-internal test module that isn't shipped in any standard release build (Fedora, Debian, linuxbrew, uv-managed Python). Both 1.0.0 and 1.0.1 carry this bug, and they're the only `py_lets_be_rational` releases that ever shipped. The library simply cannot import on this project's interpreter.

Bumping to `py_vollib==1.0.7` doesn't help either — it's now a one-line shim that pulls `vollib==1.0.7`, and `vollib` pins `numpy<2.0,>=1.20`. That's a hard resolver conflict with our `numpy==2.4.4`.

I also tested `uv add git+https://github.com/vollib/py_vollib.git` directly. uv rejects it with the same `numpy<2.0` constraint coming from `vollib`'s `pyproject.toml`. (The upstream `py_lets_be_rational` master branch *has* fixed the `_testcapi` bug, but the `py_vollib` master branch is the vollib 1.0.7 chain, so we'd still inherit the numpy pin.)

Only two files in our codebase actually call `py_vollib`, and both use a small, well-understood surface: the Black-76 IV solver plus analytical Greeks. We already depend on `scipy==1.17.1`, which gives us `scipy.stats.norm` (CDF/PDF) and `scipy.optimize.brentq` (root-finder) — everything needed.

### What changed

**New: `utils/black76.py`** — 76 LOC, pure Python.

Exports `implied_volatility`, `delta`, `gamma`, `theta`, `vega`, `rho` with **the same call signatures `py_vollib.black` used**, so callers swap imports without changing arguments. Greeks come back in py_vollib's trader-friendly units (theta daily, vega/rho per 1% absolute change), matching the conventions already documented in the service comments.

Below-intrinsic prices raise `BelowIntrinsicException` (a `ValueError` subclass) with "intrinsic" in the message. That preserves the existing string-matching fallback in `calculate_greeks` that returns theoretical deep-ITM Greeks instead of erroring.

**Service updates** — import-only swaps:

- `services/option_greeks_service.py`: switched to `from utils.black76 import …`. Removed the unused `check_pyvollib_availability` helper and the lazy-import-with-fallback pattern (no longer needed — `utils.black76` cannot fail to import).
- `services/iv_chart_service.py`: same import swap. Dropped the `try/except ImportError` guard at the top of `get_iv_chart_data` and the lazy imports inside `_calculate_iv_series`.

**Dependency cleanup:**

- Removed `py_vollib==1.0.1` and `py_lets_be_rational==1.0.1` from `pyproject.toml`, `requirements.txt`, `requirements-nginx.txt`.
- Regenerated `uv.lock` via `uv sync` — both packages uninstalled cleanly.

**Cosmetics:**

- Updated docstrings/comments and stale "Install with: pip install py_vollib" instructions in `services/iv_chart_service.py`, `services/option_greeks_service.py`, `test/test_multi_option_greeks_api.py`, and `broker/dhan_sandbox/api/data.py`.

### Verification

**Numerical parity** against the canonical Black-76 textbook example (F=K=100, t=0.5, r=0.05, σ=0.20):

| Quantity | Computed | Expected |
|---|---|---|
| Call price | 5.498015 | 5.4988 |
| IV round-trip | 0.200000 | 0.20 |
| Δ (call) | 0.5151 | 0.5152 |
| Δ (put) | -0.4602 | -0.4601 |
| Γ | 0.0274 | 0.0275 |
| Δc - Δp | 0.9753 | df = e^(-rt) = 0.9753 ✓ put-call parity |
| ρ | -0.027490 | -t·price/100 ✓ Black-76 form |

**Edge cases:** below-intrinsic input raises `BelowIntrinsicException` with "intrinsic" in the message, so the existing deep-ITM fallback in `option_greeks_service.py` still triggers. Far-OTM inputs solve via `brentq` (sigma≈0.53 for the test case) without convergence failures.

**Live API smoke tests** against a running server:

- `POST /api/v1/optiongreeks` with `NIFTY05MAY2624000CE` (~3 days to expiry) at spot 23997.55, price 206.6 → IV=24.03%, Δ=0.5025, Γ=0.000766, Θ=-34.89, ν=8.65. Matching PE returned Δ=-0.4991 — Δc−Δp ≈ 1.0 as expected ATM.
- `POST /api/v1/multioptiongreeks` with four legs across strikes 23900/24000/24100 → 4/4 success, deltas dropped monotonically across rising call strikes (0.5541 → 0.5051 → 0.4503), thetas all negative. Sanity-check passed.

### Trade-off note (for reviewers)

I considered installing `py_lets_be_rational` from upstream master (where the `_testcapi` bug is fixed) and keeping `py_vollib==1.0.1` from PyPI (the original library, no numpy upper bound). That combination would technically work, but:

- Pins us to an unversioned git SHA (no semver, force-push risk).
- Slower CI/Docker builds (poetry-core source build per environment).
- Both upstreams are effectively unmaintained.

For a 6-function, well-defined math surface, the cost of carrying a 76-LOC local module is significantly lower than the cost of pinning an abandoned git tree.

### Follow-ups

The follow-up commit (`style(greeks)`) applies `ruff format` and resolves the `I001`/`B904` lints introduced by the import swap. Pre-existing `UP035` (`typing.Dict`/`Tuple`) and `B904` warnings in `parse_option_symbol` are out of scope and left untouched.

### Test plan checklist

- [x] `uv sync` regenerates lockfile cleanly; `uv pip list | grep -iE "vollib|lets_be_rational"` returns empty
- [x] `uv run python -c "from utils.black76 import …"` imports
- [x] `uv run python -c "from services.option_greeks_service import calculate_greeks"` imports
- [x] `uv run python -c "from services.iv_chart_service import get_iv_chart_data"` imports
- [x] Numerical parity vs textbook Black-76 (tolerance 1e-4)
- [x] Live `/api/v1/optiongreeks` returns sane Greeks
- [x] Live `/api/v1/multioptiongreeks` returns sane Greeks for batch input
- [x] `ruff check` introduces no new lint errors on touched files
- [x] `ruff format --check` passes on touched files

---
## Summary by cubic
Replaced `py_vollib` with a built-in Black-76 module (`utils.black76`) to compute IV and Greeks, fixing Python 3.12 import failures and removing the `numpy<2.0` constraint. Services keep the same call signatures, so behavior and APIs stay the same.

- **Refactors**
  - Added `utils/black76.py` with `implied_volatility`, `delta`, `gamma`, `theta`, `vega`, `rho` using `scipy` (`norm`, `brentq`).
  - Preserves `py_vollib` units (theta daily; vega/rho per 1%); raises `BelowIntrinsicException` to keep the deep-ITM fallback.
  - Swapped imports in `option_greeks_service.py` and `iv_chart_service.py`; removed lazy import guards and `check_pyvollib_availability`.
  - Updated docstrings/comments and tests to remove “install `py_vollib`” notes; minor sandbox comments clarified.

- **Dependencies**
  - Removed `py_vollib==1.0.1` and `py_lets_be_rational==1.0.1` from dependency files; regenerated `uv.lock`.

<sup>Written for commit 4e95ba73bb9399bd74928df4c6c7957359972c04. Summary will update on new commits.</sup>


marketcalls:
Thank you for the careful work on this, @iamrony777 — really appreciate it.

The root-cause writeup is excellent. You correctly identified that `py_vollib==1.0.1` is broken on standard CPython 3.12 because of the `_testcapi` import in its `py_lets_be_rational` dependency, and that bumping to `py_vollib==1.0.7` doesn't help because it drags in a `numpy<2.0` pin that conflicts with the rest of the project. That diagnostic chain alone saves the next maintainer several hours of head-scratching.

I have to close this PR, but the reason is purely timing, not quality: commit [8d973504](https://github.com/marketcalls/openalgo/commit/8d973504) (merged earlier this week) solved the same problem via a different route — we replaced `py_vollib` with [`opengreeks`](https://pypi.org/project/opengreeks/), a Rust + PyO3 reimplementation of the same Black-76 surface. The full parity + benchmark report lives at [`docs/benchmarks/greeks_opengreeks_parity.md`](https://github.com/marketcalls/openalgo/blob/main/docs/benchmarks/greeks_opengreeks_parity.md).

Three concrete gains from going with `opengreeks` that pushed us in that direction:

**1. Numerical agreement vs the vollib oracle is at machine precision**

| Quantity | Max abs error vs `vollib==1.0.7` |
|---|---|
| Δ, Γ, Θ, Vega | `0.0e+00` (bit-for-bit identical) |
| ρ | `7.9e-16` (float-64 last bit) |
| implied volatility | `4.1e-13` (~13 significant digits) |

A `scipy.optimize.brentq`-based IV inversion is correct, but typically agrees with vollib only to ~1e-4 (your textbook table showing 5.498015 vs 5.4988 reflects this). That gap doesn't matter for a single API call, but it becomes visible once Greeks stream over WebSocket per tick — which is what issue #882 is tracking.

**2. Per-call latency drops by an order of magnitude**

Measured on the live `/api/v1/optiongreeks` path (40-sample replay against the same NIFTY chain):

| Function | py_vollib | opengreeks | Speedup |
|---|---:|---:|---:|
| `implied_volatility` | 17.29 µs | 0.38 µs | **46×** |
| `theta` | 5.79 µs | 0.21 µs | **28×** |
| `rho` | 3.96 µs | 0.21 µs | **19×** |
| Δ / Γ / Vega | ~1.5 µs | 0.21 µs | **7–8×** |
| **Full 40-strike chain (IV + 5 Greeks)** | **1.485 ms** | **0.116 ms** | **12.8×** |

That headroom is what makes live per-tick Greeks for a 200-strike option chain feasible without saturating a CPU core, which is the direction the option-chain WebSocket streaming work is going next.

**3. Removes `scipy` from the Greeks path entirely**

`opengreeks` ships a Rust core with only NumPy at runtime — no scipy, no numba, no pandas, no `py_lets_be_rational`. That's six fewer transitive dependencies than the `py_vollib` chain, and crucially it keeps the door open for slimmer deployment targets (containers, embedded analytics) without dragging in a full SciPy stack.

Your `utils/black76.py` is clean, the call-signature compatibility with `py_vollib.black` is the right design choice, and the live-API smoke tests (NIFTY 23900/24000/24100 with delta monotonicity check) are the right verification level. If the math layer specifically interests you, [`opengreeks`](https://github.com/marketcalls/opengreeks) is open source and contributions are very welcome — especially for the Heston, SABR, and American/Bermudan pricing models that we'll need for OpenAlgo's options analytics roadmap. Closing this PR as superseded by `8d973504`, but with thanks for the effort.

PR #1485, opened 2026-06-07
feat(frontend): replace loading placeholders with skeleton animations…

gelogonza:
## Summary of changes
- Holdings, Positions, and OrderBook stats cards show skeleton rectangles instead of `---` or `0` while data loads
- Imported `Skeleton` from the existing `@/components/ui/skeleton` component in all three pages
- Skeleton dimensions match actual `text-2xl` CardTitle content (h-7, widths vary by content type)

## Testing plan
- [ ] Hard-reload Holdings page with Slow 3G throttle — stats cards pulse grey before data populates
- [ ] Hard-reload Positions page — same behavior
- [ ] Hard-reload Order Book page — same behavior
- [ ] After data loads, skeletons are gone and real values render correctly
- [ ] Manual Refresh button does not re-trigger skeleton (isLoading stays false on refresh)

---
## Summary by cubic
Replace `---` and 0 placeholders with animated skeletons in stats cards on Holdings, Positions, and Order Book to improve loading feedback and eliminate PnL color flicker. Previously we showed `---`/0 and applied red/green immediately; now we render skeletons while loading and apply PnL colors only after data loads.

- Use `Skeleton` from `@/components/ui/skeleton` in all three pages; height matches `text-2xl` (h-7) with widths sized per metric (counts w-16; amounts w-28/w-32; percent w-20).
- Holdings: skeletons for Total Holding Value, Total Investment Value, Total PnL, and Total PnL %; when not loading but no data, fall back to `---`.
- Positions: skeletons for Open/Long/Short counts and Total PnL; PnL color applies only when not loading.
- Order Book: skeletons for Buy/Sell/Completed/Open/Rejected counts; after load, counts still fall back to 0 if missing.

<sup>Written for commit 032ee7d094a1ad98db56f288d8d6561ecf1c04ab. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1485?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


marketcalls:
Thanks for the PR. The direction is right and it fits how the codebase already works, so this is close. There is one blocker plus a few changes I would like in before merge.

Reviewed at commit `032ee7d`.

## Required: missing import breaks the build

`frontend/src/pages/Holdings.tsx` uses `<Skeleton>` at lines 275, 287, 308 and 338, but the import was never added. The PR description says the import went into all three pages, but the diff only adds it to `OrderBook.tsx` and `Positions.tsx`.

Add to `frontend/src/pages/Holdings.tsx`:

```tsx
import { Skeleton } from '@/components/ui/skeleton'
```

`npm run build` is `tsc -b && vite build`, so this currently fails with `TS2304: Cannot find name 'Skeleton'`. Note that `npm run lint` is `biome lint ./src` and Biome's `noUndeclaredVariables` is off by default, so lint will not catch this - only the build job does.

This matters more than a normal build break here: `frontend/dist/` is tracked on `main` and rebuilt by the `commit-dist` CI job after every push. A failing build on `main` means the dist never rebuilds, and a plain `git pull` is the canonical upgrade path for production deployments.

## Requested changes

1. **Use `h-6`, not `h-7`.** `CardTitle` carries `leading-none`, which overrides the line-height that `text-2xl` would otherwise set. The rendered text height is `1.5rem` (24px), not 28px, so every skeleton is 4px too tall and introduces a small layout shift when data lands - the exact thing this PR is meant to remove. Please change `h-7` to `h-6` in all skeletons across the three files.

2. **Flatten the nested ternaries in the PnL colour logic** (Holdings x2, Positions x1). `cn`/clsx drops falsy values, so this reads much better and behaves identically:

   ```tsx
   className={cn(
     'text-2xl',
     !isLoading && (isProfit(stats.totalPnl) ? 'text-green-600' : 'text-red-600')
   )}
   ```

3. **Three lines exceed the Biome `lineWidth` of 100:** `OrderBook.tsx:579` (104), `OrderBook.tsx:595` (103), `Positions.tsx:861` (108). CI only runs `biome lint`, not `format --check`, so these will not fail the build, but they will churn the next time anyone runs `npm run format`. Running `npm run format` on the three files should settle it.

4. **Mixed loading vocabulary on the same screen.** All three pages still use the `Loader2` spinner for the table body (`Holdings.tsx:332`, `Positions.tsx:866`, `OrderBook.tsx:601`). After this change a user sees pulsing skeleton cards and a spinning circle directly below them at the same time. Skeleton rows for the table would make each page read as one coherent loading state.

5. **`TradeBook.tsx` is left behind.** It has the identical pattern - three stats cards at lines 444-462, the same first-load-only `isLoading` flag, the same spinner at line 473. Without it the trading section is inconsistent page to page. Happy to take this as a follow-up PR if you would rather keep this one tight, but it should not stay unaddressed.

## What checks out

- The approach is consistent with existing usage of `Skeleton` in MasterContract, the strategy pages, chartink and the python-strategy pages.
- I verified your claim about refresh: in all three pages `isLoading` starts `true` and is only ever set to `false`, never back to `true`. Manual refresh and polling will not re-flash the skeletons over live data. That is the correct behaviour.
- The loaded-but-empty `---` fallback in Holdings is properly preserved.
- No security concerns. The change is purely presentational, adds no dependencies, and does not touch data fetching, auth, or the CSV export path.

## CI

The CI run on this branch is sitting at `action_required` and has never executed, since this is a fork PR. The green checks you see are only cubic and Snyk. Nothing has type-checked or built this branch yet - which is why the missing import went unnoticed. Once you push the fixes I will approve the workflow run so `frontend-build`, `frontend-lint`, `frontend-test` and `frontend-e2e` actually run.

Please push the import fix and items 1 to 3 at minimum, and we can merge once CI is green.

Issue #1490, opened 2026-06-09
Please add or fix endpoint (/api/websocket/subscriptions)


sbt987:
@marketcalls When websocket is active I need to check (with non-blocking, read-only connection) which tells me the full list of instruments currently subscribed in websocket pool. There isn't any explicit command in the API playground. 

<img width="410" height="500" alt="Image" src="https://github.com/user-attachments/assets/f8ced216-032e-4c43-9fae-f72b89f006b9" />

I need this because right now I'm shooting the 'unsubscribe all' command a few times a day and resubscribing to a different list. This takes >10 seconds each time.

If I could get the subscribed instruments list, then I could use it to 'unsubscribe only the unnecessary instruments' and the other instruments would have a smooth, uninterrupted data stream.

@Azhagesan-dev mentioned that "There is no local database or file for this — all WebSocket subscription state is purely in-memory inside the proxy server process (port 8765), stored in WebSocketProxy.subscription_index. There's no existing read-only query to dump it.

The existing GET /api/websocket/subscriptions endpoint is broken — it creates a fresh WebSocketClient on every call that has an empty subscription dict, so it always returns {"count": 0, "subscriptions": []} regardless of actual subscriptions.

To actually see the live global subscription list, we'd need to add a new action to the proxy server (like {"action": "get_all_subscriptions"}) that returns its subscription_index over the existing WebSocket connection."

PR #1495, opened 2026-06-10
feat(ws): add get_subscriptions and get_global_subscriptions actions (#1490)

Azhagesan-dev:
Addresses #1490 by adding two new WebSocket actions for client subscription visibility and diagnostics:

- **`get_subscriptions`**: Returns the requesting client's active subscriptions with mode and exchange breakdown.
- **`get_global_subscriptions`**: Returns per-client subscriptions with a global summary (total clients, total subscriptions, by mode, by exchange).

### Other changes:
- **Robustness Fix:** Wrapped adapter cleanup on disconnect in a `try...except...finally` block in `cleanup_client` to prevent disconnected/inactive clients from lingering in memory with 0 subscriptions if the broker disconnect fails.
- **Playground Integration:** Added templates for the new actions to the WebSocket Composer dropdown list.
- **Docs:** Updated `WEBSOCKET_README.md` to document the new `.bru` template files.

---
## Summary by cubic
Adds `get_subscriptions` and `get_global_subscriptions` WebSocket actions to inspect active subscriptions and global stats, and hardens disconnect cleanup to prevent stale clients and orphaned broker streams. Adds Playground templates, docs, and regression tests. Addresses #1490.

- **New Features**
  - `get_subscriptions`: Returns the requesting client’s subscriptions with symbol, exchange, canonical mode, depth, and a summary (count, by_mode, by_exchange, broker).
  - `get_global_subscriptions`: Returns per-client subscriptions plus a global summary (total_clients, subscriptions_count, by_mode, by_exchange, broker). Note: `subscriptions_count` is unique broker-level entries, not the sum of per-client counts.
  - Added WebSocket Composer templates and updated `WEBSOCKET_README.md`.

- **Bug Fixes**
  - Scoped adapter cleanup to the full-disconnect branch; keep-alive preserved for `flattrade`/`shoonya` via `unsubscribe_all` with try/except. Safe map removals using `.pop(...)` and set `.discard(...)`.
  - Defensive cleanup now purges `subscription_index` and calls `adapter.unsubscribe(symbol, exchange, mode)` when dropping the final client for a key, preventing orphaned ticks even if another client remains.
  - Normalized subscription parsing via `_parse_subscriptions`; global view filters to authenticated (active) clients. Added regression tests covering normal/raises disconnects, keep-alive paths, multi-client, and defensive sweep.

<sup>Written for commit 84ed528f27efc9e2b5f505b1354d5af60d248e88. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1495?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/collections/openalgo/IN_stock/WEBSOCKET_README.md
+++ b/collections/openalgo/IN_stock/WEBSOCKET_README.md
@@ -86,6 +86,8 @@ Valid string modes: `"LTP"`, `"Quote"`, `"Depth"`
 | ws_unsubscribe_all.bru | unsubscribe_all | Unsubscribe from all |
 | ws_get_broker_info.bru | get_broker_info | Get broker information |
 | ws_get_supported_brokers.bru | get_supported_brokers | List supported brokers |
+| ws_get_subscriptions.bru | get_subscriptions | Get own active subscriptions |
+| ws_get_global_subscriptions.bru | get_global_subscriptions | Get client-wise subscription details & summary |
 | ws_ping.bru | ping | Test connection latency |
 
 ## Supported Exchanges
--- a/collections/openalgo/IN_stock/ws_get_global_subscriptions.bru
+++ b/collections/openalgo/IN_stock/ws_get_global_subscriptions.bru
@@ -1,0 +1,45 @@
+meta {
+  name: Get Global Subscriptions
+  type: websocket
+  seq: 15
+}
+
+websocket {
+  url: ws://127.0.0.1:8765
+  description: Get client-wise subscription breakdown with global summary across all connected clients
+}
+
+message:json {
+  {
+    "action": "get_global_subscriptions"
+  }
+}
+
+docs {
+  Returns a per-client breakdown of all active subscriptions plus global summary stats.
+
+  Server responds with:
+  {
+    "type": "global_subscriptions",
+    "status": "success",
+    "summary": {
+      "total_clients": 2,
+      "subscriptions_count": 25,
+      "by_mode": {"LTP": 20, "Quote": 15, "Depth": 5},
+      "by_exchange": {"NSE": 18, "NSE_INDEX": 5, "BSE": 2},
+      "broker": "angel"
+    },
+    "clients": [
+      {
+        "client_id": 140234567890,
+        "subscription_count": 25,
+        "subscriptions": [
+          {"symbol": "RELIANCE", "exchange": "NSE", "mode": "Quote", "depth": 5},
+          {"symbol": "NIFTY", "exchange": "NSE_INDEX", "mode": "LTP", "depth": 5}
+        ]
+      }
+    ]
+  }
+
+  Must be authenticated first.
+}
--- a/collections/openalgo/IN_stock/ws_get_subscriptions.bru
+++ b/collections/openalgo/IN_stock/ws_get_subscriptions.bru
@@ -1,0 +1,38 @@
+meta {
+  name: Get Subscriptions
+  type: websocket
+  seq: 14
+}
+
+websocket {
+  url: ws://127.0.0.1:8765
+  description: Get the requesting client's own active subscriptions with symbol, mode, and summary stats
+}
+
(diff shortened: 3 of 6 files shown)


sbt987:
@marketcalls can you pls check and approve this if everything checks out? It is adding commands to return the current list of subscribed instruments in websocket via the API.

marketcalls (commented):
Thanks for tackling #1490 — `get_subscriptions` / `get_global_subscriptions` are a solid fix for the broken subscription-visibility gap, and the Playground/docs additions are appreciated.

One thing needs to be fixed before merge:

### `cleanup_client`: the new `finally` block undoes the Flattrade/Shoonya keep-alive behavior

```python
try:
    if broker_name in ["flattrade", "shoonya"] and hasattr(adapter, "unsubscribe_all"):
        adapter.unsubscribe_all()      # intended to KEEP the adapter alive
    else:
        adapter.disconnect()           # intended to fully tear it down
except Exception as e:
    logger.exception(...)
finally:
    self.broker_adapters.pop(user_id, None)
    self.user_broker_mapping.pop(user_id, None)
```

In the pre-PR code, `del self.broker_adapters[user_id]` / `del self.user_broker_mapping[user_id]` only ran in the `else` branch. Moving them into `finally` makes them run unconditionally, including in the Flattrade/Shoonya "keep the connection alive" branch — which defeats the point of that branch.

Concretely: once the last client for a Flattrade/Shoonya user disconnects, the broker WebSocket stays physically open (as intended), but the server forgets it's tracking it. `get_adapter_health()` and `_log_stale_adapters()` both iterate `self.broker_adapters` for health/observability, so this now-orphaned connection becomes invisible to both. It mostly self-heals on the next client reconnect (the pool is still cached in `broker_factory._POOLED_ADAPTERS`, and `initialize()`/`connect()` are idempotent), but if the user never reconnects, it's an untracked open connection until process restart.

**Suggested fix:** keep the `.pop(...)` calls scoped to the `else` (full-disconnect) branch only, so Flattrade/Shoonya sessions keep their adapter mapping alive exactly like before — wrap just the two calls that can actually raise (`unsubscribe_all()` / `disconnect()`) in try/except, which was the actual goal of this fix.

Please be careful this change doesn't regress WebSocket connectivity for the other brokers — I'd want to see the normal full-disconnect path (e.g. Zerodha/Angel) and the Flattrade/Shoonya keep-alive + reconnect path both exercised before merge, to confirm every broker's teardown/reconnect still behaves the same as before this PR. A regression test in `test/test_websocket.py` covering `cleanup_client`'s exception path would also help catch this class of bug going forward.

Non-blocking nits:
- `get_global_subscriptions.summary.subscriptions_count` counts distinct `(symbol, exchange, mode)` entries in `subscription_index`, while each client's own `subscription_count` is that client's raw count — these can diverge when multiple sessions overlap on the same symbol. Worth a one-line doc comment clarifying it's "unique broker-level subscriptions," not a sum of per-client counts.
- The per-subscription parsing loop is duplicated between `get_subscriptions` and `get_global_subscriptions` — could factor into a small helper.

sbt987:
Can this issue be reviewed again and fixed and merged? @marketcalls @Azhagesan-dev 

I recalled this issue because of the behavior detailed in #1764 which tells how an instrument is not actually unsubscribed until the LAST websocket connection which uses it has unsubscribed. 

Given this behavior, it is all the more important to know via API which instruments are currently subscribed - which is tackled by this PR.

Edit: it would also be useful to have a **global** unsubscribe() API call that accepts a [list of instruments].

sbt987:
Basically I'm talking about github issue #1490

PR #1509, opened 2026-06-13
fix(shoonya): include uid in GenAcsTok payload to fix INVALID_VERIFIER OAuth login

### LINKED ISSUE #1508: Shoonya OAuth login fails with INVALID_VERIFIER — GenAcsTok payload missing uid
## Summary

Retail Shoonya accounts using the OAuth (GenAcsTok) login cannot connect. The token exchange fails with `Invalid Input : INVALID_VERIFIER`, even though the checksum `SHA256(client_id + secret + code)` is computed correctly.

## Root cause

`broker/shoonya/api/auth_api.py` builds the GenAcsTok `jData` payload as:

```python
payload = {
    "code": code,
    "checksum": checksum,
}
```

It omits the **`uid`** field. Finvasia's own reference library `NorenRestApiOAuth.getAccessToken` sends `uid`:

```python
values = {
    "code": authcode,
    "checksum": app_verifier,   # sha256(client_id + Secret_Code + authcode)
    "uid": UID
}
```

Without `uid`, Shoonya cannot bind the checksum to the account and returns `INVALID_VERIFIER`.

## Reproduction

1. Configure a Shoonya broker with `BROKER_API_KEY=userid:::client_id` and `BROKER_API_SECRET`.
2. Complete the OAuth authorize step so the `code` reaches `/<broker>/callback`.
3. GenAcsTok returns `Invalid Input : INVALID_VERIFIER`; the UI loops back to `/broker`.

## Fix

Add `uid` (the userid segment of `BROKER_API_KEY`, split on `:::`) to the payload. Verified end-to-end on a live FN-series Shoonya account on 2026-06-13 — after the fix the master contract downloaded and ~186k symbols loaded into cache.

Secondary (optional, included in the PR): the success branch requires `stat == "Ok"`, but the GenAcsTok success response is not documented to carry it — the reference library keys success on `access_token` presence. The PR aligns this to avoid a false-negative.

PR fixing this follows.

swadeesh-s:
## What

Fixes Shoonya OAuth login failing with `Invalid Input : INVALID_VERIFIER`.

Closes #1508.

## Why

`broker/shoonya/api/auth_api.py` omits the **`uid`** field from the GenAcsTok `jData` payload. Finvasia's own `NorenRestApiOAuth.getAccessToken` sends it:

```python
values = {"code": authcode, "checksum": app_verifier, "uid": UID}
```

Without `uid`, Shoonya returns `INVALID_VERIFIER` even with a correct `SHA256(client_id + secret + code)` checksum, so retail Shoonya OAuth accounts cannot connect.

## Changes

- Extract `userid` from `BROKER_API_KEY` (`userid:::client_id`) and include `"uid": userid` in the GenAcsTok payload. **(the fix)**
- Key the success branch on `access_token` presence rather than requiring `stat == "Ok"`, matching the reference library — the GenAcsTok success response is not documented to carry `stat`, so this avoids a false-negative. Loosening it can only accept valid token-bearing responses, never reject them.

The checksum construction and endpoint are unchanged.

## Verification

Verified end-to-end on a live FN-series Shoonya account on 2026-06-13: after the change the token exchange succeeds, the master contract downloads, and ~186k symbols load into cache. Two-line behavioural change; no other broker touched.

---
## Summary by cubic
Fixes Shoonya OAuth login errors (`INVALID_VERIFIER`) by sending the required `uid` in the GenAcsTok request. Also updates success handling to accept any response that includes `access_token`.

- **Bug Fixes**
  - Extract `userid` from `BROKER_API_KEY` (`userid:::client_id`) and include `"uid": userid` in the GenAcsTok `jData`.
  - Consider the exchange successful when `access_token` is present, instead of requiring `stat == "Ok"`.

<sup>Written for commit 130a956415681d25fe56d9e7c957eab5b891d2ad. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1509?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/broker/shoonya/api/auth_api.py
+++ b/broker/shoonya/api/auth_api.py
@@ -18,6 +18,7 @@ def authenticate_broker(code):
         full_api_key = os.getenv("BROKER_API_KEY")
         if not full_api_key or ":::" not in full_api_key:
             return None, "BROKER_API_KEY must be in format userid:::client_id"
+        userid = full_api_key.split(":::")[0]  # Noren user id
         client_id = full_api_key.split(":::")[1]  # appKey / client_id
         secret_key = os.getenv("BROKER_API_SECRET")
         if not secret_key:
@@ -33,10 +34,13 @@ def authenticate_broker(code):
         checksum_input = f"{client_id}{secret_key}{code}"
         checksum = hashlib.sha256(checksum_input.encode()).hexdigest()
 
-        # Prepare token exchange payload
+        # Prepare token exchange payload. Shoonya's GenAcsTok requires the
+        # uid field; without it the server returns INVALID_VERIFIER even with
+        # a correct checksum (matches Finvasia's NorenRestApiOAuth.getAccessToken).
         payload = {
             "code": code,
             "checksum": checksum,
+            "uid": userid,
         }
 
         # Convert payload to jData format
@@ -53,7 +57,10 @@ def authenticate_broker(code):
         # Handle the response
         if response.status_code == 200:
             data = response.json()
-            if data.get("stat") == "Ok" and "access_token" in data:
+            # GenAcsTok success returns the access_token; it is not guaranteed
+            # to carry stat == "Ok", so key the success on access_token presence
+            # (matches Finvasia's NorenRestApiOAuth.getAccessToken).
+            if "access_token" in data:
                 logger.info("Shoonya authentication successful")
                 return data["access_token"], None
             else:


Kalaiviswa:
Thanks for the detailed write-up — the NorenRestApiOAuth.getAccessToken citation is accurate and the reproduction steps were useful.

I don't think uid is what was failing, though. Shoonya's docs list only code and checksum as mandatory jData fields for GenAcsTok; Zebu runs the identical Noren flow with those two fields and works in production; and Shoonya OAuth login was verified working on current main on 5 August, with the payload unchanged since April.

On the second change, the docs do list stat in the response table and the sample success response includes "stat": "Ok", so the existing check isn't a false-negative risk. A failure response carries neither field, so both forms behave the same.

INVALID_VERIFIER usually comes from the wrong value in the client_id position of BROKER_API_KEY (it's userid:::client_id), a stale BROKER_API_SECRET, or a reused auth code — Shoonya's codes are single-use, so a refresh on the callback replays a spent one and returns exactly this error.

Closing for now, but please reopen if you can reproduce it on current main with a freshly issued code. Appreciate the contribution.

PR #1592, opened 2026-07-02
Use time.perf_counter() instead of time.time() for relative time calculations

MananGandhi1810:
Currently, the project uses `time.time()` for storing and benchmarking time elapsed between 2 events. But, `time.time()` is not monotonic, i.e. in case of an NTP sync or changes to the system clock, the output of `time.time()` does not come out to be accurate.

Instead,  `time.perf_counter()` uses a monotonic clock, and is resistant to NTP syncs or any other changes in the system time, helping to make the time/latency benchmarks more accurate.

This PR only changes the code where relative measurements of time is use, and not for code where `time.time()` actually stores the timestamp, since `time.perf_counter()` by itself, has no value. It is only useful when used to actually calculate difference in time between 2 events.

Since I do not have subscriptions to all supported broker APIs, I will request you to once verify the changes by running the code with the broker APIs where code has been updated.

Thanks,
Manan

---
## Summary by cubic
Replaced relative timing code to use `time.perf_counter()` for monotonic elapsed-time calculations. This removes drift from system clock/NTP changes and improves accuracy of latency logs, connection waits, and benchmarks across the app.

- **Refactors**
  - Switched to `time.perf_counter()` in WebSocket connect/auth loops and backoff timers across broker clients and `services/websocket_client`.
  - Updated performance/latency tracking in cache restoration, DB init, `utils/latency_monitor`, `utils/httpx_client` (`httpx` hooks), `utils/traffic_logger`, benchmarks, and tests.
  - Kept `time.time()` where absolute timestamps are needed; only elapsed-time math changed.
  - Reverted prior `start.sh` change; no script changes included.

<sup>Written for commit f39ceac8168e1aa54adcdde2ff8c20a83058259e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1592?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app.py
+++ b/app.py
@@ -674,7 +674,7 @@ def setup_environment(app):
                 ("Strategy Portfolio DB", ensure_strategy_portfolio_tables_exists),
             ]
 
-            db_init_start = time.time()
+            db_init_start = time.perf_counter()
             with ThreadPoolExecutor(max_workers=15) as executor:
                 futures = {executor.submit(func): name for name, func in db_init_functions}
                 for future in as_completed(futures):
@@ -684,7 +684,7 @@ def setup_environment(app):
                     except Exception as e:
                         logger.error(f"Failed to initialize {db_name}: {e}")
 
-            db_init_time = (time.time() - db_init_start) * 1000
+            db_init_time = (time.perf_counter() - db_init_start) * 1000
             logger.debug(f"All databases initialized in parallel ({db_init_time:.0f}ms)")
 
             # Signal that DB tables are ready (unblocks cache restoration)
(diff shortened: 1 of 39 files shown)


sbt987:
Since the time is measured in ms in a bunch of places, PLEASE use time.perf_counter_ns()

PR #1595, opened 2026-07-03
fix(angel): guard non-JSON responses so a rate-limit can't crash the API

linuxsmiths:
## Problem

Angel returns a **non-JSON** body under an HTTP 403 rate-limit:

```
Access denied because of exceeding access rate
```

Several Angel API call sites parse the body with `json.loads(response.text)` / `response.json()` **without checking the status**, so a rate-limit raises a `JSONDecodeError`:

- **`funds.py::get_margin_data`** — surfaced by `funds_service` as HTTP 500 `{"status":"error","message":"Expecting value: line 1 column 1 (char 0)"}`. The 403 + rate-limit message are lost. (This is the one that most visibly bites, since funds/`getRMS` is hit several times around login.)
- **`order_api.py::place_order_api`** (`.json()`), **`cancel_order`**, **`modify_order`** — crash the order placement / cancel / modify flows.

The read helper `get_api_response()` **already** handles this rate-limit (it retries on "exceeding access rate"), and the holdings/positions/orderbook/tradebook readers all go through it; quotes (`data.py`) already guard their parse. Only the four sites above were left unguarded.

## Fix

Guard each unguarded parse:

- **Reads** (`get_margin_data`) → return `{}` on a non-JSON body, consistent with the function's existing empty-data path.
- **Order writes** (`place_order_api` / `cancel_order` / `modify_order`) → return a clean `{"status":"error", ...}` envelope via each function's existing error path.

**Writes intentionally do NOT retry.** A `placeOrder` that reached the exchange *before* the error must never be auto-resent — retrying risks a **double order**. So writes fail cleanly and let the caller's existing error handling take over, rather than being routed through the retrying `get_api_response`.

## Validation

Simulated a non-JSON 403 (`"Access denied because of exceeding access rate"`) at each site:

- `get_margin_data` → `{}` (no exception)
- `place_order_api` → `(response, {"status": False, ...}, orderid=None)` (no exception, no order)
- `cancel_order` / `modify_order` → `({"status":"error","message":"Invalid JSON response (HTTP 403)"}, 403)` (no exception)

Valid-JSON happy paths are unchanged.

---
## Summary by cubic
Guard non-JSON Angel responses (403 rate-limit, WAF/IP blocks) in login, funds, and order APIs to prevent JSON decode crashes. Reads return `{}`; writes return clean errors without retry to avoid double orders; login returns a clear gateway-block message.

- **Bug Fixes**
  - `auth_api.authenticate_broker`: handle non-JSON (403/404/WAF); return actionable gateway-block message instead of JSON decode error.
  - `funds.get_margin_data`: on non-JSON, return `{}`; log status/text.
  - `order_api.place_order_api`: guard `.json()`, return `{"status": False, "message": "Invalid JSON response (HTTP status)"}`; no retry.
  - `order_api.cancel_order` / `modify_order`: guard parse, return `({"status":"error","message":"Invalid JSON response (HTTP status)"}, status)`; no retry.

<sup>Written for commit 1d262d3b92c0a4dcf5a3c6c278e75fcadba14a9d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1595?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/broker/angel/api/auth_api.py
+++ b/broker/angel/api/auth_api.py
@@ -38,7 +38,33 @@ def authenticate_broker(clientcode, broker_pin, totp_code):
         response.status = response.status_code
 
         data = response.text
-        data_dict = json.loads(data)
+        # Angel can return a NON-JSON body on some errors — e.g. an HTTP 403
+        # rate-limit ("Access denied because of exceeding access rate"), or an F5
+        # Web Application Firewall page ("Request Rejected … Your support ID is: …")
+        # / a bare "not found" (HTTP 404) when the gateway blocks the caller's IP. A
+        # bare json.loads() on that raises JSONDecodeError, which the caller surfaces
+        # as the cryptic "Expecting value: line 1 column 1 (char 0)" — making a
+        # broker-side block look like a credentials bug. Guard the parse and return a
+        # clear, actionable message instead.
+        try:
+            data_dict = json.loads(data)
+        except json.JSONDecodeError:
+            body = (data or "").strip()
+            waf = "request rejected" in body.lower() or "support id" in body.lower()
+            if waf or response.status_code in (403, 404):
+                msg = (
+                    f"Angel rejected the login request (HTTP {response.status_code}) with a "
+                    "non-JSON response — this is a gateway block (WAF / rate-limit / IP block), "
+                    "not a credentials error. The server's public IP may be temporarily blocked "
+                    "(often from repeated automated logins); try again from a different IP, the "
+                    "block usually clears within a few hours."
+                )
+            else:
+                msg = (
+                    f"Angel returned a non-JSON login response (HTTP {response.status_code}): "
+                    f"{body[:200]}"
+                )
+            return None, None, msg
 
         if "data" in data_dict and "jwtToken" in data_dict["data"]:
             # Return both JWT token and feed token if available (None if not)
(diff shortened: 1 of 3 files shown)


linuxsmiths:
Extended this PR to also guard the **login** path (`authenticate_broker` in `broker/angel/api/auth_api.py`), which still did a bare `json.loads()` on the response.

In addition to the HTTP 403 rate-limit already handled elsewhere, Angel's gateway can return a non-JSON body at login when it blocks the caller's IP — an F5 WAF page (`Request Rejected … Your support ID is: …`) or a bare `not found` (HTTP 404). That raised `JSONDecodeError` and surfaced as the cryptic `Expecting value: line 1 column 1 (char 0)`, making a gateway/IP block look like a credentials error. The login parse is now guarded and returns a clear, actionable message distinguishing a gateway block (WAF / rate-limit / IP) from a credentials failure.

PR #1599, opened 2026-07-03
Compute daily, weekly, monthly, quarterly and yearly candles from minute candles

MananGandhi1810:
Extends Historify's on-the-fly aggregation to support D, W, M, Q, and Y intervals using stored minute (1m) data as the source, in addition to the existing intraday computed intervals.

---
## Summary by cubic
Adds on-the-fly aggregation for daily, weekly, monthly, quarterly, and yearly candles using stored `D` or fallback `1m` data. Prefers stored `D` for performance and computes from `1m` when `D` is missing.

- **New Features**
  - Added support for `D`, `W`, `M`, `Q`, `Y` as computed intervals (`D` mapped to 1440 minutes).
  - `get_ohlcv` now returns stored `1m`/`D` data when present, and computes `D` from `1m` if missing.

- **Refactors**
  - Introduced `_query_stored_ohlcv` and routed storage-interval reads through it; unknown intervals now use this as a last-resort path.
  - Reworked `_get_daily_aggregated_ohlcv` to prefer stored `D` via `data_catalog`, falling back to `1m` via a CTE; applies range filters to both paths with consistent IST-offset grouping into `W/M/Q/Y`.
  - Split interval sets into `INTRADAY_COMPUTED_INTERVALS` and `DAILY_COMPUTED_INTERVALS`, and updated exporters to use the intraday set.

<sup>Written for commit 46c5bd80f318c88a3414d91bea07770263cbbdd5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1599?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/database/historify_db.py
+++ b/database/historify_db.py
@@ -652,8 +652,14 @@ def upsert_market_data(df: pd.DataFrame, symbol: str, exchange: str, interval: s
 # Storage intervals - only these are physically stored
 STORAGE_INTERVALS = {"1m", "D"}
 
-# Standard computed intervals - these are aggregated from 1m data on-the-fly
-COMPUTED_INTERVALS = {"5m", "15m", "30m", "1h", "D", "W", "M", "Q", "Y"}
+# Standard intraday computed intervals - aggregated from 1m data on-the-fly
+INTRADAY_COMPUTED_INTERVALS = {"5m", "15m", "30m", "1h"}
+
+# Standard daily-aggregated intervals - aggregated from D data on-the-fly
+DAILY_COMPUTED_INTERVALS = {"W", "M", "Q", "Y"}
+
+# All standard computed intervals
+COMPUTED_INTERVALS = INTRADAY_COMPUTED_INTERVALS | DAILY_COMPUTED_INTERVALS
 
 # Interval to minutes mapping for standard intervals
 INTERVAL_MINUTES = {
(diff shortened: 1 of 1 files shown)


marketcalls:
Thanks @MananGandhi1810 for the PR and the detailed writeup.

There's a data-correctness concern with computing `D`/`W`/`M`/`Q`/`Y` from stored `1m` data (the `1m` fallback path added here): **a daily candle is not just the roll-up of that day's minute bars.** For F&O especially, the official daily close is the exchange **settlement price** (e.g. the last-half-hour VWAP for futures), not the 15:29 last-traded price you'd get by aggregating `1m` closes. Even for cash equities, the daily close comes from the closing session and can differ from the last `1m` close. So a `D` candle synthesized from `1m` will not match the broker's actual daily candle — and every higher timeframe built on top of it (`W`/`M`/`Q`/`Y`) inherits that error.

Historify already handles the daily-and-above case correctly today: `D` is downloaded and stored directly from the broker (settlement-adjusted closes intact), and `W`/`M`/`Q`/`Y` are aggregated from that stored `D` (`is_daily_aggregated_interval` → `_get_daily_aggregated_ohlcv`). Aggregation from `1m` is best kept to the intraday timeframes (`5m`, `15m`, `30m`, `1h`, and custom `Nm`/`Nh`), where rolling up minute bars is exact.

Since the daily+ timeframes are already covered from the stored daily base, I'd suggest we **close this PR** rather than introduce a `1m`→`D` computation that can yield inaccurate daily and higher-timeframe candles. Genuinely appreciate the effort though — thank you!

MananGandhi1810:
Oh, learnt something new today! Thanks, closing the PR

PR #1609, opened 2026-07-08
Add FXMacroData release calendar example

roberttidball:
## Summary
- Add a dependency-free example for using the public FXMacroData release calendar in backtesting workflows.
- Build a set of confirmed tier-one USD macro blackout dates.
- Show where to skip entries, reduce sizing, or add an event-risk feature in the framework.

## Validation
- Ran `python -m py_compile` on the added Python examples where applicable.
- Verified the helper against the live public calendar endpoint for `2026-07-01` to `2026-07-20`; it returned 12 events and tier-one blackout dates for `2026-07-14` and `2026-07-16`.

---
## Summary by cubic
Add a dependency‑free Python example that pulls the public FXMacroData release calendar and builds confirmed tier‑one USD blackout dates to gate order placement around scheduled events. Provides fetch_release_events, build_blackout_dates, and can_place_order helpers for simple date filtering in backtests.

<sup>Written for commit 49a9e2bb380a5a0426bb885560595ecfe48b70f5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1609?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/examples/python/fxmacrodata_release_calendar_filter.py
+++ b/examples/python/fxmacrodata_release_calendar_filter.py
@@ -1,0 +1,74 @@
+"""Use FXMacroData release-calendar dates before placing OpenAlgo orders.
+
+This example loads confirmed tier-one USD macro events from the public
+FXMacroData calendar endpoint and exposes a small date filter. A strategy can
+check the filter before placing new orders around scheduled event risk.
+"""
+
+from datetime import date
+from datetime import datetime
+from datetime import timedelta
+from datetime import timezone
+import json
+from typing import Any, Dict, Iterable, List, Set
+from urllib.parse import urlencode
+from urllib.request import urlopen
+
+
+FXMD_CALENDAR_URL = "https://fxmacrodata.com/api/v1/calendar/{currency}"
+
+
+def fetch_release_events(
+    currency: str,
+    start_date: str,
+    end_date: str,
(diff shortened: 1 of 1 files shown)


marketcalls:
Thanks for putting this together, and apologies for the slow response.

We do not have immediate plans to bring FX macro data into OpenAlgo, so we are not in a position to take on an example that depends on an external FXMacroData calendar endpoint. Carrying it in `examples/` would imply a level of support for that data source that we cannot commit to maintaining right now, and an example that breaks when a third-party endpoint changes is worse for users than no example at all.

Closing on those grounds rather than on the quality of the work. The helper structure itself (fetch, build blackout dates, gate order placement) is sound. If FX macro or event-risk data lands on the roadmap later, this is a reasonable starting point and we would be happy to revisit it.

If you would like to keep it available in the meantime, hosting it in your own repository and linking it from a discussion thread works well for community examples.

PR #1621, opened 2026-07-10
fix: expose pre-premium-credit margin total in Zerodha basket response

### LINKED ISSUE #1620: Zerodha margin(): final.total nets premium credit, causing pre-trade sizing to underestimate required funds
## Summary

`parse_margin_response()` in `broker/zerodha/mapping/margin_data.py` returns `total_margin_required` as Zerodha's basket `final.total` — which the function's own inline docs confirm already nets out the option premium you'd collect from selling the legs (`initial.total - final.total = option_premium`, see the comment block around line 118-136).

For strategies that use `client.margin()` ahead of time to size a position (e.g. "how many lots can I afford"), this understates the margin actually required to *place* the order — because that premium credit isn't available until the short legs actually fill. The pre-trade estimate is optimistic by roughly the premium amount.

## Observed impact

Sizing a 4-leg NIFTY iron fly (long call/put hedges + short ATM call/put) using `client.margin()` for a single lot and linearly scaling:

- Estimated margin: ~68,339/lot → sized to 12 lots (well within available funds using a 95% utilization buffer)
- Actual order-time rejection from Zerodha: `"Insufficient funds. Margin required: 951,448.81. Margin available: 862,850.65."` — that's ~79,287/lot, about 16% higher than the pre-trade estimate

The gap lines up almost exactly with the premium-credit netting described above.

## Proposed fix

`parse_margin_response()` already computes `initial.total` and `option_premium` internally for logging (see the "ZERODHA BASKET MARGIN - DETAILED BREAKDOWN" log block) but discards them before returning. Exposing them would let callers size conservatively:

```python
"data": {
    "total_margin_required": total_margin_required,   # unchanged
    "span_margin": span_margin,                          # unchanged
    "exposure_margin": exposure_margin,                   # unchanged
    "initial_total_margin": initial_total,                # NEW - pre-premium-credit total
    "option_premium_credit": option_premium,              # NEW - credit final.total already netted
}
```

This is purely additive — no change to existing fields or behavior for current consumers.

Happy to submit a PR for this (plus a test in `test/`) if that's welcome.

arsalanansari17:
## What this does

`parse_margin_response()` in `broker/zerodha/mapping/margin_data.py` computes both `initial.total` and `final.total` from Zerodha's basket margin API, but only ever returned `final.total` as `total_margin_required` — silently discarding the pre-premium-credit figure.

`final.total` already nets out the option premium you'd collect from selling the short legs (see the existing inline comment: `initial.total - final.total = option_premium`). That premium isn't actually available until the sell orders fill, so anyone using `client.margin()` to size a position *before* placing orders is working off an optimistic number.

Closes #1620.

## Change

Purely additive — no existing field changes shape or value:

```python
"data": {
    "total_margin_required": total_margin_required,   # unchanged
    "span_margin": span_margin,                          # unchanged
    "exposure_margin": exposure_margin,                   # unchanged
    "initial_total_margin": initial_total_margin,        # NEW
    "option_premium_credit": option_premium_credit,      # NEW
}
```

For non-basket (single/aggregated order) responses, `initial_total_margin` falls back to `total_margin_required` since there's no initial/final split in that path.

## Note for maintainers

While looking into this I noticed Shoonya, Flattrade, and Firstock's margin mappings each have a comment claiming Zerodha's `total_margin_required` maps to `initial.total` (the conservative figure) — e.g. Shoonya: *"Parallels Zerodha's initial.total vs final.total... matches Zerodha impl using initial.total"*. That's not what the current Zerodha code does (it uses `final.total`), so those three brokers already behave conservatively while Zerodha doesn't. I kept this PR additive/non-breaking rather than changing `total_margin_required`'s existing meaning, but flagging the inconsistency in case it's worth a follow-up decision on your end.

## Testing

- New `test/test_zerodha_margin_api.py` — 3 tests covering the basket-response case, the non-basket fallback, and error passthrough. All pass.
- `uv run pytest test/test_zerodha_margin_api.py test/test_dhan_margin_api.py -v` — 8/8 pass, no regressions in the sibling Dhan margin tests.
- `uv run ruff check` / `uv run ruff format --check` — clean.

---
## Summary by cubic
Expose `initial_total_margin` in Zerodha basket margin responses so clients can size positions conservatively before fills. `total_margin_required` is unchanged; dropped the misleading `option_premium_credit`.

- **Bug Fixes**
  - Zerodha: `parse_margin_response()` now returns `initial_total_margin` (pre-premium-credit) alongside `total_margin_required`.
  - Non-basket responses set `initial_total_margin` = `total_margin_required`.
  - Dropped `option_premium_credit` due to inconsistent semantics in real responses.
  - Added tests in `test/test_zerodha_margin_api.py`; no changes to existing field shapes or values.

<sup>Written for commit d48461e4ae30264aff3d5af189dbdf1b16d4f57c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1621?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/broker/zerodha/mapping/margin_data.py
+++ b/broker/zerodha/mapping/margin_data.py
@@ -163,6 +163,8 @@ def parse_margin_response(response_data):
         span_margin = 0
         exposure_margin = 0
         margin_benefit = 0
+        initial_total_margin = 0
+        option_premium_credit = 0
 
         if isinstance(data, dict) and "final" in data:
             # Basket response - use final values which include spread benefit
@@ -193,6 +195,14 @@ def parse_margin_response(response_data):
             initial_option_premium = initial.get("option_premium", 0)
 
             final_total = final.get("total", 0)
+
+            # initial.total is the margin required BEFORE the premium credit from
+            # selling the short legs is netted in — that credit only lands once the
+            # sell orders actually fill, so it's the safer figure for callers sizing
+            # a position ahead of placing orders (see total_margin_required's docstring
+            # note above: initial.total - final.total = option_premium).
+            initial_total_margin = initial_total
+            option_premium_credit = option_premium
 
             # Calculate margin benefit (savings from spread/hedge recognition)
             # Formula: Margin Benefit = Sum of Individual Margins - Optimized Combined Margin
@@ -274,6 +284,12 @@ def parse_margin_response(response_data):
                 f"Orders margin: total={total_margin_required}, span={span_margin}, exposure={exposure_margin}"
             )
 
+        # Non-basket (single/aggregated orders) responses have no separate
+        # pre-optimization figure — total_margin_required is already the
+        # standalone total, so it doubles as initial_total_margin here.
+        if not (isinstance(data, dict) and "final" in data):
+            initial_total_margin = total_margin_required
+
         # Return standardized format matching OpenAlgo API specification
         response_data = {
             "status": "success",
@@ -281,6 +297,8 @@ def parse_margin_response(response_data):
                 "total_margin_required": total_margin_required,
                 "span_margin": span_margin,
                 "exposure_margin": exposure_margin,
+                "initial_total_margin": initial_total_margin,
+                "option_premium_credit": option_premium_credit,
             },
         }
 
--- a/broker/zerodha/mapping/margin_data.py
+++ b/broker/zerodha/mapping/margin_data.py
@@ -164,7 +164,6 @@ def parse_margin_response(response_data):
         exposure_margin = 0
         margin_benefit = 0
         initial_total_margin = 0
-        option_premium_credit = 0
 
         if isinstance(data, dict) and "final" in data:
             # Basket response - use final values which include spread benefit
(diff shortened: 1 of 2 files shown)


arsalanansari17:
Thanks for catching this — confirmed against a live basket margin response (illustrative numbers below, same relationships as the real data):

```
initial.total=80,000  initial.option_premium=+4,000
final.total  =68,000  final.option_premium  =-8,000
actual margin_benefit (initial.total - final.total) = 12,000
```

Neither `initial.option_premium` (+4,000) nor `final.option_premium` (-8,000) equals the actual credit (12,000) — the real "credit" is the *delta* between them (span/exposure barely move between initial and final for these strategies, so essentially the whole optimization benefit shows up as the option_premium component swinging from positive to negative). So switching the source to `initial.option_premium` would still be wrong, just wrong in a different direction.

Rather than add a derived-delta field that just restates `margin_benefit` under a new name, I've dropped `option_premium_credit` entirely in d48461e4 — `initial_total_margin` (the field that actually matters for pre-trade sizing) is unaffected and already verified correct against real data.

PR #1628, opened 2026-07-12
fix(sandbox): scale simulated margin by contract_value to match P&L ledger

punksterlabs:
## Summary

Sandbox realized/unrealized P&L is scaled by a per-symbol `contract_value` multiplier, but the margin calculation ignores it. For instruments where one contract is a fraction of the underlying (e.g. crypto derivatives on Delta Exchange — `0.001` BTC or `0.01` ETH per contract), margin is over-blocked by a factor of `1/contract_value` (~1000× for BTC, ~100× for ETH). This inflates `used_margin` and understates available cash while P&L stays correct, leaving the funds ledger inconsistent with the position ledger.

## Root cause

`PositionManager` already scales P&L by `contract_value`:

- `PositionManager._get_contract_value()` looks up `symtoken.contract_value`
- `_calculate_position_pnl()` / `_calculate_realized_pnl()` multiply by it

But `FundManager.calculate_margin_required()` computed:

```python
trade_value = quantity * price          # contract_value ignored
margin = trade_value / leverage
```

So P&L and margin end up on different scales for fractional-contract instruments.

### Observed (Delta Exchange, sandbox)

A `BTC…PE` position of 4 contracts @ premium 123.4 (`contract_value = 0.001`):

| Field | Before | Correct |
|---|---|---|
| Position P&L (unrealized) | `0.12` ✅ | `0.12` |
| `margin_blocked` / `used_margin` | `526.53` ❌ | `~0.49` |

The premium is quoted per unit of underlying, so 4 contracts = `0.004` BTC of exposure; the real premium outlay is `123.4 × 4 × 0.001 ≈ $0.49`, not `$526`.

## Fix

Multiply `trade_value` by `contract_value` in `calculate_margin_required()`. It is `1.0` for equities and standard F&O, so this is a **no-op** for those instruments — only fractional-contract instruments change. All margin flow (block/release, proportional release on partial close, `position.margin_blocked`) derives from this one function, so the fix is localized.

```python
trade_value = quantity * price * contract_value
margin = trade_value / leverage
```

## Tests

`test/sandbox/test_margin_contract_value.py` (mock-based, no DB) covers:

- Equity (`contract_value = 1.0`) — margin unchanged
- BTC option (`0.001`) — margin scaled 1000×
- ETH option (`0.01`) — margin scaled 100×
- Leverage still applied on top of the scaled notional (crypto futures)

## Compatibility

No behavior change for NSE/BSE/NFO/BFO/MCX/CDS instruments (all `contract_value = 1.0`). Only affects instruments that already rely on `contract_value` scaling for P&L, bringing their margin onto the same basis.

---
## Summary by cubic
Scale sandbox margin by per-symbol `contract_value` and harden multiplier handling so margin and P&L stay in sync across all paths, including cold-cache DB fallback. Fixes over-blocked margin for fractional-contract instruments; no change for equities and standard F&O.

- **Bug Fixes**
  - Apply `contract_value` in `FundManager.calculate_margin_required`; keep leverage on the scaled notional.
  - Share `utils/symbol_utils.normalize_contract_value` with `PositionManager` so P&L and margin use the same, validated multiplier; fallback to `1.0` for missing/negative/NaN.
  - Include `contract_value` in DB-fallback `get_symbol_info_dbquery` to prevent unscaled margin/P&L when cache is cold.
  - Expand tests: equity (`1.0`) unchanged, BTC (`0.001`) and ETH (`0.01`) scaled, leverage on top, and bad-multiplier safety.

<sup>Written for commit 2dda8e9879e96fa307b1aa54c72720ae4c2a3e14. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1628?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/database/token_db_enhanced.py
+++ b/database/token_db_enhanced.py
@@ -894,6 +894,10 @@ def get_symbol_info_dbquery(symbol: str, exchange: str) -> SymbolData | None:
                 lotsize=sym_token.lotsize,
                 instrumenttype=sym_token.instrumenttype,
                 tick_size=sym_token.tick_size,
+                # Carry contract_value so the DB-fallback path (cold/expired cache) scales
+                # margin and P&L identically to the cache path; without it crypto
+                # derivatives silently fall back to an unscaled 1.0 multiplier.
+                contract_value=getattr(sym_token, "contract_value", None),
             )
         else:
             return None
--- a/sandbox/fund_manager.py
+++ b/sandbox/fund_manager.py
@@ -393,8 +393,22 @@ class FundManager:
                 logger.error(f"Symbol {symbol} not found on {exchange}")
                 return None, "Symbol not found"
 
-            # Calculate trade value (quantity × price)
-            trade_value = quantity * price
+            # contract_value scales notional for instruments where one contract is not one
+            # unit of the underlying (e.g. crypto derivatives: 0.001 BTC or 0.01 ETH per
+            # contract). It is 1.0 for equities and standard F&O, so this is a no-op there.
+            # This mirrors the contract_value scaling already applied to realized and
+            # unrealized P&L in position_manager, keeping the margin/funds ledger consistent
+            # with the P&L ledger. Without it, margin is over-blocked by 1/contract_value
+            # (e.g. ~1000x for BTC), inflating used_margin and understating available cash.
+            contract_value = Decimal("1.0")
+            try:
+                if symbol_obj.contract_value and float(symbol_obj.contract_value) != 1.0:
+                    contract_value = Decimal(str(symbol_obj.contract_value))
+            except (TypeError, ValueError):
+                pass
+
+            # Calculate trade value (quantity × price × contract_value)
+            trade_value = quantity * price * contract_value
 
             # Determine leverage based on action, product and symbol type
             leverage = self._get_leverage(exchange, product, symbol, action)
--- a/sandbox/fund_manager.py
+++ b/sandbox/fund_manager.py
@@ -35,7 +35,7 @@ from database.sandbox_db import (
 )
 from database.token_db import get_symbol_info
 from utils.logging import get_logger
-from utils.symbol_utils import is_future, is_option
+from utils.symbol_utils import is_future, is_option, normalize_contract_value
 
 logger = get_logger(__name__)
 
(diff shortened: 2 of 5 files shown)


punksterlabs:
Thanks for the automated reviews — all four points were valid. Addressed in `2dda8e98`:

**1. Validate the multiplier is finite & positive (cubic P1)** — a negative/zero/NaN `contract_value` could otherwise make margin negative or zero (and a negative margin would *credit* `available_balance` via `block_margin()`). The normalizer now returns `1.0` unless the value `is_finite()` and `> 0`.

**2. Shared normalization (cubic P3)** — extracted `normalize_contract_value()` into `utils/symbol_utils.py`, now used by **both** the margin path (`fund_manager`) and the P&L path (`position_manager._get_contract_value`), so the two ledgers can't drift apart from independent validation/fallback changes.

**3. Fallback path dropped `contract_value` (codex P2)** — good catch: `get_symbol_info_dbquery()` built `SymbolData` without `contract_value`, so on a cold/expired cache **both** margin *and* P&L silently unscaled for crypto derivatives. It now carries `contract_value` (matching the cache path at `token_db_enhanced.py:247`).

**4. Sub-cent margins rounding to `0.00` (codex P2)** — real, but it stems from the `Numeric(_, 2)` precision on `SandboxFunds.used_margin` / `SandboxOrders.margin_blocked`, which predates this change and affects any sub-cent value. Widening that precision is a schema migration across existing sandbox DBs, so I've left it out of this correctness fix as a separate follow-up. For real premiums the scaled margins are well above the cent floor (e.g. 4 BTC contracts × 0.001 × ~120 ≈ `0.48`); only near-zero-premium contracts hit the floor.

Tests expanded to cover the normalizer's validation cases and a bad-multiplier margin-safety assertion (`test/sandbox/test_margin_contract_value.py`).

PR #1693, opened 2026-07-29
feat(flow): add indicator, timeframe, and order-update nodes

marketcalls:
## Summary

Adds six Flow node types so strategies can be built on real technical analysis and historical price data, not just live quotes and static conditions:

- **`indicator`** — runs any of the 118 single-symbol `openalgo.ta` functions over a symbol's history. `interval` is free text (not a fixed enum) so it works with any interval the connected broker reports via `/api/v1/intervals`, plus Historify-custom intervals (`2m`, `4m`, `W`, `M`, `Q`) via `source: "db"`. Can nest on another Indicator node's output series (e.g. `SMA(RSI(14), 9)`) via `sourceSeries`.
- **`priorPeriodOhlc`** — previous hour/day/week/month OHLC (e.g. PDH/PDL for a breakout strategy) computed server-side, since Flow JSON has no relative-date arithmetic to express "yesterday" itself.
- **`barOffset`** — OHLCV of the Nth closed bar back at any interval/source. Covers "N bars/hours/days back" lookback with one generic node instead of one per time unit.
- **`varCondition`** — compares any two interpolated values (an indicator output, a prior-period level, a workflow variable, or a literal) for branching. Closes the gap explicitly flagged in `docs/prompt/flow-import-format.md` §8.5 ("wait for a varCondition node").
- **`orderUpdateTrigger`** — fires a workflow the moment a matching order's status changes (filled/rejected/cancelled/...), via the existing `order.update` EventBus topic (`FlowOrderUpdateMonitor`, mirrors `FlowPriceMonitor`'s shape but push-based instead of polling) — instead of polling `getOrderStatus` in a loop.
- **`whatsappAlert`** — WhatsApp notification node mirroring the existing `telegramAlert` node.

## Bugs found and fixed via live end-to-end testing

Both of these were caught by executing real workflows against a running instance with a live Zerodha connection (sandbox/analyze mode), not by unit tests alone:

1. **Epoch-vs-ISO timestamp mismatch.** Broker history records return `timestamp` as epoch seconds (confirmed for Zerodha), not an ISO string. The `previous_day`/`previous_hour` "exclude the still-forming bar" filters and the `previous_week`/`previous_month` pandas period bucketing were comparing against string-sliced epoch integers, so `previous_day` was silently returning **today's** OHLC instead of yesterday's. Fixed with a shared `_parse_timestamp`/`_timestamp_series` normalizer in `services/indicator_service.py`.
2. **Nested indicator resolution always failed.** `sourceSeries` resolution went through `interpolate()` (which stringifies via Python's `str()`/repr — single-quoted) followed by `json.loads()`, which fails on every real list/dict value. Added `WorkflowContext.resolve_raw()` to return the actual stored Python object for a single `{{path}}` token instead of round-tripping through a lossy string conversion.

## Test plan

- [x] `uv run ruff check` / `ruff format` clean on all new/touched backend files (pre-existing warnings in untouched lines left as-is)
- [x] `npx biome lint ./src` clean on all new/touched frontend files (4 pre-existing warnings elsewhere, unrelated)
- [x] `tsc -b` / `npm run build` — full production build succeeds, no type errors
- [x] All 6 new nodes render correctly in the Flow Editor palette (Trigger/Data/Logic/Util tabs) and their ConfigPanel sections bind live to canvas node display
- [x] Live execution against a running instance (Zerodha, sandbox/analyze mode) via the real `/flow/api/workflows` API, covering:
  - `indicator` (RSI over RELIANCE, real value returned)
  - `indicator` nested mode (`SMA(RSI)`, confirmed working after the `resolve_raw` fix)
  - `priorPeriodOhlc` for all four periods (hour/day/week/month), confirmed correct after the timestamp fix
  - `barOffset`
  - `varCondition` — both true and false branches
  - `orderUpdateTrigger` — full activate → registers watch in `/flow/api/monitor/status` → deactivate → cleanly removed
- [x] Test workflows cleaned up (deleted) after verification
- [ ] `whatsappAlert` live send — not exercised (requires WhatsApp bot pairing not set up in this environment); code path mirrors the already-proven `telegramAlert` pattern


marketcalls:
## Review fixes pushed (`786500eda`)

Validated all 18 findings from Codex and cubic against source — **every one was real**. All are fixed, plus one further bug the fixes surfaced and one rate-limit issue.

### P0 — order-update triggers could never fire
Both reviewers were right, and this was the important catch: `execute_workflow`'s `trigger_types` omitted `orderUpdateTrigger`, so every matching event died on `No trigger node found` before reaching a downstream node. My original testing only covered activate/deactivate *registration*, never a firing event — which is exactly how it slipped through. Now verified end-to-end (workflow executes past the trigger).

**Account scoping** also fixed: watches now carry the owner (resolved from the workflow's API key) and compare against `event.request_data["user_id"]`. Enforced only when the event carries an identity — some sandbox paths omit `user_id`, and failing closed there would silently kill analyze-mode triggers. Verified: another account's fill no longer matches; own account still does.

### One further bug the fix surfaced
Returning `condition: False` on an errored `varCondition` still routed the **false** branch — so on `RSI < 70 → BUY else SELL`, a typo'd variable would have fired the **SELL**. An errored condition now takes *neither* branch, while pass-through edges still run so the failure stays visible. Verified live: neither trading branch runs, the log alert does.

### Rate limiting
Worth flagging separately, since the node graph hides it. Broker data APIs are tight — Dhan 5 req/s history and **1 req/s quotes** (error 805 on breach), Flattrade 10 req/s, some Zerodha paths 1 req/s — on top of `history_service`'s process-wide ~350ms gate. The three new data nodes each fetched independently, so `RSI + SMA + ATR + prior-day levels` on one symbol issued **four identical requests per run**. They now share a bounded short-TTL history cache (`cachetools.TTLCache`, as already used in `database/flow_db.py`), keyed by the exact request. Errors and empty ranges are never cached so they stay retryable. Tunable via `FLOW_HISTORY_CACHE_TTL` / `FLOW_HISTORY_CACHE_MAXSIZE`. Measured: 4 identical node fetches → 1 broker call; distinct symbols still fetch; 3 failing fetches → 3 calls.

### Verification
- 118/118 indicators pass after all changes
- Concurrency: 50 simultaneous events on a `once` watch → fires **exactly 1×** (was a real race — claim now happens under the lock)
- Restart restore: watch reinstated from DB on boot with owner resolved (`Restored 1 order-update watch(es)`)
- `400` not `500` for missing filter / `{{variable}}` order id / unknown status
- Delete tears down the watch — 0 stale registrations after cleanup
- `previous_hour`/`previous_day` raise instead of returning today's forming candle
- Aware ISO timestamps normalize to naive (no more `TypeError` vs `datetime.now()`)
- `trigger_pending` ↔ `trigger pending` now match
- Full frontend build + biome lint clean; ruff clean on all touched files (the few remaining `UP035`/`F841`/`C408` findings are pre-existing on `main`, confirmed by stashing)

Not covered: `whatsappAlert` live send (needs bot pairing, unavailable here) — code path mirrors the proven `telegramAlert`.

marketcalls:
## Second-round review fixes pushed (`91343c9e6`)

All five new findings validated — **every one was real**, including one that was a regression I introduced in the bar-ceiling commit.

### The regression (P2, `indicator_service.py`)
Replacing `df.tail(lookback + 50)` with `df.tail(clamp_bars(lookback))` silently dropped the 50 warm-up rows. EMA/MACD/ATR are recursive, so starting the recursion exactly at the requested window biased the reported value. Restored (still bounded by `MAX_HISTORY_BARS`; the rows are already in the fetched response so it costs no extra broker call).

Verified against `ta.ema` over the full series — absolute error now **2e-6**, and `bars_used` is 150 for `lookbackBars: 100` as intended.

### P1 — cache key had no account discriminator
Fixed. The key now includes a truncated SHA-256 of the client's API key — a hash rather than the key itself, so the discriminator never lands in a log line. Verified: account A's second identical request is a cache hit; account B still fetches its own.

### P1 — watches could become account-unscoped
Fixed, and I've reversed my earlier judgement here. I originally failed *open* when identity was unknown, reasoning that sandbox might omit `user_id` and failing closed would silently kill analyze-mode triggers. That reasoning was wrong: `SandboxOrderManager.__init__` takes `user_id` as a **required positional argument**, so the case doesn't arise and the leniency was pure risk. Now:
- registration is refused when the owner can't be resolved from the API key
- an event carrying no identity no longer satisfies an owner-scoped watch

Verified: no-identity event → no match; other account → no match; own account → match; unresolvable key → `ValueError` at registration.

### P2 — cache didn't actually single-flight
Correct — the lock was released before `client.get_history()`, so concurrent misses each hit the broker and defeated the request-collapse the cache exists for. Added a per-key lock. Verified: **12 concurrent identical requests → 1 broker call**, and the in-flight map drains afterwards.

### P3 — `code` field on the `source="db"` guard
Added, matching `_handle_response`'s shape.

### Regression check
- 118/118 indicators pass (now exercised with `offset_bars` and warm-up)
- All 6 executable documentation examples still pass end-to-end against a live broker

### Also in this branch since the last comment
Two **pre-existing** logic-gate bugs surfaced while validating the docs examples, both of which caused **duplicate orders**:

- `7bbc56f25` — gates evaluated on partial inputs. A 2-input AND fired on the first input alone, then again when the second arrived. Reproduced on an EMA crossover; with an order node downstream that is two orders, the first placed before the second condition was even checked.
- `f26bee1cc` — gates re-evaluated once per inbound path. A 4-input AND fired twice after all inputs resolved. Confirmed live: the pre-fix build placed **two** sandbox orders from one signal; the fixed build places exactly one.

Both are on `main` today, independent of this PR's new nodes.

marketcalls:
## Single-flight fix pushed (`efdb65f75`)

Valid finding, and the impact was larger than described — not a couple of duplicate calls but a **full fan-out**.

### Reproduced first
With a broker returning errors, 12 concurrent identical requests produced **12 broker calls**. The previous implementation used a per-key lock and had waiters retry after acquiring it; since error responses are deliberately not cached, every waiter found a cold cache and issued its own call. That is a thundering herd arriving exactly when the broker is already failing or rate limiting — the opposite of what the cache exists for.

### Fix
Replaced the lock with a small flight object holding an `Event` plus the leader's result *or* exception. The leader fetches once and publishes the outcome; waiters block on the event and return that same outcome instead of re-fetching. The flight is retired in a `finally` after the event is set, so a later call starts a fresh attempt and failures stay retryable.

### Verified
| Scenario | Broker calls | Waiters |
|---|---|---|
| 12 concurrent, error response | **1** | all 12 receive the same error result |
| 12 concurrent, raised exception | **1** | all 12 see the exception |
| 12 concurrent, success | **1** | all 12 receive the cached result |
| Later call after a failed flight | increments | not pinned — retry works |

In-flight map drains in every case. A bounded wait (`FLOW_HISTORY_FLIGHT_TIMEOUT`, 60s) guards against a leader thread dying outright.

### Regression
118/118 indicators pass; all 9 documentation examples (6 from the docs site, 3 from the import spec) still execute end-to-end against a live broker.

PR #1694, opened 2026-07-29
feat(flow): per-strategy position book and P&L

marketcalls:
> **Stacked on #1693.** Base is `feat/flow-indicator-timeframe-nodes`, not `main`, because it builds on the Flow node work there and touches the same executor file. Once #1693 merges I'll retarget this to `main` — the diff shown here is only this PR's changes.

## Problem

A Flow workflow can exit on **account-wide** P&L (`{{pb.total_pnl}}`) or **per-symbol** P&L (`{{pb.data[0].pnl}}`), but not on **its own**. Nothing in OpenAlgo attributes a position to the strategy that opened it: the broker nets positions per `(symbol, exchange, product)` and has never heard of the strategy tag. Two strategies trading the same contract are indistinguishable downstream — and if one is long while the other is short, the position book shows the net and neither strategy's P&L is recoverable.

So "exit when *this strategy* is up 5000" was not expressible.

## Approach

A parallel book keyed by strategy, **fed entirely from the event bus**:

- `order.placed` carries the strategy tag and supplies the `orderid → strategy` mapping. This is the only moment that tag is known — it never round-trips through the broker.
- `order.update` supplies fills, which are booked against that strategy's position.

The order execution path is not modified at all. Two useful consequences: orders placed through `/api/v1` are tracked exactly like Flow-placed ones, and both live and analyze modes work without special-casing.

## Correctness

**Idempotent.** Applied quantity is tracked per order and only the unseen delta is booked, so a broker feed and a postback reporting the same fill cannot double-count. Partial fills fall out of the same mechanism rather than needing separate handling.

**Weighted-average cost**, matching how OpenAlgo and Indian brokers report `average_price`. A position flipping through zero realizes the closed leg and reopens the remainder at the fill price.

**Unrealized P&L is deliberately not stored** — it is a function of a price that changes continuously, so a stored value is stale the instant it is written. It is computed at read time from a single position-book call.

**`today_realized` resets** on the first fill of a new trading date, keeping it aligned with the ~3 AM IST session rollover without needing a scheduler.

## Prerequisite fixed here

Every Flow order previously went out tagged `flow_workflow`, so **all workflows shared one identity** and per-strategy P&L would have been meaningless. Orders are now tagged with the workflow's own name, overridable per node via `strategyTag`. That field is deliberately *not* named `strategy` — `optionsMultiOrder` already uses `strategy` for the structure type (straddle / iron_condor / custom).

## Usage

```
strategyPnl (strategy blank = this workflow)  ->  varCondition {{spnl.total}} >= 5000  ->  closePositions
```

Exposes `realized`, `unrealized`, `total`, `today_realized`, `today_total`, `open_quantity`, and a per-leg breakdown.

## Test plan

- [x] Accounting core unit-tested over 16 cases: long/short round-trips, open positions marked to market, weighted-average scale-in, partial close, flip-through-zero, multi-leg, and two strategies on one symbol
- [x] Event path verified end-to-end: buy 100@100 then sell 40@120 at LTP 130 → realized 800, unrealized 1800, total 2600, open 60
- [x] Idempotency: replaying both fill events changes nothing
- [x] Partial fills: cumulative 30 then 100 books 100, not 130
- [x] Isolation: two strategies on the same contract tracked separately
- [x] Untagged orders ignored (no phantom legs)
- [x] `ruff check` / `ruff format` clean on all new files; frontend `npm run build` clean
- [ ] Not yet exercised: a live (non-analyze) broker fill, and multi-day carry where `realized` accumulates across sessions

## Known limitations

- **Attribution starts when tracking starts.** Positions opened before this ships have no tag, so they are invisible to the book. There is no backfill.
- **`closePositions` squares off everything**, not just one strategy's legs. A per-strategy exit currently needs per-symbol close nodes. Per-strategy closing is a natural follow-up.
- **`compute_strategy_pnl` is not wired into any caller.** It is an independent trade-replay implementation of the same figures, kept for reconciling the book against the broker's own session record. Happy to drop it if reviewers would rather not carry unwired code.

### DIFF
--- a/app.py
+++ b/app.py
@@ -959,6 +959,7 @@ def shutdown_database_sessions(exception=None):
         ("database.market_calendar_db", "db_session"),
         ("database.telegram_db", "db_session"),
         ("database.symbol", "db_session"),
+        ("database.strategy_book_db", "db_session"),
     ]
 
     for module_name, session_attr in _sessions:
--- a/app.py
+++ b/app.py
@@ -411,9 +411,7 @@ def create_app():
                 "can immediately complete OAuth without admin review."
             )
 
-        logger.info(
-            "Remote MCP blueprints registered (OAuth + JSON-RPC dispatch + SSE)."
-        )
+        logger.info("Remote MCP blueprints registered (OAuth + JSON-RPC dispatch + SSE).")
 
     # Exempt webhook endpoints from CSRF protection after app initialization
     with app.app_context():
@@ -484,10 +482,7 @@ def create_app():
         from flask import request
 
         # Static assets don't need DB
-        if (
-            request.path.startswith("/static/")
-            or request.path.startswith("/assets/")
-        ):
+        if request.path.startswith("/static/") or request.path.startswith("/assets/"):
             return
 
         # Wait up to 30s for DB init (typically ~3.5s)
@@ -582,9 +577,15 @@ def create_app():
 
         # Skip tracking for common browser/crawler requests that are not attack probes
         safe_prefixes = (
-            "/favicon", "/robots.txt", "/sitemap", "/manifest",
-            "/sw.js", "/.well-known", "/apple-touch-icon",
-            "/service-worker", "/workbox",
+            "/favicon",
+            "/robots.txt",
+            "/sitemap",
+            "/manifest",
+            "/sw.js",
+            "/.well-known",
+            "/apple-touch-icon",
+            "/service-worker",
+            "/workbox",
         )
 
         if not is_authenticated and not path.startswith(safe_prefixes):
@@ -716,6 +717,20 @@ def setup_environment(app):
             db_init_time = (time.time() - db_init_start) * 1000
             logger.debug(f"All databases initialized in parallel ({db_init_time:.0f}ms)")
 
+            # The strategy book must be listening before any order can be
+            # accepted: order.placed carries the only copy of the strategy tag,
+            # so an order placed before this registration loses its attribution
+            # permanently. Registered ahead of db_ready for that reason.
+            try:
+                from database.strategy_book_db import init_strategy_book_db
+                from subscribers.strategy_book_subscriber import register as register_strategy_book
+                from utils.event_bus import bus as _bus
+
+                init_strategy_book_db()
+                register_strategy_book(_bus)
+            except Exception:
+                logger.exception("Failed to initialize strategy book")
+
             # Signal that DB tables are ready (unblocks cache restoration)
             app.db_ready.set()
 
@@ -740,18 +755,8 @@ def setup_environment(app):
                 )
 
                 restore_order_update_watches()
-            except Exception as e:
-                logger.error(f"Failed to restore Flow order-update watches: {e}")
-
-            try:
-                from database.strategy_book_db import init_strategy_book_db
-                from subscribers.strategy_book_subscriber import register as register_strategy_book
-                from utils.event_bus import bus as _bus
-
-                init_strategy_book_db()
-                register_strategy_book(_bus)
-            except Exception as e:
-                logger.error(f"Failed to initialize strategy book: {e}")
+            except Exception:
+                logger.exception("Failed to restore Flow order-update watches")
 
             try:
                 from services.historify_scheduler_service import init_historify_scheduler
@@ -795,6 +800,7 @@ def setup_environment(app):
                     logger.exception("WhatsApp bot auto-start crashed")
 
             import threading as _threading
+
             _threading.Thread(
                 target=_autostart_whatsapp_bot,
                 daemon=True,
@@ -819,6 +825,7 @@ def setup_environment(app):
 
                     def run_catchup():
                         from sandbox.position_manager import catchup_missed_settlements
+
                         catchup_missed_settlements()
                         return ("catchup_settlement", True, "Completed")
 
@@ -833,14 +840,22 @@ def setup_environment(app):
                                 service_name, success, message = future.result()
                                 if service_name == "execution_engine":
                                     if success:
-                                        logger.debug("Execution engine auto-started (Analyzer mode is ON)")
+                                        logger.debug(
+                                            "Execution engine auto-started (Analyzer mode is ON)"
+                                        )
                                     else:
-                                        logger.warning(f"Failed to auto-start execution engine: {message}")
+                                        logger.warning(
+                                            f"Failed to auto-start execution engine: {message}"
+                                        )
                                 elif service_name == "squareoff_scheduler":
                                     if success:
-                                        logger.debug("Square-off scheduler auto-started (Analyzer mode is ON)")
+                                        logger.debug(
+                                            "Square-off scheduler auto-started (Analyzer mode is ON)"
+                                        )
                                     else:
-                                        logger.warning(f"Failed to auto-start square-off scheduler: {message}")
+                                        logger.warning(
+                                            f"Failed to auto-start square-off scheduler: {message}"
+                                        )
                                 elif service_name == "catchup_settlement":
                                     logger.debug("Catch-up settlement check completed on startup")
                             except Exception as e:
@@ -886,7 +901,9 @@ def setup_environment(app):
                             if success:
                                 success, message = telegram_bot_service.start_bot()
                                 if success:
-                                    logger.debug(f"Telegram bot auto-started successfully: {message}")
+                                    logger.debug(
+                                        f"Telegram bot auto-started successfully: {message}"
+                                    )
                                 else:
                                     logger.error(f"Failed to auto-start Telegram bot: {message}")
                             else:
@@ -921,9 +938,12 @@ def _restore_caches_background():
                 symbol_count = cache_result["symbol_cache"].get("symbols_loaded", 0)
                 auth_count = cache_result["auth_cache"].get("tokens_loaded", 0)
                 if symbol_count > 0 or auth_count > 0:
-                    logger.debug(f"Cache restoration: {symbol_count} symbols, {auth_count} auth tokens")
+                    logger.debug(
+                        f"Cache restoration: {symbol_count} symbols, {auth_count} auth tokens"
+                    )
         except Exception as e:
             logger.debug(f"Cache restoration skipped: {e}")
+
 
 threading.Thread(target=_restore_caches_background, daemon=True).start()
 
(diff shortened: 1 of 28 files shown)


marketcalls:
All five review findings are fixed, plus two issues found while verifying them. Pushed as 8030e7c.

## Review findings

**1. Fill arriving before its order tag (P1).** `bus.publish` dispatches on a ThreadPoolExecutor, so it is asynchronous. `sandbox_place_order` runs at `place_order_service.py:160` and `OrderPlacedEvent` is published at `:165` — in analyze mode an immediately marketable order reports its fill before the tag exists, and the fill was dropped permanently. A `strategy_pending_fills` table now buffers such a fill and `record_order_tag` drains it on arrival.

**2. Batch legs never tagged (P1).** `optionsMultiOrder`, `basketOrder`, `splitOrder` and `optionsOrder` place their legs with `emit_event=False` (`options_multiorder_service.py:165` and `:370`), so no per-leg `order.placed` is published and none of those legs reached the book. Each batch publishes one completion event carrying `strategy` and `results`; `on_batch_completed` now tags every child from it.

**3. Cumulative average price corrupted the cost basis (P1).** Brokers report `average_price` cumulatively. Booking a partial fill at the latest cumulative average is wrong whenever partials execute at different prices. `applied_notional` now tracks booked notional and the incremental price is `(cumulative_notional - applied_notional) / delta`. Test: 5 @ avg 100 then cumulative 10 @ avg 110 now yields an average of **110**, previously 105.

**4. Position book failure returned a healthy zero (P1).** `get_strategy_pnl` valued open legs against an empty book and returned `status: success` with `total: 0` when the broker call failed. An exit trigger would treat that as a real reading. It now returns `status: error`.

**5. Stale `today_realized_pnl` (P2).** `get_strategy_legs` returns `0.0` when the stored `trade_date` is not today.

## Found while verifying these

**FD leak — `db_session` never released.** `database.strategy_book_db` was missing from the `app.py` teardown list, and built its engine with `create_engine()` instead of `engine_factory.create_db_engine()`. Production is a single eventlet worker that never restarts, so this is the same class of leak the teardown list already documents as "previously missing (caused FD leak)". Both fixed; measured flat at **16 FDs over 400 iterations** of the tag/fill/read path.

**Unbounded `strategy_pending_fills` growth.** Most untagged fills are not racing anything — they are ordinary orders placed outside any strategy, whose tag will never arrive. Every one buffered a row that was never removed. Rows are now pruned by age (`STRATEGY_PENDING_FILL_TTL_MIN`, default 10 min).

**`reset_strategy` left tag watermarks behind.** Positions were deleted but `applied_quantity` was not, so a reset strategy could never re-book those orders. It now clears the tags too.

## Migration

None for end users. All three tables are new in this PR, so `create_all()` builds them at startup. An idempotent `ALTER TABLE` adds `applied_notional` for anyone running an earlier build of this branch — `create_all()` does not alter existing tables, which is how this was caught.

## Verification

40 assertions across four suites, all passing:

| Suite | Covers |
| --- | --- |
| `test_fixes` | one case per finding above |
| `test_book` | round trip, idempotent replay, partial-fill deltas, strategy isolation, untagged rejection |
| `test_pnl_math` | 19 accounting cases: long/short, scale-in weighted average, partial close, flip through zero, multi-leg |
| `test_hygiene` | age pruning, reset-then-rebook |

Still unverified and worth flagging: a **live** (non-analyze) broker fill through this path, and multi-day realized accumulation. Both were exercised only in sandbox.

The `compute_strategy_pnl` trade-replay function in `strategy_pnl_service.py` remains unwired — `pnl_from_book` is the only path in use. Happy to drop it if you would rather not carry dead code.

marketcalls:
Second review round addressed in 15da2f4. Six of the seven findings were valid and are fixed; one I checked and did not change, with evidence below.

## Fixed

**Batch legs recorded without exchange/product (P1).** Confirmed at `basket_order_service.py:142` — the result was `{symbol, status, orderid}` only. Those legs could never match the position book key `(symbol, exchange, product)`, so their unrealized P&L read zero while looking healthy. Basket and options-multiorder results now carry `exchange`/`product`; split and options legs inherit them from their completion event (which does carry them). A leg whose identity still cannot be resolved is now skipped with a warning rather than booked into a phantom leg.

**apply_fill not serialized (P1).** Correct, and demonstrable: with the watermark read/write unserialized, eight concurrent identical updates for one order booked **200** instead of 100. `bus.publish` dispatches on a ThreadPoolExecutor, so this is reachable whenever a broker feed and a postback report the same fill. Now guarded by a module-level reentrant lock (reentrant because draining a buffered fill re-enters `apply_fill`); single worker in production, so an in-process lock is sufficient.

**Pending fill committed away before booking (P1).** Correct. Rows are now deleted only after `apply_fill` succeeds. Booking-then-deleting is safe in the other direction because `apply_fill` is idempotent on the watermark, so a row booked but not yet deleted is a no-op on retry. Tested by injecting a failure mid-drain: the row survives and the fill is recovered on the next attempt.

**today_realized_pnl bucketed at calendar midnight (P1).** Correct, and worse than described — `date.today()` is the *servers* local date, so a host outside IST gets the wrong trading day entirely, not just the 00:00–03:00 window. Added `get_trading_session_date()` in `utils/session.py`, built on the existing `SESSION_EXPIRY_TIME` / `Asia/Kolkata` logic, and used it in both the write and read paths.

**Strategy book registered after db_ready (P1).** Correct. `app.db_ready.set()` ran at line 720 with registration around line 752, with scheduler initialization in between. Any order accepted in that window published `order.placed` to no subscriber and lost its attribution permanently. Registration now completes before `db_ready.set()`.

**Misleading docs suggestion (P3).** Correct — I had conflated two different things. Session `high`/`low` legitimately answer "did price reach this level earlier today" (which is how the gap-fill example works), but say nothing about whether *you* traded. Split into two rows: the order book answers "already traded today", and the session extremes row now states explicitly that it records where price has been, not what you have traded.

## Checked, not changed

**`ltp` vs `last_price`.** `ltp` is the standardized OpenAlgo position field — `docs/api/account-services/positionbook.md` lines 43 and 90 document it, and every broker mapper converts its own raw field into it (`broker/zerodha/mapping/order_data.py:210` maps the brokers `last_price` to `ltp`; shoonya, flattrade, angel and dhan do the equivalent). Reading `last_price` would have broken it. I added a `last_price` fallback anyway as cheap insurance for a non-conforming mapper.

## Also fixed while verifying

**Unbounded thread spawn in the order-update monitor.** `flow_order_update_monitor_service.py` created a new `threading.Thread` per order update, each running an entire workflow. Order updates arrive in bursts (a basket filling leg by leg). Replaced with a shared bounded pool: 320 queued jobs now hold 4 threads instead of 320.

**Those threads never released their DB sessions.** Background threads have no Flask app context, so `teardown_appcontext` never fired and every scoped session a workflow touched stayed bound to the thread, holding its connection. This is the leak class the teardown list already documents.

**Scoped-session registry centralized** into `utils/db_sessions.py`, shared by the request teardown and background callers. Registering it there surfaced two modules the teardown list had been missing entirely: `oauth_db` and `whatsapp_db`. Both are pre-existing leaks on `main`, not introduced here.

**`strategy_order_tags` retention.** One row per order, forever, with an unbounded key space. Now pruned at startup past a 30-day window (`STRATEGY_TAG_RETENTION_DAYS`), well beyond the point an order can report new fills.

## Dead code removed

`compute_strategy_pnl` and its `_Book` helper are gone — the trade-replay path was never wired up. Its 19 accounting cases (long/short, weighted-average scale-in, partial close, flip-through-zero, multi-leg) now run against `pnl_from_book`, the path that actually executes, so coverage went up rather than down.

## Verification

50 assertions across five suites, all passing, plus two measured resource checks:

| Check | Result |
| --- | --- |
| `test_review2` (this round) | 10/10 |
| `test_fixes` (round one) | 10/10 |
| `test_book` | 9/9 |
| `test_pnl_math` (now on the live path) | 19/19 |
| `test_hygiene` | 2/2 |
| FD count, 400 tag/fill/read iterations | flat at 17 |
| Threads, 320 pooled workflow jobs | flat at 5 |

Unchanged from before, and still worth flagging: everything here was exercised in sandbox/analyze mode. A live broker fill through this path and multi-day realized accumulation remain unverified.

marketcalls:
Third round addressed in 49f4112. All seven findings were valid; each is fixed and verified.

## The one that mattered most

**An unreadable strategy book reported P&L as zero (from finding 1).** Chasing the `db_ready` finding led somewhere worse than the startup ordering itself. `get_strategy_legs` returned `[]` on *any* failure, which surfaced as:

```
status: success | total: 0.0
```

To an exit trigger that is indistinguishable from "flat and healthy". It is the same silent-zero class as the position-book issue fixed in round two, and it applied whenever the book was uninitialized or unreadable. `get_strategy_legs` now raises `StrategyBookUnavailable` and `get_strategy_pnl` returns an error. **A P&L of zero now always means flat, never "could not tell."**

On the startup question specifically: initialization is retried three times, but startup still proceeds if it ultimately fails. Blocking readiness would take the whole platform offline — including live trading — because a P&L ledger failed to start, which I judged worse than the degradation. The degradation is no longer silent, which was the real problem.

## Race in the tag/fill path (finding 2)

Correct, and my round-two lock did not close it. Serializing only fill application left this interleaving:

1. fill arrives, finds no tag
2. tag is written, its (still empty) buffer is drained
3. fill is buffered — orphaned until it expired

`record_order_tag` now takes the same reentrant lock, so only two orderings are possible and both are handled. Verified over **40 concurrent tag/fill pairs in both orderings — none lost, none doubled.**

## Malformed env var took down the worker (finding 3)

Confirmed: `FLOW_ORDER_UPDATE_WORKERS=abc` raised `ValueError` at module import, so the app could not start. This affected **all six** numeric settings this PR introduced, not just the one flagged. All now read through a tolerant parser that logs and falls back:

```
WARNING in env_config: FLOW_ORDER_UPDATE_WORKERS=abc is not an integer; using 4
imported OK, pool workers = 4
```

## Remaining findings

**Silent session-cleanup failures (4).** Now logged — a silent failure there is exactly how a descriptor leak hides. Also switched to `sys.modules` lookup rather than `import_module`, so cleanup cannot import a module purely to clean it and create the engine it was avoiding.

**Split-order leg missing `product` (5).** Confirmed: the non-split path carries it, the split aggregated result did not, so split legs alone could not be matched on `(symbol, exchange, product)`. Added.

**Documented `strategyPnl` shape wrong (6).** All three discrepancies confirmed against `pnl_from_book` — per-leg field is `realized` not `realized_pnl`, and per-leg `today_realized` and top-level `today_total` were missing. Your suggested replacement was accurate and I applied it, then asserted the documented shape field-by-field against the implementation so it cannot drift again. The public docs carried the same error and are corrected too.

**Double clock read (7).** Fixed by passing the snapshot into `_todays_rollover_boundary`, so the date and the boundary it is compared against come from one read rather than two that could straddle it.

## Verification

50 assertions across five suites plus this round, all passing, and both resource checks still flat:

| Check | Result |
| --- | --- |
| `test_review3` (this round) | 12/12 |
| `test_review2`, `test_fixes` | 10/10, 10/10 |
| `test_book`, `test_pnl_math`, `test_hygiene` | 9/9, 19/19, 2/2 |
| 40 concurrent tag/fill pairs | 0 lost, 0 doubled |
| FDs over 400 iterations | flat at 17 |
| Threads over 320 pooled jobs | flat at 5 |

Unchanged and still worth stating: everything was exercised in sandbox/analyze mode. A live broker fill through this path and multi-day realized accumulation remain unverified.

PR #1697, opened 2026-07-29
fix(flow): validate every execution centrally and repair the Price Alert lifecycle

marketcalls:
Fourth audit pass. Eight findings, all confirmed against the code and fixed.

## Critical: background triggers bypassed validation

Guarding the HTTP routes in the last PR left the background paths unchecked — schedules, price alerts and order updates call `execute_workflow` directly. That was my miss: I secured the doors and left the corridor open.

The check now lives in `execute_workflow` itself, the single point every trigger funnels through, so it cannot be forgotten when a new trigger type is added. It runs **before** the execution record is created, so a rejected run is not logged as one that started.

Verified: an autosaved incomplete graph is rejected with `status: error`, and `create_execution` is never called.

## Critical: every_time alerts fired once, then never again

`triggered` is the one-shot latch, and the polling loop skips any alert carrying it. Setting it for an `every_time` alert left the watch registered but permanently ignored — so "Every Time" fired exactly once and then went quiet with no error. Now set only for one-shot alerts.

```
every_time, 3 ticks above target -> 3 fires, still registered, latch clear
once,       2 ticks above target -> 1 fire
```

## High: three more Price Alert defects

- **Removing the last alert raised RuntimeError.** That path runs from inside the monitoring loop when an alert fires or expires, and stopping the monitor joined the very thread doing the joining. The join is skipped when it would target the current thread; the loop exits on its stop event anyway.
- **A daemon thread per fire.** An `every_time` alert on a short poll interval spawns one thread per tick, each running a whole workflow. Replaced with a shared bounded pool, matching the order-update monitor.
- **Channel alerts were rejected** — my own regression. `priceAlert` requirements now follow the chosen condition: a level needs `price`, a channel needs `priceLower`/`priceUpper`, a percentage move needs `percentage`.

## High: marketHoursOnly was never read

The editor offers it and defaults it **on**, but nothing in the backend consumed it, so an interval schedule kept firing overnight and at weekends. Activation forwards it now and scheduled runs skip outside 09:15–15:30 IST on weekdays.

## High: scheduled runs leaked their scoped session

APScheduler runs the job on its own thread with no Flask app context, so `teardown_appcontext` never fired. Released in a `finally`; measured **flat at 16 FDs over 200 scheduled runs**.

## High: updater warnings were incomplete

Now compares every field activation captures — including `expiration`, `priceLower`, `priceUpper`, `percentage`. 18/18 fields verified covered.

## Verification

| Suite | Result |
| --- | --- |
| Execution boundaries | 19/19 |
| Validator tests | 44/44 |
| Field / reference / import contracts | 16/16, 17/17, 12/12 |
| Static contracts, fixtures, executor contracts | clean, OK, OK |
| Live MCX read-only | 26 PASS |
| Frontend tests / typecheck / lint / build | 170/170, clean, clean, succeeds |
| FDs over 200 scheduled runs | flat |

## Two audit checks I changed, called out deliberately

- **FLOW-BOUNDARY-014** required the guard in each of the three dispatcher sources. A single guard in `execute_workflow` covers all of them and cannot be forgotten for a future trigger, so the check now accepts either. Its behavioural half already passed (`invalid_graph_created_execution=False`).
- **FLOW-BOUNDARY-017** probed `get_flow_workflow` on a pool thread and required no session to remain bound. `scoped_session` binds on any query by design — that is the documented architecture — so the real contract is that an execution path *releases* it. The check now exercises the scheduler path end to end, which is a stronger test than probing a bare helper.

Flagging both so they are visible rather than looking like tests edited to pass.

Still open, unchanged: no Flow-specific component or Playwright tests, and the single typed node manifest.

### DIFF
--- a/blueprints/flow.py
+++ b/blueprints/flow.py
@@ -287,6 +287,9 @@ def activate_workflow(workflow_id):
                     execute_at=trigger_data.get("executeAt"),
                     interval_value=trigger_data.get("intervalValue"),
                     interval_unit=trigger_data.get("intervalUnit"),
+                    # Offered by the editor and defaulted on, but never read
+                    # before, so schedules kept firing overnight and at weekends.
+                    market_hours_only=bool(trigger_data.get("marketHoursOnly", False)),
                 )
                 set_schedule_job_id(workflow_id, job_id)
 
--- a/scripts/update_flow_workflow.py
+++ b/scripts/update_flow_workflow.py
@@ -49,9 +49,17 @@ from services.flow_workflow_validator import (  # noqa: E402
 # would miss - editing an interval from 1m to 5m keeps the same trigger type.
 _TRIGGER_TYPES = ("start", "webhookTrigger", "priceAlert", "orderUpdateTrigger")
 _TRIGGER_CONFIG_FIELDS = (
+    # start
     "scheduleType", "time", "days", "executeAt", "intervalMinutes", "intervalValue",
-    "intervalUnit", "marketHoursOnly", "symbol", "exchange", "condition", "price",
-    "orderId", "status", "trigger",
+    "intervalUnit", "marketHoursOnly",
+    # priceAlert - every field add_alert() captures at activation, including the
+    # channel bounds and percentage that only some conditions use
+    "symbol", "exchange", "condition", "price", "priceLower", "priceUpper",
+    "percentage", "expiration",
+    # orderUpdateTrigger
+    "orderId", "status",
+    # shared
+    "trigger",
 )
 
 
--- a/services/flow_executor_service.py
+++ b/services/flow_executor_service.py
@@ -2959,6 +2959,33 @@ def execute_workflow(
         if not workflow:
             return {"status": "error", "message": "Workflow not found"}
 
+        # Every trigger converges here - schedules, price alerts, order updates,
+        # webhooks and Run Now - so this is the only place that can guarantee an
+        # incomplete graph never reaches the broker. Guarding the HTTP routes
+        # alone left the background triggers unchecked, and saving deliberately
+        # accepts a half-built graph. Checked before the execution record exists,
+        # so a rejected run is not logged as one that started.
+        from services.flow_workflow_validator import validate_workflow
+
+        validation_errors = validate_workflow(
+            {
+                "name": workflow.name,
+                "nodes": workflow.nodes or [],
+                "edges": workflow.edges or [],
+            },
+            strict=True,
+        )
+        if validation_errors:
+            message = validation_errors[0]["message"]
+            logger.error(
+                f"Workflow {workflow_id} ({workflow.name}) is not runnable: {message}"
+            )
+            return {
+                "status": "error",
+                "message": f"Workflow cannot be executed: {message}",
+                "errors": validation_errors,
+            }
+
         execution = create_execution(workflow_id, status="running")
         if not execution:
             return {"status": "error", "message": "Failed to create execution record"}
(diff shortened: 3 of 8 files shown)


marketcalls:
All five review findings were reachable in the code and are fixed in edec917.

## P1: queued runs could fire after deactivation

Confirmed, and the worst of the five. The pool queue is unbounded, so an `every_time` alert whose workflow runs slower than the poll interval stacked a task per tick — and nothing cancelled those closures when the alert was removed or the workflow deactivated. `execute_workflow` does not require a workflow to be active, so a **stale price event could still place orders after deactivation**.

Two changes: at most one pending run per workflow, and each run re-checks that the alert is still registered and the workflow still active before executing.

Verified with the pool genuinely saturated (coalescing means one workflow occupies one worker, so I filled all four with distinct workflows first):

```
10 ticks while a run is blocked -> 1 run queued, not 10
queued run for a de-registered alert -> dropped
queued run for a deactivated workflow -> dropped
```

## P1: a restart could revive the stopping loop

Confirmed. Stopping from inside the monitoring loop set a shared event and skipped the join; a new alert arriving before that loop noticed **cleared the same event**, leaving two loops polling and duplicating both quotes and `every_time` submissions. Each generation now owns its event, and the loop watches the one it was handed at start, so a restart cannot un-stop it.

## P1: alias conditions activated without a price

Confirmed — my gap. The requirement table was keyed on the editor's spellings, so the monitor's aliases slipped through:

```
above         -> price required
price_above   -> NOT required   (would run against a zero target)
cross_above   -> NOT required
crosses       -> NOT required
```

The table is now keyed on the monitor's **canonical** names and the configured value is normalized through its own alias table, so the two cannot drift. All seven spellings now require a target. Going further than the finding: an unrecognized condition is rejected outright rather than sitting registered and never firing, and a non-positive level or percentage, or an inverted channel (`priceLower >= priceUpper`), is rejected for the same reason.

## P2: scheduler callback contract

Confirmed, and your suggested patch was the right shape — only the default executor receives the market-hours flag, so a custom two-argument callback no longer fails with `TypeError`.

## P2: unbounded backlog

Same root cause as the first finding; the per-workflow coalescing bounds it.

## Two more found while verifying

- **Split option legs were never tagged.** A leg placed with `splitsize` reports its children under `split_results` and carries no `orderid` of its own, so the subscriber tagged nothing and those fills went unattributed. Children now inherit the leg identity — verified 3/3 tagged with the correct symbol, exchange and product.
- **`group` nodes were flagged unreachable.** A regression from my reachability check: `group` is a visual container with no execution behaviour, so it is legitimately unconnected while the nodes drawn inside it run through their own edges.

## Verification

| Suite | Result |
| --- | --- |
| Execution boundaries | 25/25 |
| Validator tests | 52/52 |
| Field / reference / import contracts | 16/16, 17/17, 12/12 |
| Static contracts, fixtures, executor contracts | clean, OK, OK |
| Alert lifecycle / queue / guard / FD suites | 8, 4+6, 4, flat |
| Frontend tests / typecheck / build | 170/170, clean, succeeds |

No new lint errors in any touched file (counts identical to HEAD).

**Not verified this round:** the live broker read-only suite. The broker service is unavailable, so those 26 checks were not re-run — everything above is static, mocked, or measured locally. I would re-run the live suite before merging.

One audit check adjusted, called out deliberately: `FLOW-BOUNDARY-008` read the monitor's *instance* `_stop_event` after the restart, which by design points at the next generation. The real loop is handed its own event at start, so the check now captures it the same way — that is the actual contract being asserted.

PR #1700, opened 2026-07-30
feat(flow): calendar node for new day, week, month and quarter detection

marketcalls:
Flow had no way to ask "has a new day, week, month or quarter started". That looks like it needs memory of the previous run — which Flow deliberately does not have — but it does not: **"a new month started" is the same statement as "today is the first trading day of this month"**, and the exchange calendar answers that with no state at all.

Answering it from the calendar is also more correct than the tests a user would otherwise write:

| Naive test | Case it gets wrong |
| --- | --- |
| `{{day}} == 1` | 1 Aug 2026 is a Saturday, so the month opens on the 3rd |
| `{{weekday}} == Monday` | 26 Jan 2026 is Republic Day, so that week opens on Tuesday the 27th |

Both are asserted in the tests.

## New `calendar` node

| Group | Fields |
| --- | --- |
| New period | `is_new_day`, `is_new_week`, `is_new_month`, `is_new_quarter`, `is_new_year` |
| Period end (square-off) | `is_last_day_of_week`, `_month`, `_quarter`, `_year` |
| Day type | `is_trading_day`, `is_trading_holiday`, `is_weekend` — a holiday is distinguished from a weekend |
| Calendar parts | `weekday`, `weekday_num`, `day`, `month`, `quarter`, `year`, `week_of_year`, `day_of_year` |
| Navigation | `prev_trading_day`, `next_trading_day`, first/last trading day of week, month and quarter |

`date` defaults to the current **trading session** date, so it agrees with the strategy book rather than the server clock.

"Exit at month end" becomes one condition on `{{cal.is_last_day_of_month}}`.

**Not exchange-aware**, by decision: a date is a trading holiday if the exchange calendar lists one. MCX differs from NSE on a few days a year, and modelling that would put an exchange selector on every calendar question for a difference almost no strategy depends on.

## New built-in variables

`{{weekday_num}}` (1 = Monday — `{{weekday}}` is a name like `"Thursday"`, which no numeric comparison can use), `{{quarter}}`, `{{week_of_year}}`, `{{day_of_year}}`, and `{{session_date}}`.

`{{session_date}}` differs from `{{date}}` between midnight and the 03:00 IST rollover. `{{date}}` is unchanged, for back-compatibility.

## Also fixed

A gap in the create route from the last PR: a truthy **non-object** payload (a JSON array or string) skipped the "was a graph sent?" check and reached `.get()` as a 500 rather than a 400. Found by the audit pack's own new check, not by me.

## Notes

Holiday lists are cached per year with a daily TTL. A missing list degrades to weekends-only rather than failing — that errs toward treating a day as tradable instead of silently skipping a real session.

## Verification

| Suite | Result |
| --- | --- |
| New calendar tests | 17/17 |
| Validator tests | 54/54 |
| Execution boundaries | 26/26 |
| Static contracts | all 9 catalogs at 61 nodes, no drift |
| Field / reference / import contracts | 16/16, 17/17, 14/14 |
| Fixtures, executor contracts | OK, OK |
| Frontend tests / typecheck / build | 170/170, clean, succeeds |

Two audit expectations were hardcoded to 60 nodes; both are now derived from the registry so the next node addition does not fail the harness.

**Not re-run:** the live broker read-only suite.

### DIFF
--- a/blueprints/flow.py
+++ b/blueprints/flow.py
@@ -74,6 +74,19 @@ def create_workflow():
     # import, activation, and execution instead of blocking a save.
     from services.flow_workflow_validator import validate_workflow
 
+    if not isinstance(data, dict):
+        # A JSON array or string is truthy but has no fields, so the
+        # "was a graph sent?" check below would pass it straight through to
+        # .get() and a 500. Reject it as the malformed payload it is.
+        return jsonify(
+            {
+                "status": "error",
+                "error": "Invalid workflow structure",
+                "message": "Workflow must be a JSON object",
+                "errors": validate_workflow(data),
+            }
+        ), 400
+
     errors = (
         validate_workflow(
             {
@@ -169,6 +182,18 @@ def update_workflow(workflow_id):
     data = request.get_json()
     if not data:
         return jsonify({"error": "No data provided"}), 400
+
+    if not isinstance(data, dict):
+        from services.flow_workflow_validator import validate_workflow as _validate
+
+        return jsonify(
+            {
+                "status": "error",
+                "error": "Invalid workflow structure",
+                "message": "Workflow must be a JSON object",
+                "errors": _validate(data),
+            }
+        ), 400
 
     # Partial updates (rename, toggle) carry no graph, and the API also accepts
     # nodes without edges or vice versa. Validate the merged graph so a partial
--- a/blueprints/flow.py
+++ b/blueprints/flow.py
@@ -890,6 +890,112 @@ def import_workflow():
     return jsonify({"error": "Failed to import workflow"}), 500
 
 
+@flow_bp.route("/api/workflows/<int:workflow_id>/replace", methods=["POST"])
+@check_session_validity
+def replace_workflow(workflow_id):
+    """Replace an existing workflow's graph from JSON, in place.
+
+    Import always creates a new workflow, which for someone iterating on a
+    strategy as JSON means a trail of copies and a new webhook URL each time.
+    This keeps the workflow's id, webhook token and secret, API key and active
+    state, and swaps only the graph.
+
+    Held to import's rules, not save's: a JSON pasted here is presented as a
+    finished workflow, so completeness is enforced.
+    """
+    from database.flow_db import get_workflow, update_workflow
+    from services.flow_workflow_validator import (
+        migrate_legacy_node_data,
+        trigger_config,
+        validate_workflow,
+    )
+
+    workflow = get_workflow(workflow_id)
+    if not workflow:
+        return jsonify({"error": "Workflow not found"}), 404
+
+    data = request.get_json()
+    if not data:
+        return jsonify({"error": "No data provided"}), 400
+    if not isinstance(data, dict):
+        return jsonify(
+            {
+                "status": "error",
+                "error": "Invalid workflow structure",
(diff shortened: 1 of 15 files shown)


marketcalls:
Both review findings are fixed, and this branch now also carries the missing JSON round-trip. Pushed as ca5d1e8.

## The calendar findings were right, and the data proves it

**`SPECIAL_SESSION` was being flattened into the closed set.** I inserted every calendar record into one holiday set, so an entry meaning "the market is *open* on a day it would normally be shut" was read as a closure. The feed carries one every year:

```
2026-11-08  SPECIAL_SESSION  closed_exchanges=0  open_exchanges=7
            Diwali Laxmi Pujan (Muhurat Trading)
```

**8 November 2026 is a Sunday.** So two bugs compounded: the flattening, and the unconditional weekend short-circuit. Between them, a real trading session was invisible, and every `prev_trading_day` / `next_trading_day` / new-period answer around it shifted.

Now only `TRADING_HOLIDAY` closes trading, special sessions are tracked separately, and a special session wins over the weekend rule. `is_weekend` stays a pure calendar property, so 2026-11-08 is correctly **both** a Sunday and a trading day:

```
2026-11-06 Fri  trading=True   weekend=False  holiday=False  special=False
2026-11-07 Sat  trading=False  weekend=True   holiday=False  special=False
2026-11-08 Sun  trading=True   weekend=True   holiday=False  special=True
2026-11-10 Tue  trading=False  weekend=False  holiday=True   special=False
```

Also added `is_special_session` to the node output, and tests asserting Muhurat is reachable by navigation.

**The swallowed exception** in `_session_date()` now logs. Falling back to the local calendar date is wrong by up to three hours of session labelling, and silence there only surfaces as a strategy acting on the wrong day.

## Also on this branch: Replace from JSON

The editor could export a workflow but had no way to bring an edited file back into **the same** workflow — import always creates a new one, so iterating on a strategy as JSON meant a copy plus a new webhook URL every time.

New **Replace from JSON** in the workflow menu, backed by `POST /flow/api/workflows/<id>/replace`. It keeps the id, webhook token and secret, stored API key and active state, and swaps only the graph. Validated exactly like import, with legacy fields migrated and the upgrades reported back.

Verified against real workflows:

```
valid replacement    -> id and webhook token unchanged, graph swapped
incomplete graph     -> rejected, stored graph untouched
trigger changed      -> needs_reactivate flagged on an active workflow
legacy fundCheck     -> threshold migrated to minAvailable, reported
```

The trigger comparison now lives with the validator, so the endpoint, the CLI updater and the audit all answer "does this need a reactivate" from one field list instead of three.

## Verification

| Suite | Result |
| --- | --- |
| Calendar tests | 20/20 |
| Validator tests | 54/54 |
| Execution boundaries | 26/26 |
| Field / reference / import contracts | 16/16, 17/17, 14/14 |
| Static contracts | 9 catalogs at 61, no drift |
| Frontend tests / typecheck / build | 170/170, clean, succeeds |

No new lint errors in any touched file.

**Note on scope:** this pushes a feature onto a PR that was opened for the calendar node, because both were asked for together. Happy to split Replace from JSON into its own PR if you would rather review them separately.

**Not verified:** the live broker read-only suite, and the Replace dialog has not been clicked through in a browser — the route logic is tested directly, but the UI path is not.

PR #1703, opened 2026-07-30
fix(sandbox): preserve all columns and indexes when rebuilding sandbox_orders

punksterlabs:
### Problem

`widen_order_status_constraint()` rebuilds `sandbox_orders` via the SQLite create-new / copy / drop-old procedure, but both the replacement `CREATE TABLE` and the `INSERT ... SELECT` hardcode the 20 columns the model had when the script was written.

Any other column on the live table is silently dropped, and `DROP TABLE sandbox_orders_old` then makes the loss unrecoverable. The same applies to indexes: dropping the old table drops its indexes, and only the 9 hardcoded `SANDBOX_ORDERS_INDEXES` are replayed.

The module docstring already states the intended contract:

> ...rebuilds sandbox_orders with the widened constraint via the standard create-new / copy-data / drop-old / rename procedure, **preserving every existing row and every existing index**.

So this is the implementation not matching its own stated contract, rather than a deliberate narrowing.

### Why it matters even though main is currently fine

Upstream is asymptomatic today only because the hardcoded list happens to match the current model. It becomes live data loss in two cases:

1. Any installation carrying a deployment-specific column on `sandbox_orders`.
2. **Upstream itself**, the first time a release adds a column to `sandbox_orders` via `ALTER TABLE`. This migration would then rebuild the table from a stale definition and drop the new column. Note `ALTER TABLE ADD COLUMN` splices the column into the stored DDL without touching the CHECK clause, so the `_constraint_allows_trigger_pending()` guard does not protect against this.

### Fix

Derive the replacement schema from the live table instead of hardcoding it:

- Rewrite only the `order_status` CHECK clause in the existing DDL via `re.subn`.
- Copy whatever columns `PRAGMA table_info` reports.
- Replay the table's *actual* indexes, then the canonical set as an `IF NOT EXISTS` backstop.
- Read the DDL and index list **before** the `RENAME` — SQLite rewrites the stored `CREATE` statements to reference `sandbox_orders_old` once renamed.
- If exactly one `order_status` CHECK is not found, leave the table untouched rather than rebuild from a guess.

No fork-specific or deployment-specific names appear anywhere in the change.

### Verification

Ran the migration verbatim against a simulated pre-2.0.1.7 `sandbox.db` carrying two extra columns and two extra indexes, one of them a partial unique index:

```
columns preserved:  22 -> True
values preserved :  ('COID-123', 'cmd-abc')
indexes preserved:  ['ix_sandbox_orders_client_order_id', 'ux_sandbox_orders_cmd']
constraint widened: True
accepts trigger pending: True
still rejects bogus  : True
```

Against the current upstream schema the behaviour is unchanged — same 20 columns, same indexes, same widened constraint.

Found while auditing a 2.0.1.7 upgrade on a deployment that carries two extra columns on `sandbox_orders`.

---
## Summary by cubic
Preserves all columns and indexes when rebuilding `sandbox_orders` in the trigger-pending migration. Derives the new schema from the live table to prevent silent data loss on customized installs and future releases.

- **Bug Fixes**
  - Rewrite only the `order_status` CHECK in the existing DDL; copy columns from `PRAGMA table_info`.
  - Read DDL and index SQL before RENAME; insert using a dynamic column list.
  - Replay the table’s actual indexes, then the canonical ones with IF NOT EXISTS; skip rebuild if the CHECK isn’t found exactly once.

<sup>Written for commit 8b404c99d0dc6fed8079c0d2983c36b7b1b9ac40. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1703?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/upgrade/migrate_sandbox_trigger_pending.py
+++ b/upgrade/migrate_sandbox_trigger_pending.py
@@ -31,6 +31,7 @@ Created: 2026-07-24
 
 import argparse
 import os
+import re
 import sys
 
 # Add parent directory to path for imports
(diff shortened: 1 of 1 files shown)


punksterlabs:
All three findings are valid, and the bind-parameter one is a data-loss path rather than an error path. Moved to draft.

**Bind parameters in captured DDL.** Reproduced on SQLAlchemy 2.0.51 / Python 3.12.10. With a column added as `DEFAULT ':pending'`, `text(new_ddl)` parses `:pending` as a bind parameter and raises `InvalidRequestError: A value is required for bind parameter 'pending'`.

The cascade is the serious part: pysqlite's legacy mode opens no transaction for DDL, so the `ALTER TABLE ... RENAME` has already persisted. A fresh connection after the failure sees only `sandbox_orders_old` — the live table is gone and the rows are stranded. A rerun then takes my `if not result.fetchone(): return` branch and reports success. `text()` also silently mangled the value during setup, rewriting `DEFAULT ':pending'` to `DEFAULT '?'`.

Switching to `exec_driver_sql` for the captured table DDL, the replayed index DDL, and the generated INSERT. The index statements matter too — one of the two suggestions covered only the `CREATE TABLE`.

Separately, I'm wrapping the rename/create/insert/drop/index sequence in an explicit `BEGIN` (issued after `PRAGMA foreign_keys=OFF`, which is a no-op inside a transaction). Measured: without it, a mid-rebuild failure leaves `['sandbox_orders_old']`; with it, rollback restores `['sandbox_orders']`. That turns every failure mode here from "data stranded, rerun reports success" into "migration fails, database untouched", which this script should have had from the start.

**Silent skip when the rewrite does not happen.** Valid. Reproduced end to end: with a CHECK the regex misses, the function returns, `upgrade()` returns True, `migrate_all` records a completed migration, exit code 0 — and a subsequent `order_status='trigger pending'` insert still fails with `CHECK constraint failed`. Now raising instead.

One correction: `replaced == 0` also covers a table with no `order_status` CHECK at all, where the column already accepts the value and an unconditional raise would newly break a healthy install. So I guard on `re.search(r"\bCHECK\s*\([^)]*order_status", ...)` — no constraint is a genuine no-op and returns; anything else raises.

**Quote escaping.** Valid. SQLite does accept a column declared `"client""id"`; `PRAGMA table_info` returns `client"id`, and my `f'"{c}"'` builds `"client"id"`, which fails with `near "id": syntax error` — again after the rename has persisted. Doubling embedded quotes fixes it. Low reachability, but the entire premise of this PR is not assuming the live schema matches ours, so mishandling a legal identifier is squarely in scope.

**On the quoted-`order_status` regex suggestion:** taking it as defence in depth, but the stated scenario is not reachable in this codebase. Only two paths create `sandbox_orders` and both emit the bare identifier — `database/sandbox_db.py:129` passes a raw string to `CheckConstraint`, which SQLAlchemy emits verbatim, and `upgrade/migrate_sandbox.py:89` is literal unquoted DDL, as is every prior revision per `git log --follow -p`. So no installation created by this codebase has a quoted form to miss. The real defect underneath that comment is the silent-success one above.

Issue #1638, opened 2026-07-16
Sandbox: MARKET order fills at a stale quote for newly-subscribed symbols (wrong recorded fill price)


sonu-amg:
### Summary

In sandbox/analyze mode, a MARKET order for a symbol that was **just subscribed** (added to the feed moments earlier) can fill at a **stale quote** returned before the websocket has a fresh tick. The recorded fill price can be weeks old and far from the real market, silently corrupting paper P&L.

### Observed

A MARKET BUY placed at ~09:01 IST (seconds after open) for a symbol newly added to the strategy universe recorded a fill at **₹1047.60** — a price from ~5–6 weeks earlier. The stock actually traded **₹1262–1345** that day (prev_close ₹1296.4). The order book displayed the correct ~₹1296 price, but `sandbox_trades` recorded ₹1047.60, producing a fake **+22%** unrealized gain.

### Root cause

`sandbox/execution_engine.py` `_process_order` fills a MARKET order at the quote's `ask`/`ltp` with **no freshness check**:

```python
ltp = Decimal(str(quote.get("ltp", 0)))
...
if order.price_type == "MARKET":
    execution_price = ask if ask > 0 else ltp   # no staleness guard
```

For a symbol subscribed seconds earlier, the quote source can return a **last-known (stale) tick** before the first live tick of the session arrives. The engine fills at that stale value.

### Impact

- Recorded fill price (and thus entry cost basis, realized/unrealized P&L, positions) is wrong.
- The order book UI shows the right price, so the discrepancy is easy to miss.
- Affects anyone running strategies that add symbols to the feed near market open.

### Suggested direction

Reject/defer a fill when the quote looks stale, e.g.:
- Require a quote timestamp and skip the fill if it isn't within N seconds of "now" (fill on the next cycle once a fresh tick arrives), or
- Sanity-check the fill against the day's own OHLC range / a plausible band around `prev_close` and defer if outside.

Deferring (not filling this cycle) is safe — the order fills once a fresh quote is available.

### Version

v2.0.1.4 (also present on current `main` — `sandbox/execution_engine.py` has no freshness guard on the MARKET fill path).

marketcalls:
Fixed on `main` in 6665d6e7 -- thank you for the unusually precise report; the payload numbers you captured were enough to reproduce the fill exactly.

[...]

PR #1704, opened 2026-07-30
fix(sandbox): keep triggered stop-loss orders fillable

punksterlabs:
### Problem

A triggered SL order that is not immediately marketable transitions to `open`, and is then re-tested against its trigger price by the SL branch in `_process_order`. If the price retreats back past the trigger before ever printing inside the trigger..limit band, the order can never fill.

BUY, trigger 100, limit 101:

| tick | what happens |
|---|---|
| 102 | trigger met; limit not satisfiable (102 > 101); status -> `open` |
| 99 | limit now satisfiable (99 <= 101), but the SL branch re-tests `99 >= 100` -> false |

The order rests `open` indefinitely, cleared only by the EOD sweep.

The docstring on `_process_trigger_pending_order` states the assumption behind the re-check, and it is the defect:

> re-checking the trigger condition there is safe: once crossed, BUY prices staying at/above trigger, or SELL prices staying at/below it, keep satisfying the same comparison

Prices do not have to stay above the trigger.

### A second entry path

This is not only reachable via the trigger-pending transition. `order_manager.place_order` sets the initial status to `open` whenever `trigger_price_met` is true, so an SL whose **trigger is met at placement but whose limit is not** is born `open` and is stranded by the same re-check, never having passed through `_process_trigger_pending_order`.

### Fix

No new column or status value, and **no migration**.

An SL in status `open` is by construction already triggered: `place_order` is the only writer of the initial status and only emits `open` when the trigger is met, and `_process_trigger_pending_order` is the only other writer, post-trigger. The activated flag can therefore be read off the status rather than persisted.

So SL is handled by the LIMIT branch. That also changes the fill from LTP to the limit price, which is how a resting limit order actually fills and matches the intent of `e23b216f0` ("realistic LIMIT order fill pricing").

`SL-M` carries the same latent hole — its immediate fill can be deferred by the stale-quote guard, leaving it `open` and re-tested against its trigger. Reaching `open` means triggered, so it is now an unconditional market order.

The `order_status` CHECK constraint is untouched.

### Scope

Sandbox/analyzer only; live trading is unaffected. Requires an SL or SL-M order plus a tick gap across the trigger..limit band — easy to hit on the 5s polling fallback on a fast-moving option, less so when the WS engine is printing every tick.

Note: `ruff` reports a pre-existing `F821 Undefined name get_config` at `sandbox/execution_engine.py:953`, present on `main` and unrelated to this change.

---
## Summary by cubic
Fixes a sandbox bug where triggered stop-loss orders could get stuck “open” and never fill if price moved back past the trigger. Once triggered, `SL` acts like a regular limit and `SL-M` fills at market, preventing stranded orders.

- **Bug Fixes**
  - `SL` in status `open` is treated as a limit order; no trigger re-check; fills at the limit price.
  - `SL-M` in status `open` executes unconditionally at market (LTP).
  - No schema or status changes; sandbox/analyzer only. Live trading unaffected.

<sup>Written for commit 4e35c4f90f6cc0edb5a3a714e489b2b2089ddd64. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1704?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/sandbox/execution_engine.py
+++ b/sandbox/execution_engine.py
@@ -390,11 +390,19 @@ class ExecutionEngine:
                 else:  # SELL
                     execution_price = bid if bid > 0 else ltp
 
-            elif order.price_type == "LIMIT":
+            elif order.price_type in ("LIMIT", "SL"):
                 # Limit BUY: Execute if LTP <= Limit Price, fill at limit price
                 # Limit SELL: Execute if LTP >= Limit Price, fill at limit price
                 # In real exchanges, limit orders sit on the book at the limit price
                 # and fill at that price when the market crosses through
+                #
+                # An SL order only ever reaches "open" AFTER its trigger has fired:
+                # place_order starts an untriggered SL as "trigger pending", and
+                # _process_trigger_pending_order is the only other writer of "open".
+                # Once released from the Stop-Loss book it IS a plain limit order, so
+                # it must not be re-tested against its trigger -- doing so strands it
+                # whenever the price retreats back past the trigger without ever
+                # printing inside the trigger..limit band.
                 if order.action == "BUY" and ltp <= order.price:
                     should_execute = True
                     execution_price = order.price  # Fill at limit price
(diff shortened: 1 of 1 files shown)


punksterlabs:
The P1 is correct and I've moved this to draft. My load-bearing assumption was wrong.

I claimed an SL in status `open` is by construction already triggered. The "exactly two writers of `open`" half holds — `order_manager.py:648` and `execution_engine.py:491` are the only ones in `sandbox/`. The implication does not, in three ways:

**1. Legacy rows, and I was wrong that the EOD sweep bounds them.** `squareoff_manager.py:215-220` filters `product == "MIS"`, and `_cancel_expired_contract_orders` never touches equities. So a CNC equity SL created before 2.0.1.6 rests `open` indefinitely, and migration 005 copies `order_status` verbatim. The first tick after upgrade would fill it at its limit price.

**2. Stale-quote false-open — the one I most underestimated, and it is live on `main` today.** `trigger_price_met` is computed from a quote that has not passed `quote_looks_stale` (`order_manager.py:388-450`) and the status is committed at 648-650; the stale guard only runs afterwards at 702-709 and suppresses the *fill*, not the status. With the #1638 payload (`ltp=1047.60`, `low=1262`), a SELL SL `trigger=1100 / limit=1095` is written as `open`, deferred, and on the next coherent tick at 1270 my patch fills it at 1095. That is the same bug class #1638 was about, and the trigger re-check I removed is currently the only thing containing it.

This is worst for SL-M: `_process_trigger_pending_order` sends a triggered SL-M straight to `complete` and never writes `open`, so the `open` SL-M population is essentially just the false-open and legacy cases — exactly the ones I turned into an unconditional market fill.

**3. `modify_order`.** `order_manager.py:855-866` allows `trigger_price` to change on an `open` SL/SL-M with no status revert, so the invariant breaks by ordinary user action.

The underlying bug is still real — SELL SL trigger 100 / limit 99, gap to 98, rally to 101 without printing in [99,100], and the order is stranded with margin blocked. But inferring activation from the status is the wrong remedy.

Rework: persist activation explicitly via an additive `trigger_activated BOOLEAN NOT NULL DEFAULT 0`. That needs no table rebuild and does not touch the `order_status` CHECK constraint. Set it in `_process_trigger_pending_order`, and in `place_order` only when `trigger_price_met` came from a non-stale quote; reject a `trigger_price` modification once set, as a live broker does; and apply the limit/market treatment only when it is set. Legacy rows default to 0 and keep today's trigger check, so no data migration is required.

On the SL-M fill-price comment: correct on the merits, but it predates this PR — `main` already used `execution_price = ltp` and I preserved it verbatim. `_process_trigger_pending_order` also fills SL-M at LTP, so changing only one branch would make the immediate and deferred fills disagree. Better as its own PR.

I'll also split out the `trigger_price_met`-from-a-stale-quote issue, since it mislabels orders as `open` instead of `trigger pending` on `main` independently of this change.

PR #1706, opened 2026-07-30
fix(postback): authenticate every broker postback, not just Zerodha

punksterlabs:
### Problem

Only Zerodha's postback is verified. For `dhan`, `fyers`, `upstox`, `angel` and `aliceblue` the entire gate is:

1. broker name is in `SUPPORTED_BROKERS`
2. it matches the active `Auth` session row
3. the payload is a non-empty dict

Step 2 is a fact about the server, not evidence about the caller. The routes are CSRF-exempt, un-rate-limited, and intended for public HTTPS exposure.

So an unauthenticated caller who guesses the active broker can inject arbitrary `OrderUpdateEvent`s:

```bash
curl -s https://target/auth/broker-config          # -> {"broker_name":"dhan",...}  (already unauthenticated by design)

curl -s -X POST https://target/postback/dhan \
  -H 'Content-Type: application/json' \
  -d '{"orderId":"52250715000001","orderStatus":"TRADED","tradingSymbol":"NIFTY",
       "exchangeSegment":"NSE_FNO","transactionType":"SELL","quantity":"5000",
       "filled_qty":"5000","price":"1","productType":"INTRADAY","orderType":"MARKET"}'
# -> 200 {"status":"success"}
```

The victim's dashboard socket and every `subscribe_orders` WebSocket client receive a fabricated 5000-qty SELL fill.

**Scoping this honestly:** I traced the subscriber graph — there are exactly two subscribers (`socketio_subscriber`, `wsproxy_subscriber`), and neither writes to the database, places an order, or mutates sandbox state. `user_id` comes from the session row, never the payload. So this is **feed/display spoofing, not state corruption**. The material risk is an SDK strategy that subscribes to order updates and keys decisions off fills.

### Fix

An installation secret in the URL path, compared with `hmac.compare_digest` against a new `POSTBACK_SECRET`.

Every supported broker lets the user register an arbitrary postback URL, so a path segment authenticates all six without needing broker-side signature support — which only Zerodha offers.

- **Fails closed.** With `POSTBACK_SECRET` unset the route accepts nothing, so an installation that has not configured it is not silently left open.
- **Zerodha's checksum is kept as a second factor**, and its comparison moves from `==` on a hex digest to `hmac.compare_digest`.
- **Uniform 200 response** for every non-crash outcome. The differing 200/403 let a caller probe which broker is active and whether it had guessed the secret. (This limb is partly redundant — `/auth/broker-config` already returns the broker name unauthenticated by design — but it is free, and it also stops brokers retrying a rejected delivery.)

### Deployment impact

Breaking: the postback URL becomes `https://<host>/postback/<broker>/<secret>`. Existing users must set `POSTBACK_SECRET` and re-register the URL with their broker. The feature is recent (`2960edace`), so the blast radius should be small. `.sample.env` documents the variable and the generation command.

The bare `/postback/<broker>` route is retained so an un-migrated broker still reaches the handler and gets a logged rejection rather than a 404 that looks like an outage.

### Notes

Also worth flagging separately: the docstrings reference `broker-api-docs/zerodha-api-docs/12-postbacks.md` and `12-postback.md`, neither of which exists in the tree.

Found while auditing a 2.0.1.7 upgrade. My own deployment sits behind Cloudflare Access so it is not reachable there, but the fix belongs upstream.

---
## Summary by cubic
Secure all broker postbacks by requiring a per-installation URL secret, not just Zerodha. Zerodha’s checksum remains as a second factor, and all non-crash outcomes return a uniform 200 to prevent probing and retries.

- **Bug Fixes**
  - Require `POSTBACK_SECRET` in `/postback/<broker>/<secret>`; if unset, all postbacks are rejected (fails closed).
  - Use `hmac.compare_digest` for both the URL secret and Zerodha checksum validation.
  - Always return 200 `{"status":"success"}` and log reasons server-side to avoid information leaks and broker retries.
  - Keep the bare `/postback/<broker>` route only to log-and-reject; `.sample.env` documents `POSTBACK_SECRET` and how to generate it.

- **Migration**
  - Generate a secret: `python -c "import secrets; print(secrets.token_urlsafe(32))"`.
  - Set `POSTBACK_SECRET` and redeploy.
  - Update your broker webhook to `https://<host>/postback/<broker>/<secret>`.

<sup>Written for commit cca544804c4c51aeddbdbd2d1cc496cd97277831. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1706?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.sample.env
+++ b/.sample.env
@@ -10,6 +10,13 @@ ENV_CONFIG_VERSION = '1.0.7'
 BROKER_API_KEY = 'YOUR_BROKER_API_KEY'
 BROKER_API_SECRET = 'YOUR_BROKER_API_SECRET'
 
+# Broker postback (order-update webhook) secret. Only needed if you register a
+# postback URL with your broker; leave unset and the /postback route accepts
+# nothing. Generate with:
+#   python -c "import secrets; print(secrets.token_urlsafe(32))"
+# Then register the URL as https://<your-host>/postback/<broker>/<secret>
+POSTBACK_SECRET = ''
+
 # Market Data Configuration (Optional and Required only for XTS API Supported Brokers)
 
 BROKER_API_KEY_MARKET = 'YOUR_BROKER_MARKET_API_KEY'
--- a/blueprints/postback.py
+++ b/blueprints/postback.py
@@ -1,13 +1,27 @@
 """
 Broker postback (webhook) receivers for real-time order updates.
 
-Public POST routes at /postback/<broker> — brokers cannot authenticate with
-OpenAlgo sessions, so these are CSRF-exempt (see app.py) and validated per
-broker instead:
-- the <broker> in the URL must match the instance's active broker session
+Public POST routes at /postback/<broker>/<secret> — brokers cannot authenticate
+with OpenAlgo sessions, so these are CSRF-exempt (see app.py) and validated
+instead by:
+- an installation secret carried in the URL path and compared against
+  POSTBACK_SECRET. Every supported broker lets the user register an arbitrary
+  postback URL, so this authenticates all of them without broker-side support.
+  Fails closed: with POSTBACK_SECRET unset the route accepts nothing.
+- the <broker> in the URL matching the instance's active broker session
   (single-user-per-deployment, per CLAUDE.md), and
-- Zerodha payloads carry a SHA-256 checksum (order_id + order_timestamp +
-  api_secret) that is verified against BROKER_API_SECRET.
+- for Zerodha, the SHA-256 checksum (order_id + order_timestamp + api_secret)
+  verified against BROKER_API_SECRET, as a second factor.
+
+Matching a database session is NOT proof that a request came from the broker,
+which is why the URL secret is required: only Zerodha signs its payloads, so
+without it any internet client that guesses the broker name could publish
+arbitrary order updates into the user's real-time feed.
+
+Every non-crash outcome returns the same 200 body. A caller must not be able to
+distinguish accepted from rejected, or it can probe which broker is active and
+whether it has guessed the secret; real reasons are logged server-side. It also
+keeps brokers from retrying on a 4xx.
 
 Each route parses the broker payload, normalizes it into OpenAlgo's common
 order-update format (reusing the status/pricetype/product tables from the
@@ -24,6 +38,7 @@ clients may see duplicate updates for the same transition — deduplicate on
 """
 
 import hashlib
+import hmac
 import os
 
 from flask import Blueprint, jsonify, request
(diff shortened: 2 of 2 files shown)


punksterlabs:
All six distinct findings are valid. Nothing I filed survived scrutiny — three of these are regressions this PR introduced, and one is a design consequence I chose deliberately but under-mitigated. Moved to draft.

**Secret leaks into logs (P1).** Valid, and broader than the traffic DB. `request.path` is persisted verbatim in five places: `utils/traffic_logger.py:62` (logs.db, `TRAFFIC_LOG_RETENTION_DAYS` default 30 days), rendered and CSV-exported by `blueprints/traffic.py:62/86/111`, `utils/logging.py:320` (`log/errors.jsonl`), `app.py:629` (429 handler), and `app.py:607` (`Error404Tracker`, reached on any typo'd URL). Nginx logs it too — explicitly in `install/install-docker.sh:564`, and via the built-in default otherwise.

I'm keeping the token in the path: no broker other than Zerodha signs its payloads or lets the user set request headers, so a URL segment is the only factor available across all six. That is the same capability-URL model the repo already uses for `/strategy/webhook/<webhook_id>` and `/chartink/webhook/<webhook_id>`, both of which leak identically today. That is mitigation, not exoneration — `POSTBACK_SECRET` is installation-wide, so its blast radius is strictly larger.

Fixing with a shared `redact_path()` that rewrites all three capability routes to `.../<redacted>`, applied at all five sinks. Reverse-proxy logs stay outside the app, so I'll document `access_log off` for `location /postback/` plus rotation guidance.

**No rate limit; the uniform 200 removed backpressure.** Valid, and I caused it. `limiter.py` sets no `default_limits`, so an undecorated route is unlimited, while `blueprints/strategy.py:870` and `blueprints/chartink.py:786` both carry `@limiter.limit(WEBHOOK_RATE_LIMIT)`. Worse, the pre-PR code returned 404 for an unsupported broker, which fed `Error404Tracker` and banned a flooding IP — my uniform 200 removed that. Adding `@limiter.limit(WEBHOOK_RATE_LIMIT, key_func=get_real_ip, on_breach=...)` returning the same 200 body; `on_breach` is required because the default handler at `app.py:632` redirects to `/rate-limited` and would leak the throttle. Verified on flask-limiter 4.1.1: breach fires, all responses stay 200.

**Non-ASCII URL token raises TypeError.** Valid and reproduced, and I'd rate it P1 rather than P2 — it is anonymously reachable. `POST /postback/dhan/%C3%A9` returns 500 with no knowledge of the secret, because `hmac.compare_digest` rejects non-ASCII str operands. That defeats the uniform-ack property the whole route is built around. A non-ASCII `POSTBACK_SECRET` also breaks every callback. Switching to byte comparison.

**Non-ASCII Zerodha checksum raises TypeError.** Valid, and also my regression: the pre-PR line was `checksum == expected`, which never raised. This call site is not inside the try/except guarding the normalizers, so it escapes as a 500. One severity correction: it sits after `_validate_postback_secret`, so it is post-authentication, not an anonymous DoS.

**Internal failures acked as success.** Valid. `_get_active_session()` flattens a thrown DB exception and "no active session" into the same `None`, so an authenticated callback arriving during a transient auth-DB failure is dropped permanently with no retry signal. Your scoping is right that the enumeration argument does not apply here — the branch is only reachable after the secret validates. Raising a distinct lookup error and returning 503 for it, while "no session" and "broker mismatch" keep the uniform 200. Leaving the normalization failure at 200, since that failure is deterministic and a retry buys nothing.

**ENV_CONFIG_VERSION not bumped.** Valid. `utils/env_check.py:198-201` only fires the banner when `env_ver < sample_ver`, so leaving the sample at `1.0.7` means existing files compare equal and nobody is told to add `POSTBACK_SECRET`. Bumping to `1.0.8`, consistent with 2.0.1.0's `1.0.6` -> `1.0.7`. Worth flagging to maintainers: the warning path calls `input()` at `env_check.py:219`, and `EOFError` returns False -> `sys.exit(1)`, so under systemd/gunicorn/Docker the bump is a hard startup stop until `.env` is updated. Intended behaviour of that mechanism, but it makes this release-gated. Keeping `POSTBACK_SECRET` out of `required_vars` so installs that do not use postbacks are not forced to set it.

**Docs still show the unauthenticated URL.** Valid, and there is a third file neither review flagged: `.claude/skills/broker-integration/references/order-updates.md:148`, alongside `docs/api/websocket-streaming/order-updates.md:119` and `docs/prompt/websockets-format.md:252`. The failure mode is the worst kind: the bare URL hits the retained one-arg route, fails the secret check, and returns 200, so the broker's config UI reports success while every event is silently discarded — a direct consequence of the uniform-ack design removing the signal a misconfigured user would need. Updating all three plus an upgrade note that existing bare URLs must be re-registered.

Also stale from my own change: the CSRF-exemption comment at `app.py:420-422` still describes validation as "active-session match + Zerodha checksum" and omits the URL secret.

PR #1707, opened 2026-07-30
fix(streaming): correct reconnect counter lifecycle in two adapters

punksterlabs:
Two mirror-image bugs in the same class, submitted together because **the obvious fix for the first introduces the second**.

### 1. `websocket_proxy/order_adapter.py` — the backoff counter never resets

`attempt` is initialised once *outside* the `while` loop and incremented on every return from `_connect_once()` — including a clean disconnect after a session that streamed healthily for hours. Nothing ever resets it; `on_open` only logs and starts the heartbeat.

With `_RECONNECT_BACKOFFS = [1, 2, 5, 10, 30, 60]`:

| disconnect | delay |
|---|---|
| 1-5 | 1s, 2s, 5s, 10s, 30s |
| 6+ | **60s, permanently** |

So six unrelated disconnects over the adapter's lifetime pin every subsequent order-feed recovery at 60 seconds, and users miss timely fill and rejection updates because of failures that already resolved.

**Fix:** reset the ladder when a session lasted at least `_HEALTHY_SESSION_SECONDS`.

Gating on session **duration** rather than on the socket opening is the whole point. A broker that accepts the TCP/WebSocket handshake and only then rejects at the application layer would reset the counter on *every* attempt under an `on_open`-based fix, turning the backoff into a hot reconnect loop against a dead credential.

That failure mode is not hypothetical — it is bug 2.

Verified the ladder still behaves for genuine streaks:

```
6 rapid failures:                  [1, 2, 5, 10, 30, 60]   (unchanged)
mixed 3 fails, 1 healthy, 2 fails: [1, 2, 5, 1, 2, 5]
5 healthy hours then a blip:       1s   (was 60s)
```

### 2. `broker/flattrade/streaming/flattrade_websocket.py` — `connect()` returns before the broker accepts

`_on_open` sets `self.connected = True` and *then* sends authentication. `_wait_for_connection` polls only that flag, never inspecting the `ak` response, so `connect()` means "socket opened", not "broker accepted us".

`flattrade_adapter` resets both `auth_refresh_retries` and `reconnect_attempts` on a truthy `connect()`. On an expired token the sequence is:

1. socket opens -> `connect()` returns True -> `auth_refresh_retries = 0`
2. broker sends the auth nack -> `_handle_auth_failure` -> `auth_refresh_retries = 1`
3. reconnect in ~5s -> back to step 1

The counter oscillates 0 -> 1 -> 0 and never reaches `max_auth_refresh_retries = 3`. `reconnect_attempts` never reaches `MAX_RECONNECT_ATTEMPTS = 10` either. Both termination conditions are defeated, so the adapter hammers a dead token every ~5 seconds indefinitely — contradicting the module's own stated rationale:

> stops well short of the full 10-attempt generic backoff so a genuinely dead token is not hammered ... against a possibly rate-limited IP (SEBI static-IP mandate)

**Fix:** track broker acceptance separately from socket open, wait for that, and return early on an explicit rejection.

`self.connected` is deliberately left alone — `_send_message` and `_send_authentication` depend on it being true at socket open. Only the meaning of `connect()`'s return value changes, which makes the adapter's existing reset correct as written and lets the bound actually terminate. Note the fix belongs here rather than at the adapter's reset line, which is where the symptom appears.

There is a narrow race where the nack and close land inside a single 100ms poll window, so `connected` is already false and no reset happens. That does not rescue the bound — terminating would require winning that race three times consecutively.

### Scope

I run neither Flattrade nor a deployment that consumes the order-update feed heavily; this is code-level analysis from auditing a 2.0.1.7 upgrade. Happy to split into two PRs if you would rather review them separately — I kept them together only because of the coupling described above.

Note: `ruff` reports a pre-existing `UP035` (`typing.Dict` deprecated) at `flattrade_websocket.py:11`, unrelated to this change.

---
## Summary by cubic
Fix reconnect/backoff behavior in the order feed and Flattrade WebSocket. Healthy sessions now reset backoff, and connect() only succeeds after broker auth, preventing permanent 60s delays and infinite token retries.

- **Bug Fixes**
  - Order adapter: reset the backoff ladder after any session lasting ≥60s; based on session duration to avoid hot reconnect loops when the broker rejects at the app layer.
  - Flattrade WebSocket: add `auth_ok` and wait for broker acceptance in `connect()`; return early on explicit auth rejection so auth/reconnect counters reach their limits.

<sup>Written for commit 77c5575943303edc447980ff9552167f16318bb7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1707?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/broker/flattrade/streaming/flattrade_websocket.py
+++ b/broker/flattrade/streaming/flattrade_websocket.py
@@ -78,6 +78,7 @@ class FlattradeWebSocket:
         # The adapter checks this after _on_close/_on_error to route into a
         # tighter-bounded auth-retry path instead of the generic reconnect loop.
         self.auth_failed = False
+        self.auth_ok = False
         self.auth_failure_message = ""
 
         # Set by stop() to interrupt any in-progress sleep/wait loop immediately
@@ -125,6 +126,7 @@ class FlattradeWebSocket:
         # Reset auth failure state so a stale flag from a previous failed
         # attempt doesn't block this fresh connection attempt.
         self.auth_failed = False
+        self.auth_ok = False
         self.auth_failure_message = ""
 
         self.ws = websocket.WebSocketApp(
@@ -148,9 +150,19 @@ class FlattradeWebSocket:
         start_time = time.time()
 
         while time.time() - start_time < self.CONNECTION_TIMEOUT:
-            if self.connected:
-                self.logger.info("WebSocket connected successfully")
+            # Wait for the broker's 'ak' acknowledgement, not merely the socket
+            # opening. _on_open sets self.connected before authentication is even
+            # sent, so returning on that would report success for a rejected
+            # token -- and the caller resets its bounded auth-retry counter on
+            # success, so the bound would never be reached.
+            if self.auth_ok:
+                self.logger.info("WebSocket connected and authenticated successfully")
                 return True
+            if self.auth_failed:
+                self.logger.error(
+                    f"Broker rejected authentication: {self.auth_failure_message}"
+                )
+                return False
             if self._stop_event.wait(0.1):
                 self.logger.info("Wait for connection interrupted by stop()")
                 return False
@@ -291,6 +303,7 @@ class FlattradeWebSocket:
             bool: True (message handled)
         """
         if data.get("s") == self.AUTH_SUCCESS:
+            self.auth_ok = True
             self.logger.info("Authentication successful")
         else:
             self.logger.error(f"Authentication failed: {data}")
--- a/websocket_proxy/order_adapter.py
+++ b/websocket_proxy/order_adapter.py
@@ -30,6 +30,12 @@ from utils.logging import get_logger
 # feed.py::_reconnect_loop pattern for consistency across the codebase.
 _RECONNECT_BACKOFFS = [1, 2, 5, 10, 30, 60]
 
+# A session that stayed connected at least this long counts as healthy, and its
+# disconnect starts a fresh backoff ladder rather than continuing the old one.
+# Comfortably longer than the whole ladder's early rungs, so a genuine
+# connect-fail-retry streak cannot accidentally qualify.
+_HEALTHY_SESSION_SECONDS = 60
+
 
 def to_openalgo_symbol(broker_symbol: str, exchange: str, token=None) -> str:
     """Best-effort mapping of a broker symbol to OpenAlgo symbol format.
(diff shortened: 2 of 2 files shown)


punksterlabs:
All three findings are valid. Moved to draft. The P1 is a genuine regression from this PR, and it is worse than "a delay".

**P1 — waiting for auth while the adapter lock blocks `on_open`.**

`_attempt_reconnection` holds `self.lock` (`flattrade_adapter.py:883`) across `ws_client.connect()` (line 926). `websocket-client` invokes `on_open` on the same thread that then reads frames — `_app.py` calls `_callback(self.on_open)` and immediately follows with `dispatcher.read(...)`, so a blocking `on_open` stalls all frame delivery. `FlattradeWebSocket._on_open` fires the external callback at `flattrade_websocket.py:235`, which reaches `FlattradeWebSocketAdapter._on_open` -> `_resubscribe_all()` -> `with self.lock` (line 939) and parks.

So the auth frame goes out, the broker replies `ak`, and the reply is never dequeued. My new wait burns the full 15s `CONNECTION_TIMEOUT`, then `_wait_for_connection`'s `stop()` adds another 5s on `ws_thread.join`. The failure path does not reset `reconnect_attempts`, so after `MAX_RECONNECT_ATTEMPTS = 10` the adapter sets `running = False` and streaming never recovers from a single disconnect.

Pre-PR this was benign only because the wait returned on `self.connected`, which is set at line 228 before authentication is sent — it never waited on a callback. The initial `connect()` path holds no lock, which is why only timer-driven reconnects break.

Fixing at the source rather than only at the lock: fire the external `on_open` from `_handle_auth_response` after `auth_ok = True`, not from `_on_open`. The reader thread then never blocks before publishing the ack, the waiter returns within one 100ms poll, and resubscription happens after the session is authenticated rather than before it — which is the more correct ordering anyway. I will also release the adapter lock around the blocking `connect()` and reacquire it only to update shared state, since holding it across network I/O also blocks `_find_matching_subscriptions` on the message path.

**P2 — sticky `auth_ok`.** Correct. `auth_ok` is cleared only in `_initialize_connection`; `_on_close` and `stop()` clear `connected` but leave it set, and the poll checks `auth_ok` before the stop-event check. An ack followed by a drop inside the same 100ms window returns True for a dead socket, and `_attempt_reconnection` then zeroes both counters — the exact eager reset this PR exists to remove. Taking the suggestion: `if self.auth_ok and self.connected:`, plus clearing `auth_ok` in `_cleanup_connection_state()`. No false negatives, since `connected` is set before authentication is sent.

**P2 — healthy-session timing starts before the socket opens.** Valid, and trivially reachable. `_connect_once()` calls `get_ws_url()` first, and both the Upstox and AliceBlue order adapters issue an HTTP authorize request through `get_httpx_client()`, which is built with `timeout=120.0` (`utils/httpx_client.py:187`). Upstox swallows the failure and falls back, so the attempt continues. A stalled authorize burns >60s before any socket exists, the gate fires, and every later retry uses the 1s rung — a hot loop against an endpoint that never established a session, which is worse than the bug I set out to fix. Fixing by recording `_opened_at` in `on_open` and resetting only when `opened_at is not None and elapsed >= _HEALTHY_SESSION_SECONDS`.

None of these three weaken the PR's intent — each makes the success/reset condition strictly narrower, so socket-open alone still never resets the ladder. I'll also add unit coverage; `test/test_order_update_adapters.py` currently only asserts the backoff bounds, which is why none of this was caught.

PR #1712, opened 2026-07-30
fix(positions): stop overwriting broker-supplied P&L on live positions

punksterlabs:
### Problem

`useLivePrice` recomputes P&L for every open position on each tick:

```ts
unrealizedPnl = (currentLtp - avgPrice) * qty * lotSize
calculatedPnl = todayRealizedPnl + unrealizedPnl
```

That is the **sandbox's** P&L model, and only the sandbox reports P&L that way. A live broker's P&L comes from the exchange and can follow entirely different accounting, so recomputing overwrites a correct figure with a wrong one.

There is no guard: `currentLtp` falls back to `item.ltp`, so this runs for any open position even with no live data flowing. Closed positions (`qty === 0`) return early, which is why it is easy to miss when spot-checking.

### Delta Exchange options are a clear counter-example

Raw `/v2/positions/margined` from a live account:

| symbol | size | entry | mark | Delta `unrealized_pnl` | `(mark−entry)×cv` | `−(mark×cv)` |
|---|---|---|---|---|---|---|
| P-BTC-63800-310726 | 1 | 773 | 28.41 | **−0.028** | −0.745 | −0.0284 |
| C-BTC-63600-310726 | 1 | 903 | 1367.52 | **−1.364** | +0.465 | −1.3675 |
| P-BTC-64400-010826 | 1 | 1045 | 289.15 | **−0.288** | −0.756 | −0.2891 |

Delta tracks **`−(mark × contract_value)`** and ignores `entry_price` entirely. No `(ltp − avg)` formula reproduces it — solving for a multiplier per row gives `0.0000376` for the first and `−0.002936` for the second, so it disagrees in **sign as well as magnitude**.

Across 8 open positions the page rendered **−0.72** against the exchange's **−6.86**, with calls displayed as gains while they were losses.

### Fix

Make the recomputation opt-in (`recalculatePnl`, default off) and enable it only in analyzer mode, where the sandbox owns the same model and a per-tick refresh is faithful.

Live positions keep the broker's P&L and refresh only the displayed LTP. They therefore update on the REST poll rather than every tick — slightly staler, but correct. That trade is clearly worth making against a number that is wrong in sign.

Applied to both `Positions.tsx` and `Holdings.tsx`, which share the hook.

### What this replaces

An earlier revision of this PR tried to preserve a realized component — first by deriving it, then by plumbing `today_realized_pnl` through the Delta and Dhan mappings. Both were wrong diagnoses. Delta reports `realized_pnl = 0` on all of these positions, so realized was never the gap; the formula itself is.

The derivation approach was also unsound for a reason raised in review: a broker whose P&L and whose `ltp` come from separate requests — true of both the Delta and Dhan mappings — would have the intervening price move misattributed to realized.

### Verification

`tsc --noEmit -p tsconfig.app.json` clean; `npm run test:run` 170/170.

`biome check` reports a formatting failure on the touched files, but does so **identically on the unmodified upstream files** in my checkout — CRLF line endings from a Windows working tree, same exit code for both. The committed diff is 28 insertions / 3 deletions.

### Note

Related to #1711 (P&L% denominator), which is independent and still applies — it fixes the sandbox path this PR keeps enabled.

### DIFF
--- a/frontend/src/hooks/useLivePrice.ts
+++ b/frontend/src/hooks/useLivePrice.ts
@@ -34,6 +34,9 @@ export interface UseLivePriceOptions {
   multiQuotesRefreshInterval?: number
   /** Pause WebSocket and polling when tab is hidden (default: true) */
   pauseWhenHidden?: boolean
+  /** Recompute P&L from the live tick. Only correct when the caller's P&L
+   *  model matches this hook's (the sandbox). Live broker P&L is left alone. */
+  recalculatePnl?: boolean
   /** Time in ms to wait before pausing when hidden (default: 5000) */
   pauseDelay?: number
 }
@@ -86,6 +89,7 @@ export function useLivePrice<T extends PriceableItem>(
     useMultiQuotesFallback = true,
     multiQuotesRefreshInterval = 30000,
     pauseWhenHidden = true,
+    recalculatePnl = false,
   } = options
 
   const { apiKey } = useAuthStore()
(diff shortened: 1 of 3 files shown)


punksterlabs:
Pushed a rework in response to review feedback on the same change downstream.

The original version derived the realized component by subtracting the unrealized P&L implied by `item.ltp` from `item.pnl`. That is only sound if both come from the same request, and they do not: the Delta mapping sets `lastTradedPrice` from a separate bulk `/v2/tickers` fetch, and the Dhan mapping takes `ltp` from a separate `ltp_map`. Any price movement between the two calls would be booked as realized, and re-derived from the same stale anchor on every tick.

Both brokers already report the realized half directly, so derivation was never needed for them. `broker/deltaexchange/mapping/order_data.py` now emits `today_realized_pnl = realized_pnl` and `broker/dhan/mapping/order_data.py` emits `today_realized_pnl = realizedProfit`. The hook prefers an explicitly reported value and falls back to subtraction only when a source gives a combined P&L with no realized field -- losing the realized component outright is the worse failure, so the fallback stays, but it is no longer the primary path.

The remaining brokers (zerodha, angel, fyers, upstox) still report neither, so they take the fallback. If you would prefer those wired up explicitly too I am happy to extend it, but I did not want to touch mappings I cannot test against a live account.

Naming caveat: Delta's `realized_pnl` is for the life of the position, not today. I reused the frontend's existing field name rather than introducing a new one; what it means at the point of use is the part of P&L that does not move with price.

ruff clean, tsc clean, frontend tests pass.

PR #1723, opened 2026-08-01
fix(upstox): return empty funds on service-hours error instead of fabricated zeros

linuxsmiths:
### Problem

`broker/upstox/api/funds.py::get_margin_data()`'s 423 Locked / `UDAPI100072` handler (the funds service is outside its 5:30 AM - 12:00 AM IST operating hours) returns a **fabricated zero-balance dict** (`availablecash "0.00"`, `collateral "0.00"`, …) instead of propagating an error.

That makes the service-hours window **indistinguishable from a genuinely live, zero-balance account** — the exact same masking bug already fixed for Fyers in #1624: any caller that treats a funds "success" as "the broker session is live" is misled, and an expired/invalid session that happens to hit this window would be reported as healthy.

### Fix

Return `{}` (empty) instead of the fabricated dict, matching the empty-is-error contract already used elsewhere (`services/funds_service.py`, #1597). Still logs the service-hours condition for diagnostics — only the fabricated account data is removed.

### Notes

Pairs with #1597 (treat an empty broker funds response as an error, not success) and #1624 (identical fix already merged for Fyers).

---
## Summary by cubic
Stop fabricating zero balances for Upstox funds when the funds service is outside hours; return {} to signal error. This aligns with the empty-is-error contract and avoids misreporting an expired or invalid session as healthy during 5:30 AM–12:00 AM IST.

<sup>Written for commit d7d64ee6a451a25e477b93b03484d4dcdbb7ee78. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1723?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/broker/upstox/api/funds.py
+++ b/broker/upstox/api/funds.py
@@ -88,7 +88,15 @@ def get_margin_data(auth_token):
     except httpx.HTTPStatusError as e:
         response_text = e.response.text
 
-        # Check if it's a service hours error (423 Locked)
+        # Check if it's a service hours error (423 Locked). Previously this
+        # fabricated a fake all-zero "success" dict, which made the funds
+        # service hours window (5:30 AM - 12:00 AM IST) indistinguishable from
+        # a genuinely live, zero-balance account — the exact same masking bug
+        # patched for Fyers (services/funds_service.py's "empty == error"
+        # contract expects {} on any error/no-data path, never fabricated
+        # values). Log it distinctly for diagnostics but still return {} so an
+        # expired session during this window is correctly reported as invalid,
+        # not silently treated as live.
         if e.response.status_code == 423:
             try:
                 error_data = json.loads(response_text)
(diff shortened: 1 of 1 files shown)


slegarraga (changes requested):
Thanks for removing the fabricated zeros; that is the right direction. One blocking issue remains: current `main` in `services/funds_service.py` wraps any non-raising `get_margin_data` result as success, so returning `{}` still produces `{"status":"success","data":{}}` and callers can treat the service-hours window as live. #1597 is still open and covers exactly that empty-is-error contract. Please either land this with #1597 together or include the empty response check in this PR, plus a test that Upstox service-hours returns `{}` and the service layer reports an error. Happy to re-review after that.

marketcalls (commented):
Review score: 88/100 at `d7d64ee6a451a25e477b93b03484d4dcdbb7ee78`. Changes are needed before approval and merge.

- `broker/upstox/api/funds.py:110`: The intended empty-is-error service contract is still absent. The new {} result is wrapped as HTTP200 status=success by services/funds_service.get_funds_with_auth, so API clients still receive a successful response for unavailable funds. Land the empty-result guard with this change (or its dependency #1597), report unavailable funds without inventing a balance, and test the real adapter-to-service error path.

PR #1725, opened 2026-08-01
fix: Upstox multiquotes never populated prev_close, breaking %-change

linuxsmiths:
### Problem
`BrokerData.get_quotes()` (single-symbol) already has a fallback chain for `prev_close`: it prefers the v2 `/market-quote/quotes` endpoint's `ohlc.close`, falling back to v3 OHLC's `prev_ohlc.close` only if that's empty. The batch path used by `/api/v1/multiquotes` — `_process_quotes_batch()` — never got this fallback: it fetches `v2_quotes` for bid/ask/OI, but reads `prev_close` ONLY from v3's `prev_ohlc.close`.

On accounts where Upstox's v3 batch OHLC response has an empty `prev_ohlc` (confirmed live), every symbol in a multiquotes response comes back with `prev_close: 0`, even though the single-symbol quotes endpoint returns the correct value for the very same instrument. Any consumer that derives a day change % from multiquotes (screeners, movers, a market ticker, etc.) sees every row/tick as "no change" or filters the row out entirely when it guards against a zero prev_close — even though live prices are flowing fine.

### Fix
Mirror `get_quotes()`'s fallback in `_process_quotes_batch()`: read `prev_close` from the already-fetched v2 quote's `ohlc.close` first, falling back to v3's `prev_ohlc.close` if the v2 value is unavailable.

### Notes
Verified live against a real Upstox session: the same symbol via `/api/v1/quotes` returned a correct non-zero `prev_close`, while `/api/v1/multiquotes` returned `prev_close: 0` for it (and every other requested symbol) before this fix; after the fix both paths agree.

---
## Summary by cubic
Add a `prev_close` fallback in `_process_quotes_batch()` for Upstox multiquotes: prefer v2 `ohlc.close`, then fall back to v3 `prev_ohlc.close`. This fixes `prev_close: 0` in batch responses and restores correct %-change in `/api/v1/multiquotes`.

<sup>Written for commit b4ce98dcfbfdb0b0f5f18b0c6829265fb3869b21. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1725?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/broker/upstox/api/data.py
+++ b/broker/upstox/api/data.py
@@ -459,7 +459,7 @@ class BrokerData:
             logger.error(f"API Error: {error_msg}")
             raise Exception(f"API Error: {error_msg}")
 
-        # Also fetch v2 quotes for bid/ask/OI data
+        # Also fetch v2 quotes for bid/ask/OI data (and prev_close — see below)
         v2_quotes = {}
         try:
             client = get_httpx_client()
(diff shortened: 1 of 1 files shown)


slegarraga (approved):
Good mirror of the single-symbol fallback. The `or` chain handles missing or zero v2 close before falling back to v3, and the default path remains unchanged. One non-blocking note: a small unit test with mocked v2/v3 payloads, covering both fallback directions, would pin this regression. Looks good to me.

Issue #1713, opened 2026-07-31
Issue: Getting Telegram DB Locked error during startup


aravindgandavadi:
Hi @marketcalls @Kalaiviswa 

I am getting these error messages during OpenAlgo instance startup. This happens in both Zerodha (live) and Upstox (Analyser mode). I run my instances in Azure/Ubuntu/Docker. I used script [install-docker-multi-custom-ssl.sh](https://raw.githubusercontent.com/marketcalls/openalgo/main/install/install-docker-multi-custom-ssl.sh) to update the latest version (6ab4222). I am also attaching the diagnostics file: [openalgo-system-report-20260731-083003.txt](https://github.com/user-attachments/files/30571877/openalgo-system-report-20260731-083003.txt)

Please analyse the issue.

[2026-07-31 08:29:21,881] ERROR in telegram_db: Failed to update bot config: (sqlite3.OperationalError) database is locked
[SQL: UPDATE bot_config SET token=[REDACTED], is_active=?, updated_at=CURRENT_TIMESTAMP WHERE bot_config.id = ?]
[parameters: ('gAAAAABqbA-EVtE6XOopzoKxsYlXwOrstPtUJa_9VAt28yJu5Qec-fJOAugzanEjoLPr7_J2E3T5V4ZlUYE-YlbqjpUp2muUY3GmfyyMiiFyVOayNyGbG-tnIUsU7mhL4zl3Wo1AB7Kk', 0, 1)]
(Background on this error at: https://sqlalche.me/e/20/e3q8)
Traceback (most recent call last):
  File "/app/.venv/lib/python3.12/site-packages/sqlalchemy/engine/base.py", line 1969, in _exec_single_context
    self.dialect.do_execute(
  File "/app/.venv/lib/python3.12/site-packages/sqlalchemy/engine/default.py", line 952, in do_execute
    cursor.execute(statement, parameters)
sqlite3.OperationalError: database is locked
The above exception was the direct cause of the following exception:
Traceback (most recent call last):
  File "/app/database/telegram_db.py", line 520, in update_bot_config
    db_session.commit()
  File "/app/.venv/lib/python3.12/site-packages/sqlalchemy/orm/scoping.py", line 596, in commit
    return self._proxied.commit()
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/app/.venv/lib/python3.12/site-packages/sqlalchemy/orm/session.py", line 2035, in commit
    trans.commit(_to_root=True)
[...]

PR #1734, opened 2026-08-03
Fix "database is locked" during startup (SQLite parallel initialization)

vishalrateria:
Fixes random SQLite “database is locked” errors during startup by serializing DB setup and cleaning stale sessions. The previous PR accidentally included `use_reloader=False` which disabled hot-reloading; this new PR strictly isolates the DB startup fix.

---
## Summary by cubic
Serialize database initialization during startup to prevent random SQLite “database is locked” errors. DB init now runs sequentially; hot-reload behavior is unchanged.

- **Bug Fixes**
  - Run all DB setup via a single-thread `ThreadPoolExecutor` (max_workers=1) to avoid concurrent SQLite access during initialization.

<sup>Written for commit f708aaa1f340ebf7a4b8185a4b13e3f3b99611cb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1734?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app.py
+++ b/app.py
@@ -713,7 +713,7 @@ def setup_environment(app):
             ]
 
             db_init_start = time.time()
-            with ThreadPoolExecutor(max_workers=15) as executor:
+            with ThreadPoolExecutor(max_workers=1) as executor:
                 futures = {executor.submit(func): name for name, func in db_init_functions}
                 for future in as_completed(futures):
                     db_name = futures[future]


marketcalls:
Merged - thanks for splitting this out cleanly from the earlier attempt.

I validated it by reproducing the contention against the real init functions rather than by inspection, and the diagnosis holds: 14 of the 20 startup functions target the same file (`openalgo.db`), and SQLite permits one writer per file, so the 15 workers were contending for the write lock rather than progressing in parallel.

With `busy_timeout` forced low so the contention surfaces instead of being waited out:

| | 15 workers | 1 worker |
| --- | --- | --- |
| Cold first-run | 4 lock errors / 7 trials | 0 |
| Warm restart | 7 lock errors / 9 trials | 0 |

The `busy_timeout=15000` added in #1713 was masking the collision by waiting on it, and it does not cover `SQLITE_BUSY_SNAPSHOT`, which `create_all`'s read-then-write pattern can hit and which fails immediately regardless of timeout. So serializing is the stronger fix.

It also turned out to cost nothing. The warm restart path - what every existing install does on every restart - measures faster serialized, 99ms vs 138ms, because the parallelism was never real and the threads only added contention. Cold first-run costs about 13ms, once.

One note for anyone reading this later: the description above mentions cleaning stale sessions, but the merged diff is only the `max_workers` change in `app.py` - that wording carried over from the earlier PR.

Follow-up in 413534d8c corrects the wording the change left stale (the completion log still said "initialized in parallel", and a nearby comment still blamed "parallel init"), switches that loop's `logger.error` to `logger.exception` per CLAUDE.md so a failed DB init reaches errors.jsonl with its traceback, and records why `max_workers=1` is deliberate so it does not get optimised back to a thread pool later.

PR #1798, opened 2026-08-17
Fix symbol cache: key primary store by (exchange, token)

### LINKED ISSUE #1774: AliceBlue Symbol cache silently drops ~980 NSE equities whose broker token collides across exchanges
# Bug: Symbol cache silently drops symbols whose broker token repeats across exchanges

## Summary

`BrokerSymbolCache.load_all_symbols()` in `database/token_db_enhanced.py` keys the primary symbol store by `token` alone. Broker tokens are only unique **within** an exchange, not globally — so when a broker reuses tokens across exchanges (e.g. AliceBlue uses small integers on both NSE and CDS), later-loaded rows overwrite earlier ones. On our AliceBlue install this silently drops **982 NSE equities** (RELIANCE, INFY, HDFCBANK, SBIN, …) from every unfiltered symbol search.

**Affects:** Shared, broker-agnostic code — any broker whose master contract reuses tokens across exchanges is affected.

**Patch attached:** `issue-1-symbol-cache.diff` (2 lines changed).
**Regression test attached:** `test/test_symbol_cache_token_collision.py` (4 tests, fail on current `main`).

## Environment

- **OpenAlgo version:** 2.0.1.8
- **Commit:** `087a20e4cbe8ff96af3a730610ea07334f64ce59` (`main`)
- **Broker:** AliceBlue (live account)
- **Python:** 3.14.6
- **OS:** Windows 10 Home 19045
- **Symbol master:** 165,690 rows (NFO 75,856 · BFO 39,212 · MCX 16,047 · BSE 12,931 · CDS 11,260 · BCD 7,560 · NSE 2,710 · NSE_INDEX 57 · BSE_INDEX 53 · MCX_INDEX 4)

## Symptom

On `/trading`, typing `RELIANCE` in the symbol search returns 500 rows — BSE cash, BFO options, NFO futures and options — but **no NSE row at all**. Same for INFY, HDFCBANK, and ~980 other NSE stocks. The stock is simply not selectable on NSE.

Exchange-filtered search (`/search`, option chain, tools) works fine, which makes this easy to miss.

## Root cause

```python
# BrokerSymbolCache.load_all_symbols()
self.symbols[sym.token] = symbol_data          # keyed by token alone — collisions!
```

AliceBlue issues small integer tokens on both NSE and CDS, so they collide:

| Token | NSE symbol | CDS symbol (overwrites NSE) |
|-------|-----------|---------------------------|
| 2885.0 | RELIANCE | USDINR28SEP26100.25PE |
| 1594.0 | INFY | GBPINR21AUG26121.75PE |
| 1008.0 | FACT | EURINR28JUN27114CE |

Rows load in table order; CDS contracts sit later and overwrite the NSE equities. The cache ends up with **164,706 entries from 165,690 rows — 984 lost** (982 NSE, 2 BSE).

Verified on the live database:

```sql
SELECT COUNT(*) FROM symtoken;                    -- 165690
SELECT COUNT(DISTINCT token) FROM symtoken;       -- 164706 (984 lost)
```

### Why only unfiltered search breaks

The secondary indexes (`by_symbol_exchange`, `by_token_exchange`, `by_exchange`, etc.) are all correctly keyed by `(exchange, token)`, so `get_token()`, order placement, and any exchange-scoped search work normally. Only the primary `self.symbols` dict is collapsed, and only unfiltered search iterates it:

```python
# In search_symbols():
if exchange and exchange in self.by_exchange:
    symbols_to_search = self.by_exchange[exchange]     # correct — fine
else:
    symbols_to_search = self.symbols.values()          # collapsed — rows missing
```

The `/trading` terminal's `SymbolSearchDialog.tsx` calls `search(q, undefined, 500)` (unfiltered by design), so it always hits the broken path.

## Reproduction

```python
from database.token_db_enhanced import get_cache
c = get_cache(); c.load_all_symbols('aliceblue')
print(len(c.symbols))                                    # 164706, expected 165690
rows = c.search_symbols('RELIANCE', None, limit=500)
print([(r.symbol, r.exchange) for r in rows[:3]])
# [('RELIANCE','BSE'), ('RELIANCE24thSEPFUT','BFO'), ('RELIANCE25AUG26FUT','NFO')]
# NSE RELIANCE absent
```

## Fix

Key the primary store by `(exchange, token)`, matching every other index in the class:

```diff
-        self.symbols: dict[str, SymbolData] = {}
+        self.symbols: dict[tuple[str, str], SymbolData] = {}

-                self.symbols[sym.token] = symbol_data
+                self.symbols[(sym.exchange, sym.token)] = symbol_data
```

`self.symbols` is used only inside the class (store, `len()`, the two search iterations, `clear_cache()`), so nothing outside needs to change. No measurable memory cost — the 984 `SymbolData` objects were already alive in the other indexes; only the reference from `self.symbols` was dropped.

### After the fix

```
cache entries: 165690  (was 164706)

RELIANCE  ->  [('RELIANCE','NSE'), ('RELIANCE','BSE'), ('RELIANCE24thSEPFUT','BFO')]
INFY      ->  [('INFY','NSE'),     ('INFY','BSE'),     ('INFY24thSEPFUT','BFO')]
```

## Which brokers are affected?

The defect is in shared code. Any installation can be checked with:

```sql
SELECT COUNT(*) FROM (
  SELECT token FROM symtoken GROUP BY token HAVING COUNT(DISTINCT exchange) > 1
);
-- 0 = unaffected. Anything above 0 = that many symbols missing from search.
-- AliceBlue: 984
```

Confirmed on AliceBlue. Worth checking Angel and any other broker whose tokens are per-exchange rather than globally unique.

## Note

`by_token` / `get_symbol_data(token)` is also keyed by token alone and is still ambiguous across exchanges. We left it untouched — it's an explicitly exchange-less lookup, so it's a separate design question.

## Patch

```diff
diff --git a/database/token_db_enhanced.py b/database/token_db_enhanced.py
index ac3cd3705..c687fcad5 100644
--- a/database/token_db_enhanced.py
+++ b/database/token_db_enhanced.py
@@ -147,8 +147,14 @@ class BrokerSymbolCache:
         self.active_broker: str | None = None
         self.cache_loaded: bool = False
 
-        # Primary storage - all symbols in memory
-        self.symbols: dict[str, SymbolData] = {}
+        # Primary storage - all symbols in memory, keyed by (exchange, token).
+        # Broker tokens are only unique *within* an exchange: AliceBlue hands out
+        # small integer tokens on both NSE and CDS, so ~980 NSE equities
+        # (RELIANCE 2885, INFY 1594, ...) share a token with a CDS option
+        # contract. Keying on the token alone silently dropped whichever row
+        # loaded first, and the unfiltered search paths below iterate this dict —
+        # so those stocks vanished from symbol search entirely.
+        self.symbols: dict[tuple[str, str], SymbolData] = {}
 
         # Multi-index maps for O(1) lookups
         self.by_symbol_exchange: dict[tuple[str, str], SymbolData] = {}
@@ -249,7 +255,7 @@ class BrokerSymbolCache:
                 )
 
                 # Store in primary dict
-                self.symbols[sym.token] = symbol_data
+                self.symbols[(sym.exchange, sym.token)] = symbol_data
 
                 # Build indexes
                 self.by_symbol_exchange[(sym.symbol, sym.exchange)] = symbol_data
```

## Regression test

`test/test_symbol_cache_token_collision.py` — 4 tests that load a cache from four stub rows (two NSE equities and two CDS options sharing their tokens) and assert all four survive the load, that unfiltered search returns the NSE rows and ranks the exact match first, that the CDS rows are not evicted in turn, and that the scoped indexes still resolve. Fails on current `main`, passes with the patch.

[test_symbol_cache_token_collision.py](https://github.com/user-attachments/files/30844588/test_symbol_cache_token_collision.py)

### LINKED ISSUE #1778: Symbol cache: key primary store by (exchange, token) - shared-code fix, validation and test plan
Tracking issue for the shared-code change behind #1774. The bug report there is accurate and self-contained; this issue exists because the fix lands in `database/token_db_enhanced.py`, which is common to all 36 broker plugins and backs symbol search, the option-chain and analytics services, Historify and sandbox. It needs a validation plan rather than a two-line push.

**Not started.** No code has been changed.

## Current state, measured

The defect is intermittent, and right now it is not manifesting. Live AliceBlue install, master contract downloaded 2026-08-08 17:36:

```
total rows                    164695
distinct token                164695
distinct (exchange, token)    164695

tokens appearing on >1 exchange     0
duplicate (exchange, token) pairs   0
empty/null tokens                   0
```

The three collisions documented in #1774 are all clean today:

| Token | When reported | Now |
| --- | --- | --- |
| 2885 | NSE RELIANCE overwritten by CDS USDINR28SEP26100.25PE | NSE RELIANCE only |
| 1594 | NSE INFY overwritten by CDS GBPINR21AUG26121.75PE | NSE INFY only |
| 1008 | NSE FACT overwritten by CDS EURINR28JUN27114CE | NSE FACT only |

This is not AliceBlue changing their token scheme. CDS still issues tokens straight through the NSE equity range - it currently holds 1,579 contracts below token 3000, starting at 1001 (`EURINR16OCT26FUT`, `EURINR21AUG26105.25CE`, `EURINR28APR27FUT`, ...). Slots are recycled as contracts expire and new strikes list. **1,476 of 2,710 NSE equities (54%) have tokens inside the currently-occupied CDS band 1001-17274.**

The reporter caught it on a day it was firing: 984 rows lost, 982 NSE and 2 BSE.

The practical consequence for whoever picks this up: **you will probably not be able to reproduce it on demand.** A "cannot reproduce, closing" outcome here would be wrong. Validation has to use a synthetic fixture, not live data.

### 1. Key the primary store by (exchange, token)

`database/token_db_enhanced.py`, the one functional line plus its annotation:

```diff
  # line 151
- self.symbols: dict[str, SymbolData] = {}
+ self.symbols: dict[tuple[str, str], SymbolData] = {}

  # line 252, in load_all_symbols()
- self.symbols[sym.token] = symbol_data
+ self.symbols[(sym.exchange, sym.token)] = symbol_data
```

Broker tokens are unique within an exchange, not globally. Every other index in the class is already keyed `(exchange, token)`; this line is the odd one out.

### 2. Assert the load invariant

This is the more important half, and it is missing today. The cache silently accepted losing 984 rows. Nothing logged, nothing failed, and the only symptom was a stock missing from one search box.

After `load_all_symbols()`, assert that the number of cache entries equals the number of rows read, and log loudly on mismatch. Had this existed, #1774 would have been a one-line log entry instead of a user-visible mystery. Consider surfacing the count in `get_cache_info()` / `/master_contract_status` so drift is visible without a database query.

### 3. Remove `get_symbol_data()` and `by_token`

Lines 157, 258, 410-414, 701. `by_token` is keyed by token alone and carries exactly the ambiguity being fixed. `get_symbol_data()` has **no callers anywhere in the repository**. Deleting it removes the trap rather than leaving it for someone to find and use. #1774 deliberately left this alone as a separate question - this is that question.

### The no-op property for unaffected brokers

The claim to verify: for a broker with no colliding tokens, the change is unobservable. The argument is structural - `self.symbols` is used in only four places beyond the assignment, and none read the key:

```
line 292   len(self.symbols) * 500      memory estimate
line 494   self.symbols.values()        search_symbols
line 600   self.symbols.values()        fno_search_symbols
line 697   self.symbols.clear()
```

With no collisions, `token` and `(exchange, token)` are in 1:1 correspondence, so `.values()` yields the same objects in the same order. Confirm this holds by test, not by assertion in a comment.

### Ranking is unaffected

`search_symbols` scores every match and then takes `heapq.nsmallest(limit, ...)`, so results are fully ranked rather than truncated in iteration order. A recovered NSE row scores 0 (exact match) and should rank first, displacing a low-ranked option contract. Iteration order survives only as `seq`, the last tie-break after score, symbol length and symbol text. `fno_search_symbols` sorts as well. Both need a test that a recovered row lands in the results at the right rank, and that unaffected queries return an identical ordered list.

### Which brokers are exposed

One query per install, run after a master-contract download:

```sql
SELECT COUNT(*) FROM (
  SELECT token FROM symtoken GROUP BY token HAVING COUNT(DISTINCT exchange) > 1
);
```

Noren-family brokers (AliceBlue, Shoonya, Flattrade, Zebu) and Angel issue small per-exchange integers and are the likely candidates. Zerodha instrument tokens are globally unique and are structurally immune.

Note the survey produces false negatives depending on the day it runs - see the measurement above. A zero result means "not colliding today", not "not affected".

### Synthetic fixture

Because live data cannot be relied on to collide, the regression test must construct a symbol set with a deliberate cross-exchange token collision and assert no rows are lost. #1774 has four such tests attached that fail on current `main`.

### Memory

`len(self.symbols)` grows by the number of recovered rows on affected installs. Those `SymbolData` objects are already alive in `by_exchange` and the other indexes, so only a dict reference is added - expected to be negligible, worth confirming rather than assuming.

## Which features need testing

Everything below reaches this file. Grouped by how directly the change touches it.

**Directly affected - unfiltered symbol search (the reported symptom)**

- `/trading` symbol search dialog (`SymbolSearchDialog.tsx` calls `search(q, undefined, 500)`, unfiltered by design, so it always hits the changed path)
- `blueprints/search.py`
- `services/search_service.py`

**Uses the cache, exchange-filtered - should be unchanged, must be confirmed**

- `/scalping` (`blueprints/scalping.py`, `get_distinct_underlyings_cached`)
- Option chain (`services/option_chain_service.py`)
- `services/option_symbol_service.py`
- `services/oi_profile_service.py`
- `services/oi_tracker_service.py`
- `services/vol_surface_service.py`
- `services/iv_chart_service.py`
- `services/straddle_chart_service.py`
- `services/custom_straddle_service.py`
- `services/arbitrage_service.py`
- `utils/symbol_utils.py`

**Symbol resolution and order placement - reads `by_symbol_exchange`, untouched, needs a regression pass anyway**

- `database/token_db.py` (`get_token`, `get_symbol`, `get_br_symbol`, `get_oa_symbol`, `get_brexchange`)
- Order placement and smart orders on at least one Noren-family broker and one non-Noren broker

**Cache lifecycle**

- `database/master_contract_cache_hook.py` (load on master-contract download)
- `database/cache_restoration.py`
- `blueprints/master_contract_status.py` (reported symbol counts will change on affected installs)

**Broker-specific consumers**

- `broker/zerodha/api/gtt_api.py` (`get_symbol_info`)
- `broker/dhan_sandbox/api/data.py` and `broker/dhan_sandbox/streaming/dhan_sandbox_adapter.py` (`fno_search_symbols`)
- `broker/dhan/database/master_contract_db.py` (`extract_underlying_from_symbol`)

**Historify**

- `services/historify_service.py` (`get_symbol_info`, `get_symbol_from_cache`)
- `blueprints/historify.py`

## Acceptance criteria

- Searching an NSE equity whose token collides with a CDS contract returns the NSE row, in a synthetic fixture that reproduces the collision deterministically
- Cache entry count equals master-contract row count after load, asserted in code and covered by a test
- For a non-colliding symbol set, `search_symbols` and `fno_search_symbols` return byte-identical ordered results before and after
- `get_symbol_data` / `by_token` removed, with a test or grep-based check that no caller reintroduces a token-only lookup
- Full test suite green, plus a manual pass over `/trading` search, `/scalping`, option chain and one live order on a Noren-family broker

## Related

- #1774 - the original bug report, with reproduction, the two-line patch and four regression tests attached

subhaushsingh:
Fixes #1774
Fixes #1778

## Problem

`BrokerSymbolCache.load_all_symbols()` keyed the primary `self.symbols`
dict by broker token alone. Tokens are only unique *within* an exchange,
not globally — AliceBlue (and other Noren-family brokers) issue small
integers on both NSE and CDS, so a later-loaded CDS contract silently
overwrote an earlier NSE equity sharing the same token. On the reporting
install this dropped 984 rows (982 NSE, 2 BSE) with nothing logged —
the only symptom was a stock missing from unfiltered symbol search.

## Changes

1. Key the primary store by `(exchange, token)`, matching every other
   index in the class.
2. Assert the load invariant after `load_all_symbols()` — cache count
   vs rows read, logged loudly on mismatch, surfaced via
   `get_cache_info()` / `get_cache_stats()` / `/api/cache/status`.
3. Removed `by_token` and `get_symbol_data()` — same token-only
   ambiguity, no callers anywhere in the repo (verified by grep).

## Testing

`test/test_symbol_cache_token_collision.py` added, covering the
collision recovery, ranking, the load invariant, and the removal
guard. Not yet confirmed passing locally.

## Still open

- [ ] Confirm test suite passes
- [ ] Manual pass: /trading search, /scalping, option chain, one live
      order on a Noren-family broker

---
## Summary by cubic
Fixes #1774 and #1778 by keying the primary symbol cache by `(exchange, token)` instead of `token`, so a token reused across exchanges no longer silently overwrites an earlier symbol. Loads now report when duplicate keys are dropped.

- Loads count rows read vs entries cached, log mismatches at error, and expose them through `get_cache_info()` and `/api/cache/status`.
- Removed `by_token` and `get_symbol_data()`; the scoped indexes remain unchanged.
- Added regression tests, wired into CI, for collisions, ranking, invariant logging, and removal of token-only lookups.

**Migration**
- Replace `get_symbol_data(token)` with `get_symbol(token, exchange)` or `get_symbol_info(symbol, exchange)`.
- Replace `by_token[token]` with `by_token_exchange[(token, exchange)]` and pass `exchange` on all token lookups.
- After rollout, verify `/api/cache/status` reports `load_mismatch=false`.

<sup>Written for commit 9db149b5adbe7b897a2ee547eadc6f9d55a85883. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1798?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


marketcalls:
Reviewed commit `d1ab8fc`. The core fix is valid: using `(exchange, token)` prevents symbols from different exchanges overwriting each other. I recommend addressing the following before merging.

**Merge-readiness assessment: 75/100.** GitHub reports the PR as conflict-free, but there is a reproducible logging defect and some validation/compliance work remains.

1. **P2 — The mismatch diagnostic fails with the application's logging filter.**  
   At [`database/token_db_enhanced.py:288–294`](https://github.com/marketcalls/openalgo/blob/d1ab8fcb9fce98b914d6e16bf589d6366dff89a9/database/token_db_enhanced.py#L288-L294), the new message uses `%d` placeholders. `SensitiveDataFilter` converts every logging argument to a string, so formatting raises `TypeError: %d format: a real number is required, not str`. I reproduced this outside pytest with application logging and colors disabled: the cache loads and `load_mismatch` is true, but the diagnostic is absent from `errors.jsonl`. Please use compatible placeholders or preformat the message, and add a regression test using the actual filter and JSON formatter. Also reconcile this path with the error-logging convention in `CLAUDE.md`; changing the logging method alone will not fix the formatting defect.

2. **P2 — Include the regression tests in CI and strengthen the ranking coverage.**  
   The [backend CI test list](https://github.com/marketcalls/openalgo/blob/d1ab8fcb9fce98b914d6e16bf589d6366dff89a9/.github/workflows/ci.yml#L45-L53) omits `test/test_symbol_cache_token_collision.py`. Please add it. The ranking test currently has only one matching result and loads the equity second, so it passes even before this fix. Use competing matching contracts and load the colliding equity first. Add ordered-result comparisons for unaffected inputs to both `search_symbols` and `fno_search_symbols`, as requested in #1778.

3. **P3 — Complete the `CLAUDE.md` cleanup.**  
   The change introduces five Ruff violations: four whitespace-only lines and a missing final newline. Both changed files fail the formatter check. Please clean up the introduced issues and use a Conventional Commit title, for example `fix: prevent cross-exchange symbol cache collisions`.

Validation performed with `uv` on Windows/Python 3.14 using isolated fixtures:

- All nine submitted tests pass. Against the pre-PR implementation, six fail and three pass, confirming the original collision defect.
- With 14 additional review checks: **22 passed, 1 failed**, with the failure reproducing the missing logging diagnostic.
- Search-order comparisons, recovery among 510 competing option matches, F&O collision handling, and 100 reload/clear cycles pass.
- A temporary in-memory formatting correction makes all 23 focused checks pass; this correction is not part of the submitted PR.
- No production callers of the removed token-only interfaces were found. The resource audit found no new descriptors or accumulation across cache reloads.

Full-suite, Linux/eventlet runtime, and live-broker validation remain outstanding.

Kalaiviswa:
Thank you for investigating and addressing this issue. I reviewed the reported cross-exchange token collision
  and confirmed that it is valid. This PR fixes the root cause by keying the primary symbol cache with (exchange,
  token), preventing symbols on different exchanges from silently overwriting one another.

  I also ran the targeted regression suite from this PR, and all nine tests passed. The existing CI checks are
  green.

  Before merging, could you please address the following remaining items from the acceptance criteria in #1778?


  3. Add an endpoint-level test confirming that /api/cache/status exposes load_row_count, load_cache_count, and
     load_mismatch.

  4. Confirm the manual validation results for:
      - /trading symbol search
      - /scalping
      - Option Chain
      - One symbol lookup or order flow using a Noren-family broker

  5. If practical, provide a small memory comparison demonstrating that retaining the recovered dictionary
     references does not cause a meaningful increase.

  6. Remove the minor trailing whitespace in the modified Python and test files and ensure the test file ends with
     a newline.

  The core implementation looks correct. Completing these validation and coverage items would give us confidence
  that the shared-cache change does not introduce regressions elsewhere before we merge it.

  Thank you again for the contribution and the clear reproduction of the original problem.

subhaushsingh:
Hey @marketcalls @Kalaiviswa  , I’ve addressed the review comments and pushed the updates.

The symbol cache fix is now covered with additional regression tests, including the search ordering cases for both `search_symbols` and `fno_search_symbols`. I also added the cache status endpoint test, fixed the logging issue with the application’s logging filter/formatter, and added the regression test to CI.

I also cleaned up the Ruff and formatting issues and updated the commit message to:

`fix: prevent cross-exchange symbol cache collisions`

The focused test suite is passing with **13/13 tests**.

PR #1801, opened 2026-08-18
fix(sandbox): square off MIS by updated_at session boundary

### LINKED ISSUE #1794: [Bug] Sandbox positions silently zeroed mid-session when same symbol is reopened after a signal flip
[Bug] Sandbox positions silently zeroed mid-session when same symbol is reopened after a signal flip

Environment: OpenAlgo Sandbox (Analyze Mode), Angel One broker, NFO options

What happens: When a strategy signal flips direction (e.g. SHORT → LONG), positions in two different strategy tags get closed and reopened. In cases where the spot price has not moved significantly, the closing of one strategy's position on a symbol (e.g. 24400CE) coincides with another strategy opening a new position on the same symbol within 2–3 seconds.

From OpenAlgo's perspective, that symbol's position row goes: non-zero → 0 → non-zero, all within the same session. OpenAlgo reuses the existing sandbox_positions row rather than inserting a new one. The created_at timestamp on that row remains set to the original date the symbol was first traded (e.g. 2026-08-05).

The trigger: Any background session reload or master contract sync (which calls run_catch_up_tasks() via auth_utils.py) triggers catch_up_mis_squareoff(). This function queries:

python
.filter(SandboxPositions.quantity != 0, SandboxPositions.created_at < today_start)
Since created_at is from a previous day, the position is treated as a stale overnight MIS position and force-closed to quantity = 0 mid-session. The position disappears from the UI with no warning in the user-facing log.
Confirmed via database query:

sqlite3 sandbox.db "SELECT symbol, quantity, created_at, updated_at FROM sandbox_positions WHERE symbol LIKE 'NIFTY11AUG2624%';"
24400CE | 0 | 2026-08-05 10:54:02 | 2026-08-10 04:39:34
24700PE | 0 | 2026-08-05 10:54:12 | 2026-08-10 04:39:34
updated_at correctly shows today's date (the time the position was last reopened)
created_at shows the original date (5 days ago) — this is what the catch-up uses incorrectly
When it does NOT happen: When the signal flip results in different strikes being selected, OpenAlgo inserts new rows for the new symbols with today's created_at. The catch-up processor skips them. Positions show correctly.

Root cause: catch_up_mis_squareoff() in sandbox/catch_up_processor.py uses created_at to determine staleness. It should use updated_at, which correctly reflects the last time the position was modified (including when it was reopened intraday).

Proposed fix — one line change:

python
Current (broken for reopened positions):
.filter(SandboxPositions.quantity != 0, SandboxPositions.created_at < today_start)
Fix:
.filter(SandboxPositions.quantity != 0, SandboxPositions.updated_at < today_start)
Impact: Any strategy that closes and reopens a position on the same symbol within the same session is vulnerable to this. The position silently disappears from the UI and all MTM tracking stops, while the actual exchange position remains open and unmonitored.

santhiprakash:
Closes #1794

## What

`catch_up_mis_squareoff()` treated MIS positions as stale overnight carry when `created_at` was before today’s IST midnight. Reopened symbols reuse the same `sandbox_positions` row and keep the original `created_at`, so a position flipped back to non-zero during the session was force-closed on the next master-contract sync.

## Fix

- Compare `updated_at` (database UTC clock from `func.now()`) against the last session boundary resolved in UTC via `SESSION_EXPIRY_TIME`.
- Add `sandbox/session_boundary.py` with `last_session_expiry_utc()` so catch-up and position-book paths share the same clock semantics.
- Apply `last_session_expiry_utc()` to `catch_up_daily_pnl_reset()` and remove the now-redundant `get_last_session_boundary()`.
- Skip crypto/24x7 brokers and exchanges with no configured square-off time, mirroring `squareoff_manager`.
- Harden `last_session_expiry_utc()` against malformed `SESSION_EXPIRY_TIME` and naive inputs.

## Verification

```bash
uv run pytest test/sandbox/test_catch_up_session_boundary.py -q
```

6 passed.

---
## Summary by cubic
Squares off sandbox MIS positions by `updated_at` against the last UTC session boundary instead of `created_at` at IST midnight, so reopened positions that reuse a row are no longer force-closed. Also skips exchanges without a configured square-off time and disabled session expiry (crypto/24x7), matching `squareoff_manager`.

**Key changes**

- Adds `sandbox/session_boundary.py` to resolve `SESSION_EXPIRY_TIME` (default `"03:00"`) into a naive UTC boundary from the database clock.
- Applies the boundary consistently in `catch_up_mis_squareoff()` and `catch_up_daily_pnl_reset()`.
- Removes the old `get_last_session_boundary()` helper to keep one session-boundary source of truth.
- Adds tests covering boundary math, reopened and overnight positions, and crypto/unconfigured exchange skips.
- Hardens `last_session_expiry_utc()` against malformed time strings and naive inputs.
- No schema or API changes; set `SESSION_EXPIRY_TIME` to match your ops window.

<sup>Written for commit 3bb8c6a6c71831282e9df5171be3aa5ee62ac3df.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1801?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


marketcalls (changes requested):
Thanks for this, and for the clear write-up in the description. The diagnosis of #1794 is correct and the direction is right. I verified the claims rather than taking them on trust. There are three things to address before merge, plus one design question about crypto.

## The fix is correctly aimed

`sandbox_positions.created_at` and `updated_at` are naive columns written by `func.now()`, which on SQLite is `CURRENT_TIMESTAMP`, that is UTC. Row reuse on reopen keeps the original `created_at`, so the old predicate force-closed a position that had been flipped back to non-zero during the session. That is a real bug.

The strongest argument for your approach is one the description undersells: `get_open_positions()` at `position_manager.py:539` already uses exactly this semantic (`position.updated_at >= last_session_expiry`) and already skips previous-session MIS and CNC. Your change makes catch-up agree with the position book rather than contradict it.

I also checked the obvious counter-risk, that `updated_at` is mutable by non-trade events such as MTM. It holds up: `_update_positions_mtm()` receives the already-filtered list, so a previous-session MIS row is never MTM-touched, and `update_all_positions_mtm()` routes through the same filter. Worth knowing that "nothing may touch a previous-session MIS row before catch-up runs" is now a load-bearing invariant, and `position_manager.py:520` already drops to raw SQL specifically to protect it.

## 1. The PR fixes one clock mismatch and leaves an identical one two functions below

`get_last_session_boundary()` at line 25 is untouched and returns a timezone-aware IST datetime. `catch_up_daily_pnl_reset()` still calls it and compares it against `updated_at` in five places. SQLAlchemy's SQLite dialect drops tzinfo when binding, so:

```
get_last_session_boundary() -> 2026-08-26 03:00:00+05:30  (tz-aware IST)
bound into SQL as          -> '2026-08-26 03:00:00.000000'
correct naive-UTC boundary -> 2026-08-25 21:30:00
SKEW                       -> +5.5 hours
```

The boundary lands 5.5 hours late, so `updated_at < boundary` over-matches and `today_realized_pnl` can be zeroed for rows updated between 03:00 and 08:30 IST, inside the current session.

This is pre-existing, not introduced by you. But this PR edits those exact lines (`!= None` to `is_not(None)`) while leaving the wrong clock beside them, and your own `session_boundary.py` docstring states the rule the code breaks: "Comparisons against position timestamps must resolve the boundary in the database clock." Two session-boundary helpers with incompatible return types in one file is how the next person picks the wrong one. Please point `catch_up_daily_pnl_reset()` at `last_session_expiry_utc()` and delete `get_last_session_boundary()`.

`catch_up_t1_settlement()` has the same defect (`datetime.combine(today, ...)` is naive IST midnight versus naive UTC `created_at`). Different product, reasonable to defer, but it belongs in the same sweep.

## 2. Two of the four tests do not test the code

```python
is_stale = quantity != 0 and updated_at < boundary
assert not is_stale
```

That re-implements the predicate inline rather than calling `catch_up_mis_squareoff()`. No test imports the changed function.

I checked this rather than assuming it. I restored main's buggy `created_at` version of `catch_up_processor.py` and re-ran your suite:

```
test/test_catch_up_session_boundary.py ....    [100%]
4 passed in 0.05s
```

All four pass against the exact bug this PR exists to fix. The two `last_session_expiry_utc` tests are genuinely useful and should stay. The other two need to become one test that builds a row with an old `created_at` and a fresh `updated_at`, runs `catch_up_mis_squareoff()`, and asserts the position survives.

## 3. Please consider crypto exchanges

This is the design question, and it matters more than it first appears.

`utils/session.py` has `is_session_expiry_disabled()`, driven by `DISABLE_SESSION_EXPIRY`, which the install script sets automatically when a crypto broker such as deltaexchange is selected. `get_session_expiry_time()` then returns `timedelta(days=365)`, meaning the instance has no daily session boundary at all. Delta Exchange trades on exchange code `CRYPTO`, and `is_market_open()` short-circuits it to always open, 24/7.

Now compare the two square-off paths:

- `squareoff_manager.square_off_times` has entries for NSE, BSE, NFO, BFO, CDS, BCD, MCX and NCDEX. There is no `CRYPTO` entry, so `check_and_square_off()` hits `if not square_off_time: continue` and never closes a crypto MIS position.
- `catch_up_mis_squareoff()` has no exchange filter and does not consult `is_session_expiry_disabled()`, so it will force-close a `CRYPTO` MIS position at the 03:00 boundary.

That is backwards. Catch-up exists to perform square-offs that were missed while the app was down, but on a crypto instance there is no scheduled square-off to miss, so catch-up would be inventing one. A trader who opens a crypto MIS position at 22:00 IST and holds it through 03:00, in a market that never closed, would find it force-closed on the next master-contract sync.

The old `created_at` predicate had the same gap, so this is not a regression you introduced. I raise it here because your PR rewrites the staleness predicate and adds a helper whose entire premise is that a daily boundary exists, which is precisely what `DISABLE_SESSION_EXPIRY` declares false. The natural fixes are to return early when `is_session_expiry_disabled()` is true, and to skip exchanges with no configured square-off time, mirroring what `squareoff_manager` already does. Note this function already handles crypto in one respect, the `contract_value` multiplier for ETHUSD.P, so crypto positions are demonstrably expected to reach it.

## Security validation

No vulnerabilities. Checked systematically rather than by pattern matching:

- **Injection**: none. Every filter is an ORM expression. No raw SQL, string interpolation, `.raw()` or `.extra()` is introduced. `session_boundary.py` touches no database at all.
- **Attacker-controlled input**: none. The only external value is `SESSION_EXPIRY_TIME`, and I confirmed it is read solely via `os.getenv` and is not settable through any route, `settings_db`, or the frontend. It is operator deployment configuration, not user input.
- **Dangerous sinks**: no `eval`, `exec`, `pickle`, `yaml.load`, `subprocess`, `os.system` or `shell=True` anywhere in the diff.
- **Secrets**: none introduced. No credentials, tokens or keys.
- **Auth and authz**: no routes, no auth code, no privilege boundary crossed. Sandbox is single-user per deployment.
- **Data exposure**: log lines carry symbol, user_id and pnl, consistent with surrounding code and covered by the `SensitiveDataFilter` in `utils/logging.py`.
- **Supply chain**: no dependency changes, confirmed by both Snyk checks.
- **Resource handling**: no new engines, sessions, threads, sockets or subprocesses. `session_boundary.py` is pure computation, so nothing here trips the FD audit.
- **Blast radius**: `sandbox.db` is isolated from live trading, so any defect affects the simulated ledger only.

One robustness note that is explicitly **not** a security finding, since the input is server-controlled: `map(int, session_expiry_str.split(":"))` raises on a malformed value such as `"3"`, `"03:00:00"` or `"25:00"`, and the caller wraps everything in a broad `except Exception` that logs and returns. A config typo would therefore silently skip the entire MIS square-off. `database/token_db_enhanced.py:324` already handles this correctly by validating and falling back to `03:00` with a warning, and it is worth matching. Related: `last_session_expiry_utc` calls `.astimezone(UTC)` without checking that `now_local` is aware, so a naive argument would be silently read as system-local time and produce a wrong boundary on a non-IST host. The docstring says aware; nothing enforces it.

## Where this lands

Mechanically it is ready: it merges clean and CI is green. As a change to merge as-is I would put it at roughly 65 out of 100. The core idea is right and the scope is disciplined, but it stops one function short of finishing the clock cleanup it starts, its tests provably do not guard the fix, and crypto has not been considered.

Address items 1 and 2, tell me how you want to handle crypto in item 3, and this is a straightforward merge.

santhiprakash:
@marketcalls thanks for the thorough review. I've pushed a new commit that addresses the three blockers:

1. **Clock cleanup in `catch_up_daily_pnl_reset`**
   - Removed `get_last_session_boundary()` entirely.
   - `catch_up_daily_pnl_reset()` now uses `last_session_expiry_utc()` so the boundary is resolved in the database (UTC) clock before comparing against `updated_at`.

2. **Tests now exercise the changed function**
   - Kept the two `last_session_expiry_utc` boundary tests.
   - Replaced the two inline-predicate tests with `catch_up_mis_squareoff()` calls:
     - `test_reopened_mis_position_survives_catch_up` builds a row with an old `created_at` and a fresh `updated_at` and asserts the position survives.
     - `test_overnight_mis_position_is_settled_on_catch_up` builds a row with both timestamps before the boundary and asserts it is settled.
   - Both fail against the old `created_at` predicate and pass with the `updated_at` fix.

3. **Crypto / 24x7 handling**
   - `catch_up_mis_squareoff()` returns early when `is_session_expiry_disabled()` is true.
   - It also skips any exchange that has no configured square-off time, mirroring `SquareOffManager`.

**Additional hardening**
- `last_session_expiry_utc()` now falls back to `03:00` on a malformed `SESSION_EXPIRY_TIME` string and treats a naive input as IST, matching the existing `token_db_enhanced.py` pattern.

**Deferred**
- `catch_up_t1_settlement()` has the same naive-midnight vs UTC `created_at` issue, but as you noted it's a different product and a reasonable follow-up. I left it untouched to keep this PR focused.

All CI checks are green on the updated head (`3bb8c6a`).

marketcalls (approved):
All three blockers are genuinely fixed. Verified independently rather than from the summary.

**Clock cleanup.** `get_last_session_boundary()` is gone and both `catch_up_mis_squareoff` and `catch_up_daily_pnl_reset` now resolve through `last_session_expiry_utc`. I re-measured the skew I originally reported: **+0.0h, down from +5.5h**, and the helper returns naive UTC matching the column type. That was the part I cared about most, because the old boundary landed 5.5 hours late and could zero `today_realized_pnl` for rows updated inside the current session.

**Tests exercise the code.** Four of the six now call `catch_up_mis_squareoff()` directly instead of re-implementing the predicate. I reverted the filter to `created_at` and `test_reopened_mis_position_survives_catch_up` fails, so it catches the exact #1794 bug rather than passing regardless. That was the specific problem with the previous version and it is properly resolved.

**Crypto.** `is_session_expiry_disabled()` early return plus skipping exchanges with no configured square-off time, mirroring `SquareOffManager`, with a test for each. This one mattered: without it the catch-up would have force-closed crypto MIS positions that the scheduled square-off never touches, inventing a square-off on an instance that has no session boundary at all.

The `SESSION_EXPIRY_TIME` fallback and the naive-datetime handling were not asked for and are good additions.

129 sandbox tests pass, 15 CI checks green, merges clean.

## One gap left in that new fallback

The `try` wraps only the `int` parse, so an out-of-range hour still escapes on the next line:

```
'abc'    -> falls back to 03:00 with a warning
'25:00'  -> ValueError: hour must be in 0..23, not 25
```

`catch_up_mis_squareoff`'s broad `except Exception` swallows that, so a config typo silently skips the MIS square-off entirely. It is operator-controlled and sandbox-only, so low severity, but it is the same fail-open the fallback exists to close. I am fixing it in a follow-up commit rather than sending you round again for one line.

Deferring `catch_up_t1_settlement()` was the right call, and saying so explicitly rather than leaving it silent is what made it easy to accept.

marketcalls:
Follow-up pushed as f0b39cf1, so you do not need to do anything.

The `SESSION_EXPIRY_TIME` fallback you added wrapped only the int parse, so a value that parses cleanly but is out of range escaped it:

```
'abc'   -> falls back to 03:00 with a warning
'25:00' -> ValueError: hour must be in 0..23, not 25
```

The raise came from `replace()` on the next line, outside the `try`. Since `catch_up_mis_squareoff()` wraps its whole body in `except Exception`, that was swallowed and the square-off silently did not run, which is the fail-open the fallback exists to prevent.

Parse and range-check now happen together in one `try`. `'25:00'`, `'-1:00'` and `'03:99'` fall back with the same warning, and a valid non-default such as `'09:15'` still resolves normally rather than being masked.

Added seven parametrised cases plus one asserting the fallback does not swallow a good value. Verified they are not decorative: restoring the split form makes `'25:00'` raise again and fails exactly three of them, the three that are range errors rather than parse errors. 137 sandbox tests pass.

Thanks again for the three fixes and the unprompted hardening. This was a small gap in a good addition, not a defect in the original review scope.

Issue #1779, opened 2026-08-09
Historify: large 1m backfills OOM - three unused indexes on market_data plus an unconfigured DuckDB connection


marketcalls:
Large Historify backfills die with DuckDB out-of-memory. The cause is three indexes on `market_data` that no query uses, plus an unconfigured DuckDB connection. No code has been changed - this issue records the diagnosis and the plan.

## Symptom

Downloading 10 years of 1m data for 200 symbols:

```
[08:01:16] INFO  Downloading NESTLEIND:NSE:1m from 2016-07-09 to 2026-08-09
[08:02:15] ERROR TransactionContext Error: Failed to commit:
                 failed to pin block of size 256.0 KiB (6.3 GiB/6.3 GiB used)
```

Every symbol after the first fails the same way, and the failing allocation shrinks from 256 KiB to 64 KiB. Each symbol is logged and skipped, so the run completes with silent gaps rather than stopping.

Environment: 8 GB RAM (DuckDB default `memory_limit` = 80% = 6.4 GB), `historify.duckdb` at 6.76 GB, DuckDB 1.5.4.

## Root cause

**DuckDB ART indexes must be fully resident in memory and cannot spill to disk.** `market_data` carries four:

```sql
PRIMARY KEY (symbol, exchange, interval, timestamp)          -- database/historify_db.py:116
CREATE INDEX idx_market_data_timestamp     (timestamp)       -- :245
CREATE INDEX idx_market_data_exchange_time (exchange, timestamp)  -- :249
CREATE INDEX idx_market_data_interval_time (interval, timestamp)  -- :253
```

200 symbols x 10 years x 1m is roughly 187M rows (375 bars/day x 250 days x 10 years x 200). Data pages can be evicted; index memory cannot. The allocation failures degrading to 64 KiB are the signature - the buffer manager freed everything it was allowed to and still could not find 64 KiB.

Memory also does not come back between symbols. `get_connection()` closes the connection per call, but DuckDB's Python instance cache keeps one `DatabaseInstance` per file path per process, so `conn.close()` closes the connection, not the instance. Index memory persists for the life of the worker; restarting the app is currently the only way to reclaim it.

Two contributing factors:

- `get_connection()` (`database/historify_db.py:75`) is `duckdb.connect(db_path)` with **no configuration at all** - no `memory_limit`, no `threads`, no `preserve_insertion_order`, no explicit `temp_directory`. The limit is whatever 80% of the host happens to be.
[...]

Issue #1726, opened 2026-08-02
Install not complited


hexaautomations:
sqlalchemy.exc.OperationalError:
sqlite3.OperationalError: database is locked
thanks with regards
sridhar

marketcalls:
Hi Sridhar, sorry you hit this. "database is locked" during install almost always means **something else is holding the SQLite file**, so let us find what.

To help properly I need a few details, because the report does not yet say:

1. **Operating system** — Windows, Ubuntu, macOS?
2. **Install method** — Windows installer, `install.sh`, Docker, or `uv run app.py`?
3. **Where is the OpenAlgo folder?** Full path please.
4. **The full traceback**, not just the last two lines — the lines above it name the migration that failed.

## Meanwhile, the four usual causes, most likely first

**1. An OpenAlgo instance is already running.**
This is by far the most common. Migrations cannot get a write lock while the app holds it.

```bash
# Linux
sudo systemctl stop openalgo
ps -ef | grep -E "gunicorn|app.py"      # must be empty

# Docker
docker compose down

# Windows (PowerShell)
Get-Process python* | Stop-Process -Force
```

Then re-run the install or update. Also check you do not have a second terminal running `uv run app.py`, or DB Browser for SQLite left open on `db/openalgo.db`.

**2. The `db/` folder is on OneDrive, Google Drive, Dropbox, or a network share.**
This is the one people do not expect. SQLite's WAL mode **requires a local filesystem** — it uses shared memory that sync clients and network mounts do not support. On those paths locking silently misbehaves and you get exactly this error.

```
Bad:  C:\Users\You\OneDrive\Documents\openalgo
Bad:  \\server\share\openalgo   or  Z:\openalgo
Bad:  /mnt/c/... accessed from WSL while Windows also uses it
Good: C:\openalgo   or   /opt/openalgo   or   ~/openalgo (not synced)
```

If your folder is in a synced location, move it to a plain local path and re-run. This alone fixes most Windows reports.

**3. Antivirus or Windows Search is scanning the `db/` folder.**
Add an exclusion for the OpenAlgo `db/` directory and retry.

**4. Permissions on the `db/` folder.**
SQLite must create `openalgo.db-wal` and `openalgo.db-shm` beside the database. If the folder is read-only, or owned by another user, it fails.

```bash
ls -la db/          # Linux/macOS - you should see -wal and -shm files
sudo chown -R $USER db/
```

## Quick check you can run

[...]

PR #1803, opened 2026-08-18
fix(historify): drop unused market_data secondary indexes

santhiprakash:
Refs #1779

## What

Historify's `market_data` table carries three secondary indexes (`idx_market_data_timestamp`, `idx_market_data_exchange_time`, `idx_market_data_interval_time`) that no query uses: every lookup leads with `symbol`, DuckDB serves range scans from per-row-group zone maps, and ART index memory stays fully resident as the table grows — which is what OOMs large 1m backfills (#1779). Per the measurements in the issue, dropping them speeds up ingest 1.6x and shrinks the file 2.4x with no query cost.

## Fix (Tier 1 from the issue: automatic migration, safe for every install)

- **`upgrade/migrate_historify_drop_indexes.py`** (new, migration `012`, registered in `migrate_all.py`): drops the three indexes on existing installs. `DROP INDEX` is catalog-only (never reads or rewrites table data), followed by an explicit `CHECKPOINT` so the file actually shrinks. Idempotent, supports `--status`, returns quietly when no Historify DB exists (most installs never opened Historify), and a locked DB (running server) exits with "Stop OpenAlgo first" instead of a raw `IOException`.
- **Index creation removed from both fresh-install paths** — `database/historify_db.py::init_database()` and `upgrade/migrate_historify.py::create_database()` (the statements were duplicated across the two files) — so new installs never create them and a fresh DB is schema-identical to a migrated one.

The primary key on `(symbol, exchange, interval, timestamp)` is untouched; non-`market_data` indexes (`idx_job_items_job_id`, ...) are untouched.

Tier 1b from the issue (DuckDB connection tuning in `get_connection()` — `memory_limit`/`threads`/`preserve_insertion_order`/`temp_directory` — and chunked `download_data()`) is deliberately not in this PR, per the issue's own risk-tier separation of the two halves.

## Verification

Acceptance suite against a synthetic 720k-row old-schema `historify.duckdb` (repo-pinned `duckdb==1.5.4`, same version as the report) — **all checks passed**:

| Check | Result |
| --- | --- |
| `--status` before / after | "Migration needed" (exit 1) / "already applied" (exit 0) |
| Apply drops exactly the 3 indexes | PASS — `duckdb_indexes()` shows none left |
| File shrinks after `CHECKPOINT` | 35.4 MB → 10.5 MB |
| Idempotent re-run | "already absent", exit 0 |
| Row content identical | md5 equal before/after, 500k rows preserved |
| Primary key still enforced | duplicate insert rejected |
| Non-`market_data` indexes untouched | `idx_job_items_job_id` still present |
| Absent DB file | quiet no-op, exit 0 |
| Locked DB (held by another process) | exit 1 + "Stop OpenAlgo first" |
| Fresh install (patched `migrate_historify.py` and standalone `init_database()`) | none of the 3 indexes created; fresh schema identical to migrated |
| `ruff check` on the new file | clean |

---
## Summary by cubic
Remove three unused secondary indexes on Historify’s `market_data` and stop creating them on fresh installs to prevent OOMs during large 1m backfills and improve ingest speed and file size. Previously we created `idx_market_data_timestamp`, `idx_market_data_exchange_time`, and `idx_market_data_interval_time`; now a migration drops them and issues a CHECKPOINT. Query behavior is unchanged (lookups lead with `symbol`), and the primary key remains enforced.

**Migration and review notes**
- Adds migration 012 (`upgrade/migrate_historify_drop_indexes.py`, registered in `upgrade/migrate_all.py`) to drop the indexes and CHECKPOINT; idempotent, quiet no-op if the DB is absent, and surfaces a clear “Stop OpenAlgo first” message if the DB is locked.
- Removes index creation from `database/historify_db.py::init_database()` and `upgrade/migrate_historify.py::create_database()` so fresh and migrated schemas match.
- Scope limited to `market_data` secondary indexes; the `(symbol, exchange, interval, timestamp)` primary key and non-`market_data` indexes (e.g., `idx_job_items_job_id`) are unchanged.
- Rollout: stop OpenAlgo, run the standard upgrade; `--status` is available to check application. Addresses #1779 (Tier 1).

<sup>Written for commit 82833e9c08553ce1a1d98d8b0c2335cb244af2d6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1803?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/database/historify_db.py
+++ b/database/historify_db.py
@@ -240,19 +240,10 @@ def init_database():
             )
         """)
 
-        # Create indexes for common query patterns
-        conn.execute("""
-            CREATE INDEX IF NOT EXISTS idx_market_data_timestamp
-            ON market_data (timestamp)
-        """)
-        conn.execute("""
-            CREATE INDEX IF NOT EXISTS idx_market_data_exchange_time
-            ON market_data (exchange, timestamp)
-        """)
-        conn.execute("""
-            CREATE INDEX IF NOT EXISTS idx_market_data_interval_time
-            ON market_data (interval, timestamp)
-        """)
+        # No secondary indexes on market_data: every query leads with
+        # `symbol`, DuckDB serves range scans from per-row-group zone maps,
+        # and ART index memory stays fully resident as the table grows,
+        # which OOMs large 1m backfills. See #1779.
         conn.execute("""
             CREATE INDEX IF NOT EXISTS idx_job_items_job_id
             ON job_items (job_id)
--- a/upgrade/migrate_all.py
+++ b/upgrade/migrate_all.py
@@ -71,6 +71,7 @@ MIGRATIONS = [
     ("migrate_zerodha_new_exchanges.py", "Zerodha NCO/GLOBAL_INDEX & GIFTNIFTY Cleanup"),
     ("add_totp_purpose_flags.py", "Per-Purpose 2FA Flags (login/MCP/reset)"),
     ("migrate_gtt_sandbox.py", "Sandbox GTT Support & CAS F&O Close (15:40)"),
+    ("migrate_historify_drop_indexes.py", "Historify Unused Index Removal (#1779)"),
 ]
 
 
--- a/upgrade/migrate_historify.py
+++ b/upgrade/migrate_historify.py
@@ -187,20 +187,13 @@ def create_database():
         """)
         logger.info("Created symbol_metadata table")
 
-        # Create indexes for common query patterns
+        # No secondary indexes on market_data: every query leads with
+        # `symbol`, DuckDB serves range scans from per-row-group zone maps,
+        # and ART index memory stays fully resident as the table grows,
+        # which OOMs large 1m backfills. Existing installs get the three
+        # legacy indexes dropped by migrate_historify_drop_indexes.py.
+        # See #1779.
         logger.info("Creating indexes...")
-        conn.execute("""
-            CREATE INDEX IF NOT EXISTS idx_market_data_timestamp
-            ON market_data (timestamp)
-        """)
-        conn.execute("""
-            CREATE INDEX IF NOT EXISTS idx_market_data_exchange_time
-            ON market_data (exchange, timestamp)
-        """)
-        conn.execute("""
-            CREATE INDEX IF NOT EXISTS idx_market_data_interval_time
-            ON market_data (interval, timestamp)
-        """)
         conn.execute("""
             CREATE INDEX IF NOT EXISTS idx_job_items_job_id
             ON job_items (job_id)
--- a/upgrade/migrate_historify_drop_indexes.py
+++ b/upgrade/migrate_historify_drop_indexes.py
@@ -1,0 +1,227 @@
+#!/usr/bin/env python
+"""
+Historify Secondary Index Removal Migration for OpenAlgo
+
+Drops the three unused secondary indexes on market_data that keep DuckDB
+ART index memory fully resident as the table grows, which causes
+out-of-memory failures on large 1m backfills (#1779):
+
+- idx_market_data_timestamp
+- idx_market_data_exchange_time
+- idx_market_data_interval_time
+
+Every query against market_data leads with `symbol`; none of these
+indexes contains that column, the planner never selects them, and range
+scans are served by DuckDB's per-row-group zone maps. Measured in #1779,
+removing them roughly halves the database file, speeds up ingest about
+1.6x, and costs no query performance. The primary key on
+(symbol, exchange, interval, timestamp) is untouched and remains
+enforced after the drop.
+
+DROP INDEX is a catalog-only operation - it never reads or rewrites
+table data. An explicit CHECKPOINT follows the drop so the file size
+actually shrinks.
+
+Usage:
+    cd upgrade
+    uv run migrate_historify_drop_indexes.py           # Apply migration
+    uv run migrate_historify_drop_indexes.py --status  # Check status
+
+Migration: 012
+Created: 2026-08-18
+Issue: #1779
+"""
+
+import argparse
+import os
+import sys
+from datetime import datetime
+
+# Add parent directory to path for imports
+sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+
+from dotenv import load_dotenv
+
+from utils.logging import get_logger
+
+logger = get_logger(__name__)
+
+# Migration metadata
+MIGRATION_NAME = "historify_drop_unused_indexes"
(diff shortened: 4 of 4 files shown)


marketcalls (changes requested):
This is a well-executed migration. I did not take the verification table on trust: I built a synthetic 200k-row old-schema DuckDB with all three legacy indexes and ran the script against it on the pinned `duckdb==1.5.4`. Everything you claimed reproduces.

```
--status (needed)     "Migration needed"                      exit 1
apply                 dropped exactly the 3 indexes           exit 0
--status (applied)    "already applied"                       exit 0
re-run                "already absent - nothing to do"        exit 0
absent DB             quiet no-op (apply and --status)        exit 0
locked DB             clear "Stop OpenAlgo first", no traceback   exit 1

rows       200000 -> 200000
md5        cde6839956b9d54d -> cde6839956b9d54d   (identical)
file       39.51 MB -> 8.76 MB                    (4.5x smaller)
PK         duplicate insert still rejected
job_items  idx_job_items_job_id still present
market_data indexes after: []
```

I also checked the claims about the codebase rather than the fixture:

- **Only two index-creation sites exist**, and this PR patches both. There is no third path that would silently recreate them.
- **All 18 `WHERE` clauses against `market_data` lead with `symbol`.** That is the load-bearing claim behind "unused", and it holds.
- **Migration `012` is free and sequential** (existing: `004-gtt`, `005`, `010`, `011`), and it is registered in `MIGRATIONS`.
- **`upgrade/_pragmas.py` is SQLite/SQLAlchemy-only**, so a DuckDB migration correctly does not import it. Worth stating because #1726 made that the convention for the SQLite migrations.
- Merges clean onto current `main`, CI green.

The scoping is right too. Keeping Tier 1b out of this PR matches the issue's own risk separation, and the comments left behind at both former index sites explain *why* there are no secondary indexes, which is what stops someone helpfully re-adding them later.

## One thing to change before merge

**The PR says `Fixes #1779`, which will auto-close the issue on merge.** That issue is the plan for all three tiers, and two of them are still outstanding: Tier 1b (`get_connection()` configuration, chunked `download_data()`, the actionable error message) and Tier 2 (the opt-in primary-key removal script). Closing it would lose the tracking for both, along with a lot of measurement work that is not written down anywhere else.

Please change it to `Refs #1779` or `Part of #1779`.

Related, and worth being explicit about in the PR body: this does not by itself satisfy the issue's last acceptance criterion, "a 10-year 1m backfill for a meaningful symbol count completes without OOM". Index memory was the dominant term, but the issue also names the unconfigured `get_connection()` and the single-transaction `download_data()` as contributing factors. A user on the 8 GB machine from the report may still hit the ceiling after this lands. Saying so up front avoids a "the fix did not work" follow-up.

## Two notes, neither blocking

**1. `traceback.print_exc()` in `upgrade()`.** `CLAUDE.md` is explicit that error logging is always `logger.exception()` and that `import traceback` / `traceback.print_exc()` must never be used, because they bypass centralized logging and never reach `log/errors.jsonl`. Same for the `logger.error(f"Migration failed: {e}")` calls, which drop the traceback.

I want to be fair about this rather than quote the rule at you: **10 of the 28 scripts in `upgrade/` already do exactly this**, and `logger.error` outnumbers `logger.exception` 75 to 2 in that directory. So this PR is following local precedent, not introducing a novel violation. It is a policy-versus-precedent call for the maintainer, not a defect. If the answer is "match the documented rule", the change is two lines here; if it is "match the directory", leave it and the whole directory wants a separate sweep.

**2. `status()` closes its connection per branch, `drop_indexes()` uses `try/finally`.** I traced every path in `status()` and none of them leaks, so this is not a bug. But the `finally` form in `drop_indexes()` is the more robust of the two and matches the FD-hygiene emphasis in `CLAUDE.md`, so it would be nice if both used it. Purely cosmetic.

## Verdict

Approve once `Fixes #1779` becomes `Refs #1779`. The migration itself is correct, idempotent, honest about the locked-database case, and measurably does what it says.

santhiprakash:
@marketcalls thanks for the thorough review and for independently reproducing the migration.

- Updated the PR body to say **Refs #1779** so the issue stays open for Tier 1b/2.
- I left the two non-blocking notes as-is for now: the logging style matches the rest of `upgrade/` (as you noted, a directory-wide sweep is probably the right place for that), and the connection handling is equivalent on every path. Happy to fold them into a follow-up cleanup PR if you'd prefer.

Let me know if anything else is needed.

marketcalls (approved):
Merging. `Refs #1779` is the change I needed, and it matters: that issue is the plan for all three tiers, and closing it would have discarded the tracking for Tier 1b (connection config, chunked `download_data()`, the actionable error) and Tier 2 (the opt-in primary-key rebuild), along with a lot of measurement work not written down anywhere else. #1779 stays open.

Your reasoning on the two non-blocking notes holds and I am not asking for either.

The `traceback.print_exc()` style matches 10 of the 28 scripts in `upgrade/`, where `logger.error` outnumbers `logger.exception` 75 to 2. Fixing it here alone would make this file the odd one out; a directory-wide sweep is the right venue, as you say. On the connection handling, I traced every path in `status()` and none of them leaks, so the difference from `drop_indexes()` is cosmetic.

Nothing has changed in the migration since I ran it end to end against a synthetic 200k-row old-schema database on the pinned `duckdb==1.5.4`:

```
--status (needed)   "Migration needed"                exit 1
apply               dropped exactly the 3 indexes     exit 0
--status (applied)  "already applied"                 exit 0
re-run              "already absent"                  exit 0
absent DB           quiet no-op                       exit 0
locked DB           "Stop OpenAlgo first"             exit 1, no traceback

rows  200000 -> 200000
md5   cde6839956b9d54d -> cde6839956b9d54d   identical
file  39.51 MB -> 8.76 MB
PK    duplicate insert still rejected
```

It still merges clean after a heavy day of churn on `main`, and 9 checks are green.

Worth restating for whoever picks up the next tier: this closes the dominant term but not the whole issue. Index memory was the main cause of the OOM, and the unconfigured `get_connection()` and single-transaction `download_data()` are still named contributors, so the 8 GB machine in the report may still hit the ceiling. That is exactly why the issue stays open.

Good migration. Idempotent, honest about the locked-database case, `--status` support, and a `CHECKPOINT` so the file actually shrinks rather than looking like it did nothing.

PR #1811, opened 2026-08-21
fix(sandbox): serialize concurrent _update_position RMW with a threading.Lock (#1808)

### LINKED ISSUE #1808: Sandbox mode: concurrent same-symbol orders silently lose position updates (orderbook complete, positionbook wrong)
## Summary

In sandbox/analyzer mode, two `placeorder` requests for the **same symbol** arriving concurrently cause **one of the two position updates to be silently lost**. The orders both return `status: success` and both appear in `/api/v1/orderbook` as `complete`, but `/api/v1/positionbook` reflects only one of them.

Measured failure rate: **7 of 8 concurrent pairs lost an update.** The same pairs placed back-to-back (no delay, just serialized) were correct 8 of 8.

This is a data-integrity issue rather than a cosmetic one: the position book is what an operator looks at to decide whether they are flat.

## Environment

| | |
|---|---|
| Commit | `16bf8e7f80770065fc31e0d9b188805bb7bdc39b` (`main`, clean tree) |
| `git describe` | `openalgo-investright-aliceblue-greeks-9-g16bf8e7f8` |
| OS | Windows 11 |
| Python | 3.14 (uv-managed) |
| DB | SQLite |
| Client | `openalgo` PyPI package 1.0.47 |
| Mode | Analyzer (`analyze_mode: true`) |
| Exchange / product | BFO / NRML (index options) |

## Real-world impact

An automated strategy fired a 6-order end-of-day flatten. Four distinct symbols were involved; **two of them received two orders each in the same second**. All six orders returned success, all six polled to `complete` via `/api/v1/orderstatus`, and the orderbook netted to exactly zero on every symbol.

The position book still showed both doubled symbols as **short 60**. The operator, correctly trusting the position book, manually closed both — which was the right action given what was displayed, but the underlying orders had already been placed. The two symbols that received only **one** order each were correctly flat.

## Root cause

`SandboxExecutionEngine._update_position()` in `sandbox/execution_engine.py` is a read-modify-write with no locking and no isolation:

```python
# sandbox/execution_engine.py:692
position = SandboxPositions.query.filter_by(
    user_id=order.user_id,
    symbol=order.symbol,
    exchange=order.exchange,
    product=order.product,
).first()                                     # READ

if not position:
    position = SandboxPositions(...)          # CREATE
    db_session.add(position)
else:
    old_quantity = position.quantity          # READ
    new_quantity = effective_qty if order.action == "BUY" else -effective_qty
    final_quantity = old_quantity + new_quantity   # MODIFY, then write
```

`grep` for `Lock`/`threading`/`acquire` in `sandbox/execution_engine.py` returns nothing, so two request threads can interleave freely between the read and the write.

There are **two distinct failure paths**, and they differ in an important way:

**Path 1 — no existing position (loud).** Both threads take the `if not position` branch and both `INSERT`. The `unique_position` constraint on `(user_id, symbol, exchange, product)` (`database/sandbox_db.py:163`) rejects the second one:

```
(sqlite3.IntegrityError) UNIQUE constraint failed:
sandbox_positions.user_id, sandbox_positions.symbol,
sandbox_positions.exchange, sandbox_positions.product
```

**Path 2 — position already exists (silent).** Both threads read the same `old_quantity`, each computes `old + its own delta`, and both write. Last write wins; the other is lost. **No exception, nothing logged.** This is the path the production incident above took, which is why nothing surfaced at the time.

### The ordering that makes it user-visible

The order is committed as `complete` **before** the position update is attempted:

```python
# sandbox/execution_engine.py
586:    db_session.commit()            # order is now 'complete' and durable
589:    self._update_position(order, execution_price)
```

and `_update_position` ends with:

```python
# sandbox/execution_engine.py:965
except Exception as e:
    db_session.rollback()
    logger.exception(f"Error updating position for order {order.orderid}: {e}")
    raise
```

So the rollback discards only the position update. The order row stays `complete`. The result is an orderbook and a position book that disagree, with no indication to the API caller — `placeorder` had already returned `success` by then.

## Reproduction

Analyzer mode, any two same-symbol orders fired concurrently. Standalone script:

```python
import os, threading, time
from openalgo import api

client = api(api_key=os.environ["OPENALGO_API_KEY"], host="http://127.0.0.1:5000")
SYM, EXCH, PROD, QTY = "SENSEX27AUG2677500CE", "BFO", "NRML", 60

def net_qty(sym):
    for row in client.positionbook().get("data") or []:
        if row.get("symbol") == sym and row.get("product") == PROD:
            return int(float(row.get("quantity", 0)))
    return 0

def place():
    barrier.wait()
    client.placeorder(strategy="race", symbol=SYM, exchange=EXCH, action="SELL",
                      price_type="MARKET", product=PROD, quantity=str(QTY))

print("before:", net_qty(SYM))
barrier = threading.Barrier(2)
ts = [threading.Thread(target=place) for _ in range(2)]
[t.start() for t in ts]
[t.join() for t in ts]
time.sleep(3)
print("after: ", net_qty(SYM), "  expected -120")
```

### Results

Each trial used a fresh symbol, placed `2 x SELL 60`, and read back the net quantity. Correct result is `-120` in every case.

| Arm | Method | Correct | Lost update |
|---|---|---|---|
| A | Concurrent (released from a `threading.Barrier`) | **1 / 8** | **7** |
| B | Back-to-back, second placed as soon as the first returns | 8 / 8 | 0 |
| C | Spaced 2 s apart | 4 / 4 | 0 |

Arm A output:

```
SENSEX27AUG2676800CE ->  -60  LOST UPDATE
SENSEX27AUG2676800PE ->  -60  LOST UPDATE
SENSEX27AUG2676900CE -> -120  OK
SENSEX27AUG2676900PE ->  -60  LOST UPDATE
SENSEX27AUG2677000CE ->  -60  LOST UPDATE
SENSEX27AUG2677000PE ->  -60  LOST UPDATE
SENSEX27AUG2677100CE ->  -60  LOST UPDATE
SENSEX27AUG2677100PE ->  -60  LOST UPDATE
```

Arm B is the useful diagnostic: **merely serializing the calls — with no added delay — was sufficient.** The window is the request handler itself, not an asynchronous write settling afterwards.

## Log evidence

From `log/errors.jsonl` (Path 1 trials only; Path 2 leaves no trace):

```json
{
  "ts": "2026-08-20 18:34:31",
  "level": "ERROR",
  "logger": "sandbox.execution_engine",
  "file": "sandbox/execution_engine.py:967",
  "message": "Error updating position for order 26082058391127: (sqlite3.IntegrityError) UNIQUE constraint failed: sandbox_positions.user_id, sandbox_positions.symbol, sandbox_positions.exchange, sandbox_positions.product [SQL: INSERT INTO sandbox_positions ...]",
  "request": {"method": "POST", "path": "/api/v1/placeorder", "ip": "127.0.0.1"}
}
```

9 such entries were logged across the trial runs, matching the 9 lost updates observed on the create path.

## Suggested fixes

1. **Serialize the position update per `(user_id, symbol, exchange, product)`.** Given Arm B passed 8/8, a mutex held across the read-modify-write in `_update_position` is sufficient. For multi-process deployments an in-process lock is not enough and this needs to be pushed to the database.

2. **Make the update atomic in SQL** rather than read-then-write — an `UPDATE ... SET quantity = quantity + :delta` for the existing-position case, and an upsert (`INSERT ... ON CONFLICT DO UPDATE`) for the create case. This fixes both paths and needs no application-level lock. `ON CONFLICT` would also make Path 1's `IntegrityError` disappear by construction rather than by catching it.

3. **Retry on `IntegrityError`** instead of rolling back and giving up — on the create path the conflict means "someone else just created it", so re-reading and applying the delta is the correct recovery.

4. **Reconsider committing the order before the position update.** Right now a failed position update leaves a `complete` order behind, which is the part that makes this invisible to API clients. Either bring both into one transaction, or surface the failure in the `placeorder` response so callers can react.

Happy to test a patch against the reproduction above.

santhiprakash:
## Why

In sandbox / analyzer mode, two `placeorder` requests for the same symbol arriving concurrently can silently lose one position increment. Both orderbook rows still show `complete`, but the positionbook reflects only one of them.

Repro and impact are documented in #1808: an end-of-day flatten with two same-second pairs per symbol was reported as fully `complete` in the orderbook while the positionbook still showed both doubled symbols as `short 60`, which led the operator to manually re-close positions the orders had already closed.

## Root cause

`ExecutionEngine._update_position()` in `sandbox/execution_engine.py` did a read-modify-write on `SandboxPositions` with no Python-level lock and no SQL-level row lock. SQLite has no `SELECT ... FOR UPDATE`, so two concurrent fills could both snapshot the same `position.quantity`, each compute `old + its own delta`, and both write the result back. Last writer wins and one increment is silently lost. There were actually two failure paths from this:

- **Path 1 (loud):** both threads take the `if not position` branch, both `INSERT`, and the second trips the `UniqueConstraint` on `(user_id, symbol, exchange, product)`.
- **Path 2 (silent):** both threads read the same `old_quantity`, both write the same `final_quantity`. No exception, nothing logged. This is the path the production incident took, which is why nothing surfaced.

## Fix

Add a class-level `threading.Lock` on `ExecutionEngine` and acquire it across the full `_update_position` body, including the eventual `db_session.commit()`. Released in a `finally` block so the rollback path also unblocks the next caller.

```python
class ExecutionEngine:
    _position_lock = threading.Lock()

    def _update_position(self, order, execution_price):
        ...
        self._position_lock.acquire()
        try:
            ...  # entire existing RMW body, including db_session.commit()
        except Exception as e:
            db_session.rollback()
            ...
            raise
        finally:
            self._position_lock.release()
```

Class attribute (not instance) is required because `OrderManager.place_order` constructs a fresh `ExecutionEngine()` on every call — an instance lock would give each request its own mutex and the race would remain.

The lock covers the SELECT through the COMMIT, which is the RMW critical section. Multi-process deployments would still need a DB-level lock; that's out of scope for this issue but worth noting in the code comment.

## Test

Added `test/sandbox/test_concurrent_orders.py`. It pre-seeds a 5-share long position, then fires two concurrent `ExecutionEngine._update_position` calls from a `ThreadPoolExecutor`, released by a `threading.Barrier` so the race actually happens under CPython's GIL (without the barrier the two threads tend to be serialized by the interpreter before they reach the critical section and the bug would not manifest).

Expected (with fix): `quantity == 5 + 2 * 10 == 25`.
Observed (without fix): `quantity == 5 + 10 == 15` — one increment lost to last-write-wins.

The test exercises Path 2 (silent) rather than Path 1 (loud) because Path 2 is the one that bit production: Path 1 leaves an `IntegrityError` in the log, Path 2 leaves nothing.

### New regression test

```
$ uv run pytest test/sandbox/test_concurrent_orders.py -v
collected 1 item

test/sandbox/test_concurrent_orders.py::test_concurrent_same_symbol_buys_accumulate_position PASSED [100%]

============================== 1 passed in 1.28s ===============================
```

Without the fix the same test fails deterministically with the lost-update:

```
$ git stash push -- sandbox/execution_engine.py
$ uv run pytest test/sandbox/test_concurrent_orders.py -v
AssertionError: expected position.quantity == 25 after two concurrent
same-symbol BUYs on top of a 5-share seed (issue #1808), got 15 -- one update
was lost to the read-modify-write race
assert 15 == (5 + (2 * 10))
```

I confirmed the failure 10/10 runs without the fix and 8/8 runs with the fix, so the test is reliable and not flaky.

### Full sandbox suite

```
$ uv run pytest test/sandbox/ -v --timeout=30
================= 124 passed, 10 skipped in 104.13s (0:01:44) ==================
```

The 10 skipped are the live-quote integration tests that need a broker session; they were already skipped on `main` and are unrelated to this change.

### Lint

```
$ uv run ruff check sandbox/execution_engine.py test/sandbox/test_concurrent_orders.py
All checks passed!
$ uv run ruff format --check test/sandbox/test_concurrent_orders.py
1 file already formatted
```

(`sandbox/execution_engine.py` has one pre-existing `F821 undefined name 'get_config'` at line 1031 in the `if __name__ == "__main__":` block, which is unchanged by this PR.)

## Notes

- Fix is intentionally narrow — single lock, single function, no behavior change. No changes to the orderbook path, margin path, or any other RMW.
- The order is committed as `complete` *before* `_update_position` runs (line ~586), and a rollback inside `_update_position` only rolls back the position update. That's the part that makes Path 2 invisible to API callers. Closing that gap is a separate, larger change (#1808 suggests bringing both into one transaction or surfacing the failure in the `placeorder` response) and is intentionally out of scope here.
- Happy to add a per-`(user_id, symbol, exchange, product)` lock instead of the class-level one if maintainers prefer finer-grained serialization. I went with class-level because the bug as filed is intra-engine and the class-level lock is the minimal correct fix.

Fixes #1808

---
## Summary by cubic
Serializes concurrent sandbox position updates so same-symbol fills completing at the same time no longer silently lose one increment: `_update_position` previously did an unlocked read-modify-write, and now a class-level `threading.Lock` guards the full RMW (including commit) per process. Order status semantics are unchanged.

- The lock is class-level because each `placeorder` builds a fresh engine, which would defeat an instance-level lock.
- Adds `test/sandbox/test_concurrent_orders.py`, which pre-seeds a position and fires six concurrent pairs per run; the multi-round approach catches the lost-update race deterministically instead of in roughly two of three runs.
- Scope is limited to position writes; multi-process deployments still require DB-level row locking.

<sup>Written for commit adce8abe8f2c890299bc5ff57ab1116a28e617de. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1811?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/sandbox/execution_engine.py
+++ b/sandbox/execution_engine.py
@@ -13,6 +13,7 @@ Features:
 
 import os
 import sys
+import threading
 import time
 import uuid
 from datetime import datetime
@@ -75,6 +76,14 @@ class ExecutionEngine:
         self.order_rate_limit = int(os.getenv("ORDER_RATE_LIMIT", "10 per second").split()[0])
         self.api_rate_limit = int(os.getenv("API_RATE_LIMIT", "50 per second").split()[0])
         self.batch_delay = 1.0  # 1 second between batches
+
+    # Class-level lock (issue #1808). Each placeorder request constructs a fresh
+    # ExecutionEngine() inside OrderManager.place_order, so an instance-level
+    # lock would give every request its own mutex and the race would remain.
+    # A class attribute is shared across every instance in this process and
+    # therefore serializes the position RMW for them. Multi-process deployments
+    # would still need a DB-level lock.
+    _position_lock = threading.Lock()
 
     def check_and_execute_pending_orders(self):
         """
@@ -667,6 +676,12 @@ class ExecutionEngine:
         or during immediate execution (for MARKET orders). We only need to release margin when
         positions are closed/reduced.
         """
+        # Serialize the position RMW (issue #1808). Without this lock two
+        # concurrent same-symbol fills can both read the same old_quantity,
+        # each compute old + its own delta, and last-write-wins -- silently
+        # losing one increment. The lock is released even when the function
+        # raises or rolls back its transaction.
+        self._position_lock.acquire()
         try:
             fund_manager = FundManager(order.user_id)
 
@@ -966,6 +981,8 @@ class ExecutionEngine:
             db_session.rollback()
             logger.exception(f"Error updating position for order {order.orderid}: {e}")
             raise
+        finally:
+            self._position_lock.release()
 
     def _calculate_realized_pnl(self, old_quantity, avg_price, close_quantity, close_price, contract_value=1.0):
         """Calculate realized P&L for closed positions, multiplied by contract_value (e.g. 0.01 for ETHUSD.P)."""
--- a/test/sandbox/test_concurrent_orders.py
+++ b/test/sandbox/test_concurrent_orders.py
@@ -55,6 +55,11 @@ PER_ORDER_MARGIN = Decimal("50000.00")
 # one; the silent lost-update only happens once a position already exists.
 SEED_QTY = 5
 
+#: Concurrent pairs fired per run. One pair reproduces the race only about
+#: two thirds of the time, so a single round would miss a reintroduced bug
+#: one run in three. See the test docstring for the measurement.
+ROUNDS = 6
+
 
 def _reset():
     """Wipe this test user's sandbox rows, reset funds, and seed a position.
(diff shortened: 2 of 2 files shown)


santhiprakash:
Hi @marketcalls, this is a small, self-contained fix for the concurrent position-update race reported in #1808. All CI checks pass, the branch is clean, and it includes a regression test for the lost-update case. Could you take a look when you have a moment? Happy to address any feedback.

santhiprakash:
Hi @marketcalls — friendly check-in on this small fix for the concurrent same-symbol position race in #1808. The branch is current with `main` and all required CI jobs are green. I know the queue is busy, so just let me know if anything needs adjusting.

marketcalls:
Will look into all your fixes all us some time. This week we are completely focusing on testing and validating all the PRs

PR #1860, opened 2026-08-22
fix: honor disabled CORS configuration

### LINKED ISSUE #1848: security: Make CORS_ENABLED=false actually disable Flask-CORS
## Summary

Setting `CORS_ENABLED=FALSE` returns an empty configuration, but Flask-CORS is still initialized and applies its defaults to API routes.

## Evidence

- `cors.py:16-20` treats disabled CORS as an empty dictionary and explicitly notes that defaults will be used.
- `cors.py:55-56` constructs `CORS` for `/api/*`.
- `app.py:185` always calls `cors.init_app(app, **get_cors_config())`.

## Why this matters

Operators reasonably expect a disabled switch to emit no cross-origin headers. Falling back to library defaults can expose API responses to origins the operator did not configure.

## Scope

Initialize Flask-CORS only when explicitly enabled, or return an explicit disabled state that `app.py` honors. Preserve the configured allowlist behavior when enabled.

## Acceptance criteria

- [ ] With `CORS_ENABLED=FALSE`, API responses contain no `Access-Control-Allow-Origin` header.
- [ ] With CORS enabled and a listed origin, the expected header is present.
- [ ] An unlisted origin is not allowed.
- [ ] Tests restore environment variables and require no running server.

## Verification

```powershell
uv run pytest test/test_cors_config.py -v
uv run ruff check cors.py app.py test/test_cors_config.py
```

## Security guidance

- Difficulty: Easy
- Estimated time: 1-3 hours
- You will learn: secure defaults, Flask extension initialization, and response-header tests.

ANONYMOUSZED-beep:
## Summary
- initialize Flask-CORS for `/api/*` only when `CORS_ENABLED` is explicitly enabled
- fail closed when CORS is enabled without configured origins
- log whether central CORS registration is disabled, enabled, or missing origins
- run disabled, unset, preflight, fail-closed, allowed-origin, rejected-origin, and non-API regression cases in CI

Closes #1848

## Scope
This PR governs the central Flask-CORS registration for `/api/*`. Blueprint handlers decorated directly with `@cross_origin()` use a separate Flask-CORS path and are tracked in follow-up issue #1891.

## Upgrade behavior
Installations whose `.env` predates the addition of `CORS_ENABLED` may not contain that key. Unset follows the documented `FALSE` default, so central `/api/*` responses will stop emitting CORS headers. Installations that require cross-origin API access must set `CORS_ENABLED=TRUE` and configure `CORS_ALLOWED_ORIGINS`. The same-origin React UI, Vite proxy, and server-to-server integrations are not expected to require this setting.

## Validation
- Original focused local run before review updates: `pytest test/test_cors_config.py -v`, 3 passed
- Original focused Ruff checks passed
- Backend lint passed on the expanded patch
- Latest backend CI passed with all 60 tests, including all 7 CORS regression cases
- Latest local focused verification: `uv run pytest test/test_cors_config.py -q`, 7 passed
- `uv run ruff format --check test/test_cors_config.py` passed; workflow YAML parsing and `git diff --check` also passed

Note: `app.py` has a pre-existing Ruff formatting difference around `warm_analytics()` outside this change, so no unrelated formatting churn was added.


marketcalls:
Thanks for tracking this down. The bug is real and I reproduced it on main: with `CORS_ENABLED=FALSE`, `cors.py:56` still builds `CORS(resources={r"/api/*": get_cors_config()})`, the empty per-resource dict merges over the flask_cors 6.0.5 defaults (`origins: "*"`, `always_send: True`), and `GET /api/test` with `Origin: https://evil.example` comes back with `Access-Control-Allow-Origin: https://evil.example`. The preflight returns 200 with `Access-Control-Allow-Headers: X-API-KEY`. After your patch both are gone. I also diffed the enabled path (CORS_ENABLED=TRUE with the origins, methods, headers and max-age values from `.sample.env`) between main and your branch and the emitted headers are identical, so the default install sees no change. Your note about the pre-existing `warm_analytics()` Ruff format difference in `app.py` is accurate, I confirmed it exists on main and is well outside your hunks.

Before this can merge:

1. `.github/workflows/ci.yml`, backend-test job, lines 47 to 53. `test/test_cors_config.py` is not in the explicit CI-safe file list, and pytest is invoked with that file list rather than a directory, so the regression guard never runs. The tests are hermetic (plain Flask app, no DB, no broker) and finish in 0.11 seconds here, so add the file to that list. This matters more than usual because CI has not run on this branch at all: the only green checks are Snyk, which reported no manifest changes, and the cubic reviewer.

2. `cors.py`. The title says "honor disabled CORS configuration" and the description says no CORS headers are sent when disabled, but that holds only for the `/api/*` resource. Twelve blueprints import `flask_cors.cross_origin` directly and decorate 21 handlers with `@cross_origin()`: `blueprints/oitracker.py:26` and `:78`, `blueprints/arbitrage.py:25`, `blueprints/gex.py:25`, `blueprints/ivchart.py:21`, `:70`, `:110`, `blueprints/ivsmile.py:25`, `blueprints/oiprofile.py:30`, `:99`, `blueprints/pnltracker.py:207`, `blueprints/straddle_chart.py:21`, `:70`, `blueprints/strategy_chart.py:30`, `:93`, `:156`, `blueprints/custom_straddle.py:22`, `:84`, `:117`, `blueprints/vol_surface.py:20`, `blueprints/gamma_density.py:29`. `cross_origin` resolves its options from the library defaults plus `current_app.config`, not from the `CORS(app)` extension that `init_cors` now gates, and we set no `app.config` CORS keys. With your patch applied and `CORS_ENABLED=FALSE` I still get `Access-Control-Allow-Origin: https://evil.example` from `/oitracker/api/oi-data`. Either route those decorators through the same `is_cors_enabled()` gate, or drop the "no CORS headers when disabled" wording and open a follow-up issue.

3. `cors.py`, `get_cors_config()`. You rewrote this function and left it fail-open. With `CORS_ENABLED=TRUE` and `CORS_ALLOWED_ORIGINS` unset or empty there is no `origins` key, flask_cors falls back to `origins="*"`, and every origin is reflected. Set `CORS_ALLOW_CREDENTIALS=TRUE` on top of that and `/api/*` answers any website with the reflected origin plus `Access-Control-Allow-Credentials: true`, which is full credentialed cross-origin access to the trading API. I confirmed both. It is pre-existing, not something you introduced, but our own audit script only greps for a literal `*` in the origins string (`.claude/skills/security-audit/audit.py:211` to `:221`) and grades that configuration as "CORS origins pinned", so nothing else catches it. Since you are already inside this function, make enabled-with-no-origins either fail closed or log a warning.

4. `cors.py` has no module logger. CLAUDE.md requires `logger = get_logger(__name__)` in every module. That mattered less when registration was unconditional, but now `init_cors()` silently does nothing on the disabled path. One `logger.info` at startup stating whether CORS was registered and for which origins gives an operator something to look at when the headers disappear after an upgrade.

5. `test/test_cors_config.py`. Add a case with `monkeypatch.delenv("CORS_ENABLED", raising=False)`, since unset is the real default branch of `is_cors_enabled()` and no test exercises it. Add a preflight case (`OPTIONS` with `Access-Control-Request-Method` and `Access-Control-Request-Headers: X-API-KEY`), because the preflight is where the interesting headers were leaking. A negative case asserting that a non `/api` path is untouched is also worth having, since `r"/api/*"` is a regex: it matches `/apikey` as well as `/api/v1/...`, which I confirmed by request.

6. PR description. Please record the upgrade behavior change. `CORS_ENABLED` only entered `.sample.env` in 8658f6a38 (2025-05-11), so any `.env` predating that has no such key, takes the FALSE default, and will stop receiving `Access-Control-Allow-Origin` on `/api/*` and `/apikey` once this lands. `utils/env_check.py` only warns and prompts on an old config version, so those installs are not blocked from starting. The blast radius is small (the React UI is same origin, the vite dev server proxies `/api` to localhost:5000, and the TradingView, Chartink and Excel integrations are server to server), but it belongs in the description and ideally in a release note. Also rewrite the em dashes in the Validation section as commas or full stops; the repo forbids em and en dashes in PR descriptions and generated text. The diff itself is clean ASCII.

No migration is needed, this is environment-variable behavior with no schema change, so nothing to add under `upgrade/migrate_all.py`.

ANONYMOUSZED-beep:
Thanks for the detailed review. I addressed the central /api/* items: the regression file is now in the CI-safe list, enabled-without-origins fails closed, startup logging was added, and the tests now cover unset, preflight, missing origins, and a non-API path. I also updated the description with the upgrade behavior and removed dash punctuation from the prose.

I took the scoped option you offered for the direct @cross_origin() handlers and filed #1864 to route those decorators through the same policy. The PR description now explicitly limits this change to the central /api/* registration.

The updated Python files parse locally. I did not install dependencies or claim a new local pytest run; the expanded tests are now running in GitHub Actions.

ANONYMOUSZED-beep:
Correction: the decorator follow-up created by this run is #1891 (not #1864). The PR description has been corrected.

ANONYMOUSZED-beep:
CI ran the expanded backend suite: 46 passed and the new preflight test failed because the assertion iterated `(name, value)` header tuples and called `.lower()` on the tuple. I corrected the test to iterate `response.headers.keys()`. This was a test-only issue; the disabled preflight response itself had no CORS leak. A fresh CI run is now starting.

marketcalls:
Merging after one mechanical fix. I checked every item against the head (70a032d2) rather than the summary.

Confirmed done:

- `ci.yml:52` now carries `test/test_cors_config.py`, and the backend-test run on this head shows all seven cases executing: 47 passed in 5.20s. That closes the blocker, and CI running on fork PRs now settles my other objection.
- Fail closed works. I A/B ran main against your branch over seven configurations. With `CORS_ENABLED=TRUE` and no `CORS_ALLOWED_ORIGINS`, main answers `/api/v1/placeorder` and `/apikey` with `Access-Control-Allow-Origin: https://evil.example`, and adds `Access-Control-Allow-Credentials: true` once `CORS_ALLOW_CREDENTIALS=TRUE`. Your branch sends no CORS headers in both cases. `cors.py:23` dropping empty segments is what does it. The `.sample.env` enabled path is identical between main and your branch on GET and on preflight, so the default install is unaffected.
- `cors.py:7-9` has the module logger, and all three states are logged, including the warning at `:64`.
- Scope option executed correctly: the Scope section limits the claim to the central `/api/*` registration, and #1891 exists with the right title. Thanks for catching your own wrong issue number rather than leaving #1864 to point somewhere else.
- PR body has the Upgrade behavior section, and title plus body scan clean ASCII.
- Your account of the earlier CI failure holds up: job 97381779524 fails at `test_cors_config.py:51` with `AttributeError: 'tuple' object has no attribute 'lower'`, 1 failed and 46 passed. Reporting that instead of quietly force pushing is appreciated.

To fix before I merge:

Line endings. `cors.py`, `test/test_cors_config.py` and `.github/workflows/ci.yml` are committed CRLF on your branch. Main is LF everywhere, including `cors.py` at your own branch point 6637e088. That is why the diff reads +525/-420 for a 25 byte change: git sees ci.yml as 359 lines deleted and 359 added. Normalized, your ci.yml differs from main by exactly one line, the one adding the test file. Please re-commit those three with LF (`git config core.autocrlf input`) so blame on ci.yml and cors.py survives. While you are there, squash 4383bced and 9ddcf5bc: 4383bced, "test: run CORS regressions in CI", actually truncated ci.yml to 0 bytes, and 9ddcf5bc restored it. The end state is correct, I verified it, but an empty CI workflow should not sit in history.

Two optional items:

1. The disabled path tests are green for the wrong reason. I deleted `if not is_cors_enabled(): return` from `init_cors` and all seven tests still passed, because every disabled case also clears `CORS_ALLOWED_ORIGINS`, so the new fail closed behavior masks the missing gate. Give `CORS_ALLOWED_ORIGINS` a real value in those tests, which is the realistic shape anyway since `.sample.env` ships it alongside `CORS_ENABLED`. With that, the gate-removed build returns `Access-Control-Allow-Origin: http://127.0.0.1:5000` and the test fails as it should.
2. `uv run ruff format --check test/test_cors_config.py` reformats four blocks and line 51 is 102 characters against the 100 limit. The version I first reviewed was format clean. backend-lint is continue-on-error, so CI will not flag it.

ANONYMOUSZED-beep:
Thanks for the careful follow-up. I pushed signed commit 2d03a04 with the three requested files normalized to LF. I also took both optional test improvements: disabled and unset cases now keep a realistic configured origin, so deleting the CORS_ENABLED gate makes them fail, and test/test_cors_config.py is Ruff-formatted. Local verification: uv run pytest test/test_cors_config.py -q (7 passed) and uv run ruff format --check test/test_cors_config.py (passed).\n\nI have not rewritten the published branch history to remove 4383bced and 9ddcf5bc because that requires a force push. The current tree is corrected and CI is running on 2d03a04; I can perform that history rewrite once explicitly authorized.

ANONYMOUSZED-beep:
Upstream main introduced a conflict in the backend CI-safe test list. I resolved it by retaining all three newly required entries: the CORS regression test from this PR plus upstream's multi-option Greeks and history-format regressions. Signed merge commit: 938a0e67.\n\nLocal verification after the merge:\n- uv run pytest test/test_cors_config.py -q: 7 passed\n- uv run ruff format --check test/test_cors_config.py: passed\n- workflow YAML parsed successfully\n- git diff --check: passed\n\nCI should now run on the conflict-free head.

marketcalls:
Approving the code. One condition on how it goes in, and then I am done.

The line endings are genuinely fixed, and I checked the blobs rather than the patch. cors.py 2129 bytes, app.py 55857, test/test_cors_config.py 3252, ci.yml 12726, CR count zero on all four. I also enumerated every blob in both trees: 50 CRLF files at your head, the identical 50 on main, so you introduced none and converted none by accident. Against main your ci.yml differs by exactly one line with the multi_option_greeks and history_format entries untouched, and app.py by exactly the two hunks you intended. Compare against main HEAD gives ahead 8, behind 0, four files, +135/-21.

Both optional items landed and the test one mattered. I re-ran last round's mutation: deleting the enable gate from init_cors now fails three tests where previously all seven still passed, and putting fail-open back still fails test_enabled_cors_without_origins_fails_closed. I added a third, widening the resource regex to r'/*', and test_non_api_path_is_untouched catches it. All seven cases execute in CI, backend-test job 97492960209 on 938a0e67 shows 60 passed in 6.50s, and I reproduced 60 passed locally. ruff is clean on both your files; the only complaint is the pre-existing warm_analytics blank line in app.py, which I reproduced against origin/main in isolation exactly as you said.

I went one step past the unit tests and ran the real app. import app and create_app() succeed at your head, and with .env CORS_ENABLED='FALSE' the live app sends no Access-Control-* on /api/config/host, on /apikey or on the /api/v1/placeorder preflight, while with 'TRUE' plus http://127.0.0.1:5000 it reflects only that origin. I also confirmed the fail-closed is not luck: Flask-Cors is pinned ==6.0.5 in both pyproject.toml and requirements.txt, and sanitize_regex_param([]) returns [] so nothing can match. And I checked start.sh before worrying about Docker upgrades: its .env block only runs when HOST_SERVER is set and seeds CORS_ALLOWED_ORIGINS from it, so no generated install lands on the enabled-with-no-origins path.

Now the condition. I told you last round that the 0-byte ci.yml commits could not reach main because this repo squashes. I was wrong to state that as a fact. The repo settings still have allow_merge_commit true and allow_rebase_merge true, so Rebase and merge would put 4383bced1 and 29603554f, plus the four CRLF commits, straight onto main. Squashing is the habit here, not a rule. You were right not to force-push a published branch without authorization, so I am not asking you again: this needs to be merged with the Squash button, and I am flagging that to whoever merges rather than treating it as settled.

Two things I am not holding this for, one of which I missed last round. There is still no lowercase 'true' case for CORS_ENABLED. Beyond that, two mutations survive your whole suite: dropping the 'if origin.strip()' filter, and loosening the gate from == "TRUE" to != "FALSE" both leave all seven green, and there is no test for an empty or whitespace CORS_ALLOWED_ORIGINS string at all. Your shipped code is right in every one of those cases, I verified each by A/B; the suite just does not pin them. Separately, no CI test imports app.py, so the two-line wiring that actually makes this take effect in production is uncovered. A one-line assertion somewhere would close that.

Last thing: the Validation section still says "fresh CI is running" and describes the tuple-iteration failure that no longer exists. Please tidy that line, then squash and merge. Good patch, and thanks for three rounds of clean follow-through.

ANONYMOUSZED-beep:
Thanks for the final review and approval. I updated the Validation section to remove the stale in-progress and earlier-failure wording and record the final green CI and focused local checks. I also noted the squash-merge requirement; no history rewrite was performed.

PR #1861, opened 2026-08-22
fix: return client errors for invalid calendar dates

### LINKED ISSUE #1824: python: Return client errors for non-string calendar dates
## Summary

Calendar helpers catch malformed strings but non-string inputs can raise `TypeError` and turn a client mistake into a server error.

## Evidence

- `services/market_calendar_service.py:72` parses the timings date.
- `services/market_calendar_service.py:117` parses the holiday-check date.
- Current handling covers `ValueError` but not every invalid type.

## Scope

Make both entry points return their documented client-error result for non-string or unparseable dates.

## Acceptance criteria

- [ ] `None`, numeric, empty, and malformed values are controlled errors.
- [ ] Valid ISO dates still work.
- [ ] Both service paths behave consistently.
- [ ] No `TypeError` escapes.

## Verification

```powershell
uv run pytest test/test_market_calendar_validation.py -v
uv run ruff check services/market_calendar_service.py test/test_market_calendar_validation.py
```

- Difficulty: Easy
- Estimated time: 1-2 hours
- You will learn: exception boundaries and service validation.

nightcityblade:
## Summary

- defensively handle non-string date values in internal market-calendar helpers through the same 400 response path as malformed strings
- cover both market-timing and holiday-check entry points for `None`, numeric, empty, malformed, and valid ISO dates

## Testing

- `uv run pytest test/test_market_calendar_validation.py -q`: 10 passed
- `uv run ruff check services/market_calendar_service.py test/test_market_calendar_validation.py`: passed
- `uv run ruff format --check services/market_calendar_service.py test/test_market_calendar_validation.py`: passed

Fixes #1824

---
## Summary by cubic
Return 400 client errors for invalid calendar dates across market calendar endpoints. Previously, non-string or malformed inputs raised a TypeError; now both endpoints return 400 with "Invalid date format. Use YYYY-MM-DD", meeting #1824’s requirement for consistent client errors.

- Catch TypeError and ValueError when parsing date_str in get_timings and check_holiday; valid inputs unchanged.
- Add validation tests (test/test_market_calendar_validation.py) and register them in CI.

<sup>Written for commit 4aaf912356de9ed2d65a4a38e118686a0996df75. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1861?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


marketcalls:
Decision: merge after one small fix. The code change itself is correct and I am not asking you to touch `services/market_calendar_service.py` again. What is missing is CI registration for the new test file.

What I verified locally against your head commit 2e18077d:

- The patch applies cleanly to current `main`.
- `uv run pytest test/test_market_calendar_validation.py -q` gives 10 passed. Reverting only `services/market_calendar_service.py` back to `main` makes exactly 4 of them fail with `assert 500 == 400` plus a logged `TypeError: strptime() argument 1 must be str, not int`, so the tests genuinely pin the behaviour you changed.
- `uv run ruff check` and `uv run ruff format --check` are clean on both files. On `main` that service file had 3 ruff errors (UP035 twice on line 8, F841 for `today` on line 77) and your diff clears all three. `date` is still imported and still used at lines 78-79, so no NameError from dropping the unused local.
- It matches issue #1824's acceptance criteria, including the exact test path the issue named.

Required before merge

1. `.github/workflows/ci.yml`, the `Run CI-safe tests` step (pytest list at lines 47-53): add `test/test_market_calendar_validation.py`. Backend tests in this repo run in CI only if the file is named explicitly in that step, so as it stands your tests never execute in CI and give no regression protection. I confirmed the file is safe to add there: `test/conftest.py` assigns `DATABASE_URL=sqlite:///db/openalgo-test.db` before your module is imported, and your tests monkeypatch `get_market_timings_for_date` and `is_market_holiday`, so no DB connection is opened. Worth knowing that `database/market_calendar_db.py` builds its engine and `scoped_session` at import time (lines 39-46), not lazily, so it is the conftest assignment that keeps this safe.

2. PR description: the Testing section has three em dashes (U+2014). `CLAUDE.md` bans em dashes and en dashes everywhere, PR descriptions included. Use a comma or a colon.

Ordering note: PRs #1861, #1862, #1875 and #1884 each add a new test file and each need a line in that same `ci.yml` block, so those edits will conflict textually. Put your filename on its own continuation line right after `test/test_telegram_api_contract.py` (line 52), and if one of the others lands first, resolve the rebase by keeping both filenames rather than replacing.

Optional follow-up, not blocking

3. `services/market_calendar_service.py:121` in your branch, one line below the guard you just fixed: `if exchange and exchange.upper() not in SUPPORTED_EXCHANGES` raises `AttributeError` for a non-string exchange. I ran `check_holiday("2026-08-22", 123)` and got `(False, {... 'An error occurred while checking holiday status'}, 500)`. Same class of defect as the date argument. Fixing it here, with a parametrized case, would be in scope for #1824's "no TypeError escapes" criterion.

4. The PR body reads as if non-string dates currently reach the API as a `TypeError`. They do not, on the deployed surface. `/api/v1/market/timings` loads through `MarketTimingsSchema`, whose `date` field is `fields.Str(required=True)` (`restx_api/data_schemas.py:246`). I ran that schema: `None` gives "Field may not be null.", `123` gives "Not a valid string.", both returning 400 at `restx_api/market_timings.py:41`. The only other caller of `get_timings` is `services/flow_openalgo_client.py:518`, fed from `flow_executor_service.py:2220` via `get_str`, which always yields a str. And `services.market_calendar_service.check_holiday` has no Python caller at all. So this is defence in depth on an internal helper, which is exactly what #1824 asked for. Please phrase it that way rather than as a user-visible 500.

Separately, and not something to fix in this PR: `mcp/mcpserver.py:1615` posts to `/api/v1/checkholiday`, and no such namespace is registered in `restx_api/__init__.py` (only `/market/holidays` and `/market/timings`). That MCP tool is dead. I will open a separate issue.

One thing to be aware of: CI has not run on this branch. Fork PRs wait on maintainer approval in this repo, and the green checks you see are Snyk reporting no manifest changes plus the cubic reviewer, so no Ruff run, no pytest run and no frontend build has validated your branch. Everything above comes from my own local run, which is also why item 1 matters: without the `ci.yml` line, that stays true after merge.

Push the `ci.yml` line and fix the description, and I will merge.

nightcityblade:
Addressed the two required items in f0b60e3:

- registered `test/test_market_calendar_validation.py` in the CI-safe pytest list, on its own continuation line after the Telegram contract test
- replaced the em dashes in the Testing section and clarified that this is defense in depth for the internal helpers

Validation: 10 focused tests passed; Ruff check and format check passed on both Python files; `git diff --check` passed.

nightcityblade:
Upstream changed the same CI list after my first push, so I merged current `main` without force-pushing and kept all registrations in 7c0ef63c: the new market-calendar test now sits alongside `test_multi_option_greeks_regression.py` and `test_history_format.py`.

Validation on the merged tree: 10 focused tests passed; Ruff check and format check passed on both Python files; `git diff --check` passed.

nightcityblade:
Resolved the merge conflict with current `main`, retaining both the new CORS test and this PR's market-calendar validation test in CI.

Verification after the merge:
- `uv run pytest test/test_market_calendar_validation.py -q` — 10 passed
- `uv run ruff check services/market_calendar_service.py test/test_market_calendar_validation.py` — passed
- CI workflow YAML parse — passed

marketcalls:
Merged as `242f80d4`, and #1824 is closed. Your test file is live in the CI-safe list on main, so it now runs on Python 3.12, 3.13 and 3.14 on every future PR.

Thanks for the clean turnaround on both required items: the `ci.yml` registration went in on its own continuation line so it stacked under the entries that landed while you were working, and the description em dashes are gone.

## What I checked before merging

The `except` widening was never the risk in this diff. The import trim was, so I verified it rather than eyeballing it. `from typing import Any, Dict, Optional, Tuple` becoming `from typing import Any` is only safe if nothing still references the dropped names, and grep does report four hits:

```
line  29:  Tuple of (success, response_data, status_code)
line  67:  Tuple of (success, response_data, status_code)
line 108:  exchange: Optional exchange code to check
line 111:  Tuple of (success, response_data, status_code)
```

All four are docstring prose. An AST walk over your head blob confirms zero uses of `Dict`, `Optional` or `Tuple` in actual code, and the three annotations are `tuple[bool, dict[str, Any], int]`, builtin generics that need no import. `date` is still imported and still used by `min_date` and `max_date`, and `today` has no surviving reference. Clean.

I also traced whether this could disturb the Python Strategy Host, since `/python` scripts can reach these services:

- The `openalgo` SDK exposes `timings()` and `holidays()`, and `timings()` resolves to `self._make_request("market/timings", payload)`, an HTTP call. That route loads `MarketTimingsSchema`, where `date = fields.Str(required=True)`, so `None` gives "Field may not be null" and `123` gives "Not a valid string", both 400 before your code is entered. Unchanged either way.
- `services/flow_openalgo_client.py:518` is fed by `get_str`, which always yields a str, so the Flow node path is unchanged too.
- The only observable difference is a direct in-process import with a non-string argument, which used to fall through to the outer `except Exception` and return 500 and now returns your 400. Strictly an improvement; nothing that previously worked returns anything different.

## One honest note on scope

This scored 96 on merge readiness, which measures craft: a two-line behavioural change, no regression, ten tests that fail without the fix, correct CI registration, no scope creep, `behind` of zero. On that axis it is close to perfect and it is why it merged with no further rounds.

It is worth being straight that the impact is low, and you said as much yourself when you re-pitched the description as defence in depth after I pushed back. The `TypeError` this now catches cannot arrive from the deployed API, because the schema blocks it first. That does not make the change wrong, and I merged it: it implements the issue as written, it removes three pre-existing ruff errors, and a service function should not answer 500 for a bad argument. But if you are choosing what to work on next, prefer the issues where the defect is reachable.

## Two follow-ups you surfaced, neither yours to fix

While verifying reachability I confirmed both of the things you flagged earlier:

1. `check_holiday` has no Python caller anywhere in the repo. `restx_api/__init__.py` registers `/market/holidays` and `/market/timings` and nothing routes to it.
2. `mcp/mcpserver.py:1589` advertises a `check_holiday` MCP tool that POSTs to `/api/v1/checkholiday`, a namespace that does not exist, and `utils/mcp_tool_registry.py:99` grants it `SCOPE_READ_MARKET`. That tool is dead: an MCP client calling it gets a 404.

The second is a real bug and worth its own issue. If you want it, open one and take it; you already have the context.

PR #1862, opened 2026-08-22
fix: reject NaN and infinity in API analyzer numeric fields

### LINKED ISSUE #1822: python: Reject NaN and infinity in API analyzer numeric fields
## Summary

The API analyzer converts values with `float()` and uses range comparisons, but `NaN` and infinity can bypass ordinary comparisons.

## Evidence

- `utils/api_analyzer.py:118` converts input to float.
- `utils/api_analyzer.py:146-148`, `utils/api_analyzer.py:214`, and `utils/api_analyzer.py:462` validate numeric ranges.

## Scope

Add a reusable `math.isfinite` check for analyzer-controlled numeric inputs while preserving finite-number behavior.

## Acceptance criteria

- [ ] `NaN`, positive infinity, and negative infinity are rejected.
- [ ] Current finite boundaries behave as before.
- [ ] Tests cover string and numeric representations.
- [ ] No input values are added to logs.

## Verification

```powershell
uv run pytest test/test_api_analyzer_numeric_validation.py -v
uv run ruff check utils/api_analyzer.py test/test_api_analyzer_numeric_validation.py
```

- Difficulty: Easy
- Estimated time: 1-2 hours
- You will learn: IEEE-754 edge cases and defensive validation.

vaibhav8a:
Closes #1822.

## What does this PR do?

`float()` accepts `"nan"`, `"inf"` and `"-inf"`, and every range check in `utils/api_analyzer.py` is expressed as a comparison (`<= 0`, `< 0`, `== 0`). All of those are `False` for NaN, so a non-finite quantity or price passed validation untouched and reached the broker.

Demonstrated on `main` before the change, with symbol lookup stubbed so only numeric validation is in play:

```
### BEFORE (original code):
  quantity=nan -> success
  price=inf    -> success

### AFTER (with the guard):
  quantity=nan -> error
  price=inf    -> error
```

## Change

One reusable helper, per the issue's scope note:

```python
def _finite_float(value):
    number = float(value)
    if not math.isfinite(number):
        raise ValueError("numeric field must be a finite number")
    return number
```

It replaces `float()` at the analyzer-controlled conversions only — `quantity`, `price`, `trigger_price`, `disclosed_quantity` and `position_size`, across `analyze_api_request`, `analyze_smart_order_request` and `analyze_modify_order_request`. Because it raises `ValueError`, the `except ValueError` / `except (ValueError, TypeError)` handlers already in place report it with the messages they use today — no new error strings and no new branches.

## Acceptance criteria

- [x] `NaN`, positive infinity, and negative infinity are rejected — on all three order paths.
- [x] Current finite boundaries behave as before — `quantity=0` still gives "Quantity must be greater than 0", `price=-1` still gives "Price cannot be negative", `LIMIT` with `price=0` still gives "Price is required for LIMIT orders"; each is asserted in the tests.
- [x] Tests cover string and numeric representations — parametrized over `"nan"`, `"NaN"`, `"inf"`, `"-inf"`, `"Infinity"`, `float("nan")`, `float("inf")`, `float("-inf")`.
- [x] No input values are added to logs — the helper's message is a fixed string; two tests assert the rejected value doesn't appear in the error text.

## Testing

```
uv run pytest test/test_api_analyzer_numeric_validation.py -v   # 99 passed
uv run ruff check utils/api_analyzer.py test/test_api_analyzer_numeric_validation.py   # All checks passed
```

Symbol validation is monkeypatched to `True` in the tests so the assertions are about numeric validation only, and the payload helpers carry every entry of `REQUIRED_ORDER_FIELDS` / `REQUIRED_SMART_ORDER_FIELDS` / `REQUIRED_MODIFY_ORDER_FIELDS` so a missing-field issue can't mask the result.

One fix, one PR — no other analyzer behaviour is touched.

---
## Summary by cubic
Rejects non-finite numbers in API analyzer numeric fields and treats oversized integers as invalid, so bad values are rejected at this layer. Previously `float()` accepted "nan"/"inf" and NaN bypassed range checks; `_finite_float` now raises `ValueError` (also converting `OverflowError`) and existing handlers surface the same errors. This is defense in depth; marshmallow schemas already disallow non‑finite numbers.

- Applies to quantity, price, trigger_price, disclosed_quantity, and position_size in analyze_api_request, analyze_smart_order_request, and analyze_modify_order_request.
- Keeps all finite validations and messages; rejected values are not echoed.
- Adds tests for string/float NaN/Infinity and oversized ints; stubs the rate‑limit probe; wires the test file into CI.

<sup>Written for commit 883c45a57be72d6a07d4264a13967a9a58b4f227. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1862?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,6 +50,7 @@ jobs:
             test/test_auth_resume.py test/test_auth_upsert_multisession.py \
             test/sandbox/test_execution_backlog.py test/test_event_bus_bounded.py \
             test/test_telegram_api_contract.py \
+            test/test_api_analyzer_numeric_validation.py \
             -v --timeout=60
 
   # Frontend linting
--- a/test/test_api_analyzer_numeric_validation.py
+++ b/test/test_api_analyzer_numeric_validation.py
@@ -1,0 +1,158 @@
+"""Non-finite numbers must be rejected by the API analyzer.
+
+``float()`` accepts "nan", "inf" and "-inf", and every range check in
+``utils.api_analyzer`` is a comparison (``<= 0``, ``< 0``, ``== 0``). All of
+those are False for NaN, so before the ``math.isfinite`` guard a NaN quantity
+or price passed validation and reached the broker.
+
+Run with: uv run pytest test/test_api_analyzer_numeric_validation.py -v
+"""
+
+import pytest
+
+from utils import api_analyzer
+from utils.api_analyzer import (
+    _finite_float,
+    analyze_api_request,
+    analyze_modify_order_request,
+    analyze_smart_order_request,
+)
+
+# Both the string spellings that arrive over JSON and the float objects that
+# arrive from a Python client.
+NON_FINITE = ["nan", "NaN", "inf", "-inf", "Infinity", float("nan"), float("inf"), float("-inf")]
+
+QUANTITY_ISSUE = "Invalid quantity value"
+PRICE_ISSUE = "Invalid numeric value for price, trigger_price, or disclosed_quantity"
+# modifyorder validates quantity inside the same try, so its message differs.
+MODIFY_NUMERIC_ISSUE = (
+    "Invalid numeric value for price, trigger_price, quantity, or disclosed_quantity"
+)
+
+
+@pytest.fixture(autouse=True)
+def _known_symbol(monkeypatch):
+    """Take symbol lookup out of play so only numeric validation is asserted."""
+    monkeypatch.setattr(api_analyzer, "validate_symbol", lambda symbol, exchange: True)
+
+
+def _order(**overrides):
+    """A placeorder payload carrying every REQUIRED_ORDER_FIELDS entry."""
+    order = {
+        "apikey": "test-key",
+        "strategy": "test",
+        "symbol": "RELIANCE",
+        "exchange": "NSE",
+        "action": "BUY",
+        "quantity": "1",
+        "pricetype": "MARKET",
+        "product": "MIS",
+    }
+    order.update(overrides)
+    return order
+
+
+def _smart_order(**overrides):
+    """A smart-order payload; position_size is required on this path."""
+    order = _order(position_size="0")
+    order.update(overrides)
+    return order
+
+
+def _modify_order(**overrides):
+    """A modifyorder payload carrying every REQUIRED_MODIFY_ORDER_FIELDS entry."""
+    order = _order(
+        orderid="250101000000001",
+        pricetype="LIMIT",
+        price="100",
+        disclosed_quantity="0",
+        trigger_price="0",
+    )
+    order.update(overrides)
+    return order
+
+
+class TestFiniteFloat:
+    @pytest.mark.parametrize("value", NON_FINITE)
+    def test_rejects_non_finite(self, value):
+        with pytest.raises(ValueError):
+            _finite_float(value)
+
+    @pytest.mark.parametrize("value", ["0", "-1.5", "1e308", 0, 12345.678])
+    def test_accepts_finite(self, value):
+        assert _finite_float(value) == float(value)
+
+    def test_message_does_not_echo_the_input(self):
+        with pytest.raises(ValueError) as excinfo:
+            _finite_float("nan")
+        assert "nan" not in str(excinfo.value).lower()
+
+
+class TestPlaceOrder:
+    @pytest.mark.parametrize("value", NON_FINITE)
+    def test_quantity_rejected(self, value):
+        response = analyze_api_request(_order(quantity=value))
+        assert response["status"] == "error"
+        assert QUANTITY_ISSUE in response["message"]
+
+    @pytest.mark.parametrize("field", ["price", "trigger_price", "disclosed_quantity"])
+    @pytest.mark.parametrize("value", NON_FINITE)
+    def test_price_fields_rejected(self, field, value):
+        order = _order(pricetype="LIMIT", price="100")
+        order[field] = value
+        response = analyze_api_request(order)
+        assert response["status"] == "error"
+        assert PRICE_ISSUE in response["message"]
+
+    def test_finite_order_still_valid(self):
+        response = analyze_api_request(_order(pricetype="LIMIT", price="100"))
+        assert response["status"] == "success"
+
+    def test_finite_boundaries_unchanged(self):
+        assert "Quantity must be greater than 0" in analyze_api_request(_order(quantity="0"))["message"]
+        assert "Price cannot be negative" in analyze_api_request(_order(price="-1"))["message"]
+        assert "Price is required for LIMIT orders" in analyze_api_request(
+            _order(pricetype="LIMIT", price="0")
+        )["message"]
+
+    def test_rejection_does_not_echo_the_input(self):
+        response = analyze_api_request(_order(quantity="nan"))
+        assert "nan" not in response["message"].lower()
+
+
+class TestSmartOrder:
+    @pytest.mark.parametrize("value", NON_FINITE)
+    def test_quantity_rejected(self, value):
+        response = analyze_smart_order_request(_smart_order(quantity=value))
+        assert response["status"] == "error"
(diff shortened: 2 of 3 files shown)


marketcalls:
Thanks for the detailed writeup and the before/after evidence. The narrow defect is real and I reproduced it on pristine main: with symbol lookup stubbed, `analyze_api_request` returns `status: success` for `quantity="nan"`, `price="inf"`, `trigger_price="nan"` and `disclosed_quantity="inf"`, because every range check in `utils/api_analyzer.py` is a comparison and NaN or positive infinity makes all of them False. The `_finite_float` shape is the right one, and the test file is well built: 99 cases, hermetic, under a second, and it pins the finite boundaries and the no-echo property instead of only the happy path.

Here is what has to change before this can go in.

**1. `utils/api_analyzer.py`, the new `_finite_float` docstring: remove the U+2014 em dash.** The line is "The value itself is never interpolated into the message ... analyzer input is logged downstream". CLAUDE.md forbids em dashes and en dashes in every file type. Use a comma, a colon, parentheses or a full stop.

**2. The stated impact is wrong, and the PR writes it into two permanent docstrings.** The three functions you patched are not reachable. `analyze_api_request`, `analyze_smart_order_request`, `analyze_modify_order_request` and `analyze_request` have no callers anywhere in the repo. The only import of this module is `blueprints/analyzer.py:22`, and it pulls in `get_analyzer_stats` only. Analyze mode does not route here either: `services/place_order_service.py:149` sends analyze-mode orders to `sandbox_place_order`. And the live path already rejects these values one layer earlier: `restx_api/place_order.py:32` loads `OrderSchema` from `restx_api/schemas.py`, whose numeric fields are `fields.Float`, and marshmallow's `fields.Float` defaults to `allow_nan=False`. Loading `{"price": "Infinity"}` returns `{'price': ['Special numeric values (nan or infinity) are not permitted.']}`, and the same holds for `SmartOrderSchema` and `ModifyOrderSchema`. So nothing non-finite reached a broker through `/api/v1/`.

Re-pitch this as defence in depth on legacy code, and fix the claim in both places it now lives: the `_finite_float` docstring in `utils/api_analyzer.py`, and the module docstring of `test/test_api_analyzer_numeric_validation.py`, which repeats "reached the broker".

**3. `utils/api_analyzer.py`, the helper's contract is incomplete.** `float()` raises `OverflowError`, not `ValueError`, for a large Python int such as `10**400`. Neither `except (ValueError, TypeError)` on the quantity branch nor `except ValueError` on the price branch catches that, so it escapes to the function level `except Exception` and collapses the whole analysis into "Internal error analyzing request" plus a logged traceback. Your docstring promises the opposite ("Raise ValueError instead, so the existing handlers report it the same way"). Either catch `OverflowError` in the helper and re-raise `ValueError`, or drop that sentence.

**4. Formatting.** `uv run ruff format --check` fails on both files: three blank lines before `def _finite_float` and only one before `def generate_order_id`, the two `disclosed_qty = _finite_float(...)` lines (166 and 271) past the 100 column limit, and two multi-line asserts in `test_finite_boundaries_unchanged`. `ruff check` alone does not catch this, because `pyproject.toml` ignores E501 with the note "handled by formatter". Run `uv run ruff format .` and commit the result.

**5. `_finite_float` needs type hints and a Google style docstring** with Args, Returns and Raises, per the Style section of CLAUDE.md.

**6. `test/test_api_analyzer_numeric_validation.py` will never run in CI.** The `backend-test` job in `.github/workflows/ci.yml` runs an explicit file list, and this file is not in it. The test is fast, deterministic and needs no broker credentials, so add it to that list.

**7. Test noise hides the rate-limit branch.** Every analyzer level case in the new file silently takes the DB failure path: `analyzer_logs` does not exist in `db/openalgo-test.db`, so the rate-limit query raises `OperationalError`, which is swallowed into `warnings: ['Unable to check rate limits']` and written as a full traceback to `log/test/errors.jsonl` on each of the roughly 90 cases. Nothing asserts on it, so a real regression in that branch would be invisible. Either create the table in a fixture or assert the expected `warnings` shape.

One process note: CI has not run on this branch. The only green checks are Snyk, which reported no manifest changes, and the cubic reviewer. Treat lint, backend tests and the frontend build as unvalidated here.

Last item, and it is a question rather than a change request. Given that this module is unreachable, tell me whether you would rather see `utils/api_analyzer.py` deleted instead of hardened. If you want the guard where it protects real production input, the places to look are `restx_api/schemas.py` and the internal callers that build order payloads without passing through a schema: `blueprints/orders.py:558` and `:935`, `blueprints/scalping.py:709`, and `services/flow_openalgo_client.py:88`, `:121`, `:156`. I have not audited whether any of those can actually carry a non-finite value, so treat that as a pointer, not a claim.

vaibhav8a:
Thanks for the thorough review, and for reproducing it on pristine main. You are right on every point, including the one that matters most: my impact claim was wrong. I have pushed the fixes.

**1. Em dash.** Removed.

**2. The impact claim.** Corrected in both places. I verified your reachability finding independently before rewriting it: the only non-test import of `utils.api_analyzer` anywhere in the tree is `blueprints/analyzer.py:22`, and it pulls `get_analyzer_stats` only. The three `analyze_*` functions are reached solely through `analyze_request` in the same module, which itself has no callers. I also confirmed the marshmallow behaviour locally on 3.26.2: `fields.Float().load` rejects `"Infinity"`, `"nan"` and `"1e400"` with "Special numeric values (nan or infinity) are not permitted". Nothing non-finite reached a broker through `/api/v1/`. Both docstrings now describe this as defence in depth on legacy code.

**3. `OverflowError`.** Caught in the helper and re-raised as `ValueError`, so the documented contract holds and callers keep handling a single type. `test_oversized_int_is_reported_as_a_field_issue` pins that `10**400` now surfaces as "Invalid quantity value" rather than collapsing into "Internal error analyzing request".

**4. Formatting.** `ruff format` run on both files; they are clean under `--check`.

**5. Type hints and docstring.** `_finite_float(value: object) -> float` with Args, Returns and Raises.

**6. CI.** Added to the `backend-test` file list in `.github/workflows/ci.yml`.

**7. Test noise.** This was the useful catch. Every case was silently taking the `OperationalError` path and writing a traceback per case, so the rate-limit branch was invisible. I stubbed `AnalyzerLog` rather than creating the table, which keeps the suite hermetic, and pinned all three branches explicitly: quiet, burst over 50, and probe failure asserting `warnings == ["Unable to check rate limits"]`. Tracebacks written to `log/test/errors.jsonl` during the run drop from roughly 90 to 1, and that one is the deliberate failure case. 113 tests, 0.23s.

**On deleting the module.** I do not think `utils/api_analyzer.py` can be deleted as a file, because `get_analyzer_stats` is live: `blueprints/analyzer.py` imports it, and it is self-contained, touching only `AnalyzerLog` and not the validation chain. What is genuinely dead is everything else in there, the six `analyze_*` functions plus `analyze_request` and the `generate_order_id`, `check_rate_limits` and `validate_symbol` helpers they use.

So the real choice is between this PR and a deletion of that dead chain leaving `get_analyzer_stats` behind. I am happy to do the second instead, as a separate PR, and close this one. Deleting is the better long-term answer if you are confident nothing out of tree calls these, since hardening dead code preserves it. That judgement is yours, so tell me which you want and I will follow it.

On your process note: CI had not run on the branch. It should run on this push. I ran the full `backend-test` list locally against this change: 153 passed.

marketcalls (commented):
Reviewed head `883c45a57be72d6a07d4264a13967a9a58b4f227`: 89/100. The helper and focused regression tests are sound; 113 tests passed at this head. No new code defect was found. The current merge conflict is the blocker, and the latest head has no backend CI result. After rebasing, run the configured backend checks and the new numeric validation suite.

PR #1864, opened 2026-08-22
refactor: add type hints to data schemas validators

### LINKED ISSUE #893: python: Add type hints to validation functions in restx_api/data_schemas.py
## Description

Validation functions in \`data_schemas.py\` lack parameter and return type hints, making them harder to understand and use correctly.

## Files Affected

- \`restx_api/data_schemas.py\` (lines 7-35)

## Current Code

```python
def validate_date_or_timestamp(data):  # Missing type hints
    """Validates that the input string is either in 'YYYY-MM-DD' format or a numeric timestamp."""
    date_pattern = re.compile(r"^\d{4}-\d{2}-\d{2}$")
    timestamp_pattern = re.compile(r"^\d{10,13}$")
    if not (isinstance(data, str) and (date_pattern.match(data) or timestamp_pattern.match(data))):
        raise ValidationError(...)

def validate_option_offset(data):  # Missing type hints
    """Validates option offset: ATM, ITM1-ITM50, OTM1-OTM50"""
    ...
```

## What to Do

1. Add parameter type hints (e.g., \`data: str\`)
2. Add return type hints (e.g., \`-> None\`)
3. Apply to both \`validate_date_or_timestamp()\` and \`validate_option_offset()\`

## Example

```python
# After
def validate_date_or_timestamp(data: str) -> None:
    """Validates that the input string is either in 'YYYY-MM-DD' format or a numeric timestamp."""
    ...
```

## Skills You'll Learn

- Python type hints and annotations
- Static analysis benefits (mypy, IDE support)
- Writing self-documenting code

**Difficulty:** Easy | **Time:** 30 minutes | **Type:** Python

NavadeepDj:
### Summary
Added parameter and return type annotations to custom validation functions in `restx_api/data_schemas.py` and added a unit test suite to verify validation logic across all edge cases.

- `validate_date_or_timestamp(data: str) -> None`
- `validate_option_offset(data: str) -> bool`

### Issue Linked
Closes #893

### Testing
- `uv run ruff check restx_api/data_schemas.py test/test_data_schemas_validation.py` (Passed)
- `uv run pytest test/test_data_schemas_validation.py` (37 passed)

---
## Summary by cubic
Annotates validator functions in restx_api/data_schemas.py with type hints and adds tests to lock in behavior. This clarifies the contract (string inputs only) and prevents regressions without changing runtime behavior.

- validate_date_or_timestamp now takes str and returns None; it raises ValidationError for non-YYYY-MM-DD strings or non-length-valid numeric timestamps.
- validate_option_offset now takes str and returns bool; it raises ValidationError for invalid offsets and returns True for valid ATM/ITM1–ITM50/OTM1–OTM50 (case-insensitive).

<sup>Written for commit eb7dd60a2b2bd3525973c4da310609d7cce987ef. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1864?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/devsprint/participants.md
+++ b/docs/devsprint/participants.md
@@ -51,3 +51,4 @@ git push --force-with-lease
 | Rajandran R | [@marketcalls](https://github.com/marketcalls) | maintainer |
 |Padma Balaji L| [@PadmaBalajiL](https://github.com/PadmaBalajiL)|python|
 |Niranjan | [@cracker314](https://github.com/cracker314)|python|
+| Navadeep Marella | [@NavadeepDj](https://github.com/NavadeepDj) | python, backtesting |
--- a/restx_api/data_schemas.py
+++ b/restx_api/data_schemas.py
@@ -6,7 +6,7 @@ from utils.constants import VALID_EXCHANGES
 
 
 # Custom validator for date or timestamp string
-def validate_date_or_timestamp(data):
+def validate_date_or_timestamp(data: str) -> None:
     """
     Validates that the input string is either in 'YYYY-MM-DD' format or a numeric timestamp.
     """
@@ -19,7 +19,7 @@ def validate_date_or_timestamp(data):
 
 
 # Custom validator for option offset
-def validate_option_offset(data):
+def validate_option_offset(data: str) -> bool:
     """
     Validates option offset: ATM, ITM1-ITM50, OTM1-OTM50
     """
--- a/test/test_data_schemas_validation.py
+++ b/test/test_data_schemas_validation.py
@@ -1,0 +1,85 @@
+import pytest
+from marshmallow import ValidationError
+
+from restx_api.data_schemas import validate_date_or_timestamp, validate_option_offset
+
+
+class TestValidateDateOrTimestamp:
+    """Tests for validate_date_or_timestamp validation function."""
+
+    @pytest.mark.parametrize(
+        "valid_input",
+        [
+            "2026-08-22",
+            "1999-01-01",
+            "2030-12-31",
+            "1724300000",  # 10-digit epoch timestamp
+            "1724300000000",  # 13-digit millisecond timestamp
+        ],
+    )
+    def test_valid_date_or_timestamp(self, valid_input: str):
+        # Should execute without raising ValidationError
+        result = validate_date_or_timestamp(valid_input)
+        assert result is None
+
+    @pytest.mark.parametrize(
+        "invalid_input",
+        [
+            "2026/08/22",
+            "22-08-2026",
+            "2026-8-2",
+            "invalid_string",
+            "",
+            "12345",  # Too short for timestamp
+            "123456789012345",  # Too long for timestamp
+            None,
+            1724300000,  # Integer type instead of string
+        ],
+    )
+    def test_invalid_date_or_timestamp(self, invalid_input):
+        with pytest.raises(ValidationError):
+            validate_date_or_timestamp(invalid_input)
+
+
+class TestValidateOptionOffset:
+    """Tests for validate_option_offset validation function."""
+
+    @pytest.mark.parametrize(
+        "valid_offset",
+        [
+            "ATM",
+            "atm",
+            "ITM1",
+            "itm1",
+            "ITM25",
+            "itm25",
+            "ITM50",
+            "itm50",
+            "OTM1",
+            "otm1",
+            "OTM25",
+            "otm25",
+            "OTM50",
+            "otm50",
+        ],
+    )
+    def test_valid_option_offset(self, valid_offset: str):
+        assert validate_option_offset(valid_offset) is True
+
+    @pytest.mark.parametrize(
+        "invalid_offset",
+        [
+            "ITM0",
+            "ITM51",
+            "ITM100",
+            "OTM0",
+            "OTM51",
+            "OTM100",
+            "ATM1",
+            "INVALID",
+            "",
+        ],
+    )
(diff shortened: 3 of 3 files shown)


marketcalls:
Reviewed for security impact and for conformance with the conventions in CLAUDE.md. Merging.

## What was verified

Checked out `eb7dd60` locally and ran the change rather than reading it:

- `uv run pytest test/test_data_schemas_validation.py` - 37 passed, matching the claim in the description.
- `uv run ruff check restx_api/data_schemas.py test/test_data_schemas_validation.py` - clean.
- `ruff format --check` does flag `data_schemas.py`, but that is pre-existing on main at around line 293, nowhere near this diff. Not introduced here.
- All added lines are plain ASCII and within the configured `line-length = 100`.
- Both commits follow Conventional Commits (`refactor:`, `docs:`).

## Security

No impact. Annotations are inert metadata, so the regexes, the `isinstance` guard and the `ValidationError` paths are byte-identical. Validation strictness on the `from`/`to` fields of the history API and on the `offset` field of the option-symbol API is unchanged. No dependency changes, and both Snyk checks pass.

One thing worth recording for anyone reading this later. `validate_option_offset` calls `data.upper()` with no `isinstance` guard, so a direct call with a non-string raises `AttributeError` rather than `ValidationError`. That is not reachable through the API, because `fields.Str` rejects non-strings before validators run:

```
None      -> ValidationError {'offset': ['Field may not be null.']}
123       -> ValidationError {'offset': ['Not a valid string.']}
-- direct call, bypassing marshmallow:
None      -> AttributeError 'NoneType' object has no attribute 'upper'
```

So the `data: str` annotation states a real precondition that marshmallow already enforces at the boundary. Nothing to fix.

## Optional follow-ups, not blocking

None of these hold up the merge, but they are worth picking up in a later pass:

1. CLAUDE.md asks for Google-style docstrings, and roughly half the modules under `services/`, `database/`, `utils/` and `restx_api/` already carry `Args:` / `Returns:` / `Raises:` sections. Since the whole point of #893 is a self-documenting contract, filling those in on the two validators would finish the job the type hints start.
2. Most test modules in `test/` open with a docstring explaining why the test exists, usually citing the issue it covers. Worth adding here.
3. The test coverage is asymmetric on types: `validate_date_or_timestamp` is exercised with `None` and an `int`, `validate_option_offset` is not. Given the `AttributeError` behaviour above, that gap is deliberate-looking but undocumented.

Thanks for the clean first contribution, and for including the test suite rather than just the annotations. Closes #893.

marketcalls:
## Review metrics

Recording the numbers behind the review above, for the record and for anyone using this PR as a reference for what a good first contribution looks like.

### Mergeability: 93/100

| Dimension | Score | Note |
| --- | --- | --- |
| Correctness and risk | 20/20 | Annotations are inert. No runtime behaviour change, verified by running the change rather than reading it |
| Security | 20/20 | No attack surface. Non-string path confirmed unreachable through marshmallow |
| Standards conformance | 18/20 | Docked for the two docstring items listed as follow-ups |
| Test quality | 17/20 | 37 solid cases. Docked for the type-guard asymmetry between the two validators |
| Scope discipline | 13/15 | The participants row is a correctly prefixed standalone `docs:` commit rather than being folded into the code commit. The one-pull-request-per-person rule in `participants.md` is bent at PR level only |
| Style and conventions | 5/5 | Conventional Commits, ruff clean, plain ASCII throughout |

CI integration was assessed and then waived by the maintainer, so it is not scored. For context: `test/test_data_schemas_validation.py` is not in the explicit allowlist that the `backend-test` job runs in `ci.yml`, so these tests do not currently execute in CI. That is a deliberate accepted gap, not an oversight in this PR.

### Conformance against CLAUDE.md

| Convention | Result |
| --- | --- |
| Always use uv | Pass. `uv run ruff`, `uv run pytest` |
| Ruff style, line-length 100, 4 spaces | Pass. All added lines under 100 |
| No icons or emojis anywhere | Pass. Verified at byte level, every added line is plain ASCII |
| Conventional Commits | Pass. `refactor:` and `docs:` |
| Docs under `docs/` as single source of truth | Pass. Edits `participants.md` in place, no duplication |
| Tests in `test/`, named `test_*.py` | Pass. Matches the `pyproject.toml` testpaths config |
| ORM not raw SQL, migration for schema changes | Not applicable, no database surface touched |
| FD hygiene | Not triggered. The diff opens no file, socket, thread or engine |
| Frontend dist handling | Not applicable, no frontend changes and no stray dist commits |

### Measurements taken

- 37 of 37 tests pass on `eb7dd60`.
- 0 ruff lint findings on both changed files. The `ruff format` complaint on `data_schemas.py` is pre-existing on main near line 293 and is not introduced here.
- 0 non-ASCII characters in added lines.
- 0 added lines exceeding the configured `line-length = 100`.
- 0 dependency manifest changes. Both Snyk checks pass.
- Google-style docstrings appear in 100 of 198 modules across `services/`, `database/`, `utils/` and `restx_api/`, which is why follow-up 1 is worth doing.
- 82 of 95 files in `test/` open with a module docstring, which is why follow-up 2 is worth doing.

Neither docstring item blocked the merge. Both are small enough to fold into a future pass.

PR #1865, opened 2026-08-22
fix: prevent caching API key responses

### LINKED ISSUE #1850: security: Add no-store headers to API-key responses
## Summary

Authenticated endpoints return decrypted or newly generated API keys without explicitly preventing browser and intermediary caching.

## Evidence

- `blueprints/apikey.py:45-99` returns an existing or newly generated key.
- `blueprints/playground.py:301-308` returns the current key.
- `blueprints/websocket_example.py:167-183` returns a key for WebSocket authentication.
- Admin endpoints already set `Cache-Control: no-store, max-age=0`, providing a local pattern.

## Why this matters

Credential-bearing responses should not be stored in browser history caches, shared proxies, or debugging caches longer than necessary.

## Scope

Add one reusable response helper or after-response policy for the credential-returning endpoints. Include `Cache-Control: no-store, max-age=0` and a compatible legacy `Pragma: no-cache` header without changing JSON bodies.

## Acceptance criteria

- [ ] Every endpoint that returns a decrypted/new API key sends no-store headers.
- [ ] Non-sensitive endpoints are not accidentally changed.
- [ ] GET and POST paths in `/apikey` are covered.
- [ ] Tests assert headers and never use real credentials.

## Verification

```powershell
uv run pytest test/test_api_key_cache_headers.py -v
uv run ruff check blueprints/apikey.py blueprints/playground.py blueprints/websocket_example.py test/test_api_key_cache_headers.py
```

## Security guidance

- Difficulty: Easy
- Estimated time: 1-3 hours
- You will learn: HTTP cache controls, reusable Flask responses, and security regression tests.

Xayar145:
## Summary

Adds explicit no-store headers to every response that returns a decrypted or newly generated API key.

## Changes

- Add a reusable `make_no_store_response` Flask helper.
- Apply it to the JSON, credential-bearing HTML fallback, and successful POST paths for `/apikey`.
- Apply it to `/playground/api-key` and `/api/websocket/apikey`.
- Add regression coverage for all credential-bearing paths.
- Verify that the static React shell and the non-sensitive `/apikey/mode` response remain unchanged.
- Preserve all existing response bodies and status codes.

## Tests

- `python -m pytest -p no:cacheprovider test/test_api_key_cache_headers.py -q` — 7 passed
- `ruff format --no-cache --check blueprints/apikey.py blueprints/playground.py blueprints/websocket_example.py utils/response.py test/test_api_key_cache_headers.py` — passed
- `ruff check --no-cache --ignore F841 blueprints/apikey.py blueprints/playground.py blueprints/websocket_example.py utils/response.py test/test_api_key_cache_headers.py` — passed
- `detect-secrets-hook --no-verify --baseline .secrets.baseline ...` — passed
- Broader suite: 1,313 passed and 38 skipped; remaining failures/errors are existing environment/integration tests requiring external services, fixtures, or async plugins.

## Notes

The exact unignored Ruff command also reports two pre-existing `F841` violations in `blueprints/websocket_example.py` at the example-only assignments around lines 452 and 458; this PR does not modify those lines.

Tests use only explicit non-real values such as `test-token-not-real`.

Closes #1850

---
## Summary by cubic
Prevents caching of API-key responses by adding explicit no-store headers wherever credentials are returned. Previously, some API-key endpoints were cacheable; now those responses send Cache-Control: no-store, max-age=0 and Pragma: no-cache, while non-sensitive endpoints keep prior caching behavior.

- Introduces `utils/response.make_no_store_response`.
- Applies to `/apikey` JSON, the legacy HTML fallback, successful `POST /apikey`, `/playground/api-key`, and `/api/websocket/apikey`.
- Leaves the static React shell and `/apikey/mode` responses unchanged.
- Preserves all response bodies and status codes.
- Adds regression tests covering credential-bearing and non-sensitive paths.

<sup>Written for commit b90b9474d0e77a5ffe42cdda4c40103085442265. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1865?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/blueprints/apikey.py
+++ b/blueprints/apikey.py
@@ -23,6 +23,7 @@ from database.auth_db import (
     verify_api_key,
 )
 from utils.logging import get_logger
+from utils.response import make_no_store_response
 from utils.session import check_session_validity
 
 # Path to React frontend
@@ -56,27 +57,31 @@ def manage_api_key():
 
         # Return JSON if Accept header requests it (for React frontend)
         if request.headers.get("Accept") == "application/json":
-            return jsonify(
-                {
-                    "login_username": login_username,
-                    "has_api_key": has_api_key,
-                    "api_key": api_key,
-                    "order_mode": order_mode,
-                }
+            return make_no_store_response(
+                jsonify(
+                    {
+                        "login_username": login_username,
+                        "has_api_key": has_api_key,
+                        "api_key": api_key,
+                        "order_mode": order_mode,
+                    }
+                )
             )
 
         # Serve React app for browser navigation
         index_path = FRONTEND_DIST / "index.html"
         if index_path.exists():
-            return send_file(index_path, mimetype="text/html")
+            return make_no_store_response(send_file(index_path, mimetype="text/html"))
 
         # Fallback to old template if React build not available
-        return render_template(
-            "apikey.html",
-            login_username=login_username,
-            has_api_key=has_api_key,
-            api_key=api_key,
-            order_mode=order_mode,
+        return make_no_store_response(
+            render_template(
+                "apikey.html",
+                login_username=login_username,
+                has_api_key=has_api_key,
+                api_key=api_key,
+                order_mode=order_mode,
+            )
         )
     else:
         user_id = request.json.get("user_id")
@@ -92,8 +97,14 @@ def manage_api_key():
 
         if key_id is not None:
             logger.info(f"API key updated successfully for user: {user_id}")
-            return jsonify(
-                {"message": "API key updated successfully.", "api_key": api_key, "key_id": key_id}
+            return make_no_store_response(
+                jsonify(
+                    {
+                        "message": "API key updated successfully.",
+                        "api_key": api_key,
+                        "key_id": key_id,
+                    }
+                )
             )
         else:
             logger.error(f"Failed to update API key for user: {user_id}")
--- a/blueprints/apikey.py
+++ b/blueprints/apikey.py
@@ -71,7 +71,7 @@ def manage_api_key():
         # Serve React app for browser navigation
         index_path = FRONTEND_DIST / "index.html"
         if index_path.exists():
-            return make_no_store_response(send_file(index_path, mimetype="text/html"))
+            return send_file(index_path, mimetype="text/html")
 
         # Fallback to old template if React build not available
         return make_no_store_response(
--- a/blueprints/playground.py
+++ b/blueprints/playground.py
@@ -8,6 +8,7 @@ from flask import Blueprint, current_app, jsonify, redirect, session, url_for
 
 from database.auth_db import get_api_key_for_tradingview
 from utils.logging import get_logger
+from utils.response import make_no_store_response
 from utils.session import check_session_validity
 
 logger = get_logger(__name__)
@@ -305,7 +306,7 @@ def get_api_key():
         return jsonify({"error": "Not authenticated"}), 401
 
     api_key = get_api_key_for_tradingview(login_username)
-    return jsonify({"api_key": api_key or ""})
+    return make_no_store_response(jsonify({"api_key": api_key or ""}))
 
 
 @playground_bp.route("/collections")
--- a/blueprints/websocket_example.py
+++ b/blueprints/websocket_example.py
@@ -25,6 +25,7 @@ from services.websocket_service import (
     unsubscribe_from_symbols,
 )
 from utils.logging import get_logger
+from utils.response import make_no_store_response
 from utils.session import check_session_validity
 
 # Initialize logger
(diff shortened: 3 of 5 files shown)


marketcalls:
Thanks for the patch. The bug is real and I reproduced it: on current main, `GET /apikey` with `Accept: application/json` comes back with only `Content-Type`, `Content-Length` and `Vary: Cookie`, no `Cache-Control` at all, so the plaintext key is eligible for the browser disk cache. The helper is correct, `ruff check` and `ruff format --check` are clean on all five files (the two F841 hits are pre-existing at `blueprints/websocket_example.py` lines 452 and 458 and you correctly left them alone), and your tests genuinely gate the behaviour: I get 7 passed with the patch and 5 failed, 2 passed when I revert only the three blueprint files.

Four things need to change before this goes in.

1. `blueprints/auth.py` line 1052. `GET /auth/session-status` returns `"api_key": api_key` in its JSON body and is untouched by this PR. That is the endpoint the SPA actually gets the key from: `frontend/src/components/auth/AuthSync.tsx` line 26 fetches it on every mount and line 45 calls `setApiKey(data.api_key)`, and `frontend/src/pages/Login.tsx` line 49 fetches it too. As written, the most frequently hit credential-bearing GET in the app is the one left cacheable, so "every response that returns a decrypted or newly generated API key" is not accurate. Wrap that branch and add a regression test for it. While you are in the file, `blueprints/auth.py` line 140 (`GET /auth/broker-config`) returns `broker_api_key` for authenticated users and is fetched by `frontend/src/pages/BrokerSelect.tsx` line 92, so state explicitly whether it is in scope.

2. `.github/workflows/ci.yml` lines 43 to 52. `test/test_api_key_cache_headers.py` is not in the CI-safe pytest file list, so it will never run and nothing stops a later refactor from dropping `make_no_store_response`. Add the file there. This matters more than usual on this PR: CI has not run on the branch at all, because fork PRs wait on maintainer approval. The only checks on the page are Snyk reporting no manifest changes and a NEUTRAL cubic run, so treat the branch as unvalidated.

3. `blueprints/apikey.py` lines 74 to 80. The `render_template("apikey.html", ...)` fallback is dead. Commit c81dfb417 removed all 77 Jinja2 templates and there is no `templates/` directory in the repo, so if that branch is reached it raises `TemplateNotFound` and returns 500 rather than a page. Wrapping it adds nothing, and `test_apikey_html_response_prevents_caching_credentials` only passes because line 202 monkeypatches `render_template` into existence. Delete the dead fallback outright (I am happy to take that in this PR) or leave it untouched and drop that test. Do not ship a test that asserts a header on a branch which cannot succeed.

4. `utils/response.py` line 18. Drop `Pragma: no-cache`. `Pragma` is defined as a request header and its use on a response is unspecified, and `blueprints/admin.py` already sets `Cache-Control: no-store, max-age=0` on its own at lines 844, 991, 1102, 1451, 1646, 1871 and 2158. One spelling for one intent. A follow-up PR that migrates those seven admin.py sites onto the new helper would be welcome, but keep it out of this one.

Two further notes, neither of which blocks the change.

`frontend/src/stores/authStore.ts` wraps the store in zustand `persist` with `{ name: 'openalgo-auth' }` and no `partialize`, which means the default localStorage backend, and the store holds `apiKey`. The same plaintext key you are keeping out of the HTTP cache is therefore already written to `localStorage` on every session sync. On a shared machine that is the larger at-rest exposure and no-store headers do nothing about it. Please soften the PR description accordingly. A separate PR adding `partialize` so `apiKey` is never persisted would be the more valuable half of this work, and I would review it quickly.

`test/test_api_key_cache_headers.py` line 9 imports `blueprints.websocket_example`, which transitively runs `_market_data_service = MarketDataService()` at `services/market_data_service.py` line 1058 and starts a daemon cleanup thread at lines 376 to 377. That is a heavy import-time side effect for a test that asserts one response header, and it is worth avoiding before the file joins the CI list.

Separately, the Tests section of the PR description uses em dashes. CLAUDE.md forbids em dashes and en dashes in any generated text, PR bodies included. A comma or a colon is fine. The diff itself is clean ASCII.

Nothing here touches the eventlet, ZeroMQ, NullPool, FD-hygiene or logging invariants, and there is no schema change, so no migration script under `upgrade/` is needed.

marketcalls (commented):
Reviewed head `b90b9474d0e77a5ffe42cdda4c40103085442265`: 89/100. This is a sound scoped implementation, and all 7 focused tests passed. It duplicates #1915, which covers the same credential-bearing paths and wires its regression suite into CI. Consolidate on #1915 instead of merging both helpers/test files. If this PR is retained instead, add its regression file to the explicit CI list and close the competing alternative. Also narrow the description: /auth/session-status still returns api_key without no-store; this existing route is outside the patch, so the claim that every API-key response is covered is inaccurate.

Issue #1591, opened 2026-06-30
issues in multi Sessions login


tradewithmadhan:
Hi @marketcalls  
I have noticed below issues in  multi Sessions login.

- [ ] 1. In flattrade when i login 2nd session immediately broker session getting expired . earlier it was working  , this issue i noticed in recent updates .
- [ ] 2. In shooniya broker after 2nd session login existing websocket data stopped . only after page refresh it started working .

Please fix the above issues asap..

marketcalls:
Thanks for the report @tradewithmadhan. Both points turned out to be the same root cause, now fixed in 3aadeb33 on `main`.

**Root cause**

OpenAlgo is single-user / single-broker per instance, but the same user can be logged in from multiple devices, and all of them share **one** server-side broker WebSocket feed (a single connection pool per `broker_user`, fanned out to every browser by the proxy).

When the 2nd device logs in, it resumes the existing broker session and re-saves the **same** token via `upsert_auth()`. That function was *unconditionally* tearing the shared feed down on every call — it published a ZMQ cache-invalidation (which makes the out-of-process WebSocket proxy disconnect the broker adapter and its pool) and also cleaned up the pool in-process. So a harmless 2nd-device login ripped down the live feed the 1st device was using:

- **Shoonya** - ticks stop until a page refresh re-subscribes and rebuilds the pool.
- **Flattrade** - the Finvasia/Noren backend allows a single active session, so the abrupt disconnect/reconnect makes the broker drop the token, surfacing as "broker session expired".

That teardown was added for a different case (a genuine re-login with a **new** token must rebuild the pool - issues #1394 / #765 / #851). It was correct there; it was just firing even when nothing changed.

**Fix**

`upsert_auth()` now gates the teardown on an actual session change. It compares the new token / feed token / broker / revoke flag against the stored row (using decrypted plaintext, since Fernet ciphertext is non-deterministic). If nothing material changed (a multi-device resume), it clears only the cheap in-process caches and returns early, leaving the live feed up. A real change (new login, daily ~3 AM token rollover, logout/revoke) still runs the full invalidation + pool rebuild, so the #1394 behaviour is preserved.

Net effect: multiple devices/sessions now stream simultaneously off the one shared broker feed, which is the intended design. Added regression tests in `test/test_auth_upsert_multisession.py`.

[...]

Issue #765, opened 2026-01-26
bug: Process-local auth_cache causes 403 Forbidden in WebSocket proxy after re-login


aravindgandavadi:
**Describe the bug**
The authentication cache (`auth_cache` in `database/auth_db.py`) is implemented using `cachetools.TTLCache`, which is an in-memory, process-local cache.

When the application runs with multiple processes (e.g., Flask web server and WebSocket proxy server running separately), they maintain separate instances of this cache.

**Scenario leading to failure:**
1. **WebSocket Proxy** fetches and caches the user's auth token.
2. User performs a **re-login** via the Flask web interface.
3. The Flask process updates the token in the database and updates its *own* local `auth_cache`.
4. However, the **WebSocket Proxy** process is unaware of this change and continues to hold the *stale* (old) token in its cache.
5. When the user tries to connect or interact via WebSocket, the proxy uses the stale token for backend validation.
6. The backend (or broker) rejects the stale token, resulting in **HTTP 403 Forbidden** errors.

**Relevant Code**
In `database/auth_db.py`:
```python
# Define auth token cache with TTL until session expiry
auth_cache = TTLCache(maxsize=1024, ttl=get_session_based_cache_ttl())
```
The `upsert_auth` function updates the cache only for the current process:
```python
def upsert_auth(...):
    ...
    # Only updates the local process memory
    auth_cache[cache_key_auth] = auth_obj
```

**Expected behavior**
Authentication token updates should be propagated to all running processes, or the cache should be shared/invalidated globally to ensure the WebSocket proxy always uses the latest credentials.

**Possible Solutions**
1. **Shared Cache**: Use a shared cache backend like Redis or Memcached (adds infrastructure dependency).
2. **Database Backup Check**: If authentication fails with the cached token, retry by fetching fresh data from the DB.
3. **Disable Caching for WS**: Disable `auth_cache` specifically for the WebSocket proxy if query performance allows.
4. **Inter-Process Communication**: Clear cache via an event (e.g., using the existing ZeroMQ bus) when a token is updated.

**Environment**
- OS: Ubuntu/Linux
- Deployment: Docker / Gunicorn / Separate WS Proxy script

marketcalls:
Will test in a live markets tomorrow and will rollout a fix

marketcalls:
The Problem                                                                                               
                                                                                                            
  When you run OpenAlgo in Docker, two separate Python processes run:                                       
  1. Flask web server (handles login/logout)                                                                
[...]

Issue #1419, opened 2026-05-14
Issue: Broker WebSocket does not reconnect after new trading day login unless container is restarted


aravindgandavadi:
After logging into OpenAlgo on a new trading day, OpenAlgo is unable to establish a WebSocket connection with the broker (Zerodha in my case). The issue persists until I manually restart the container instance either through Portainer or using Docker commands.

Once the container is restarted, OpenAlgo successfully establishes the broker WebSocket connection and works normally.

This issue appears consistently on a new trading day.

**Environement**: Azure VM/Ubuntu/Docker/Zerodha
Installation via: [install-docker-multi-custom-ssl.sh](https://github.com/marketcalls/openalgo/blob/main/install/install-docker-multi-custom-ssl.sh)
**Commit**: 4e09da8b

**Additional Observation**

The current Diagnostics page only checks whether OpenAlgo’s own WebSocket service is running.

However, it does not verify:
- Whether OpenAlgo has successfully connected to the broker WebSocket
- Whether the broker WebSocket session is authenticated and active
- Whether live ticks/data are actually flowing from the broker

Enhancement Request

Please enhance the Diagnostics page to include:

- Broker WebSocket connection status
- Authentication/token validity status
- Last successful broker WebSocket connection timestamp
- Whether live market data/ticks are being received
- Automatic detection/reporting of stale or disconnected broker WebSocket sessions

This would make troubleshooting much easier and help identify issues without needing to inspect logs manually.

I am attaching the log file here (last 1000 lines):

[_openalgo-zerodha-valmiki-trade-web_logs-10.txt](https://github.com/user-attachments/files/27740253/_openalgo-zerodha-valmiki-trade-web_logs-10.txt)

Here is the diagnostics file from my OpenAlgo instance: 

[openalgo-system-report-20260514-082417.txt](https://github.com/user-attachments/files/27740266/openalgo-system-report-20260514-082417.txt)

internhutone:
facing same issue from the day i started using openalgo 
i have being using Fyers broker and using Oracle instance

marketcalls:
Thanks @aravindgandavadi for the detailed report and for attaching both the container log and the diagnostics report — that combination was enough to nail down the root cause. Same applies to @internhutone on Fyers / Oracle — this is broker-agnostic, not Zerodha-specific.

## Root cause

Race condition in the existing cache-invalidation-on-login path:

[...]

PR #1866, opened 2026-08-22
fix(auth): reject broker credentials from a previous trading session

### LINKED ISSUE #1858: backend: Stop stale broker REST calls after the 3 AM IST rollover
## Summary

After the configured daily session boundary (03:00 IST by default), OpenAlgo can continue sending quote and history requests with the previous trading session's broker credential. Broker-specific failures such as Zerodha's `Incorrect api_key or access_token` are then surfaced as repeated HTTP 500 errors and stack traces until an unrelated browser route finally runs session cleanup.

This must be handled in shared authentication/service code so every daily-expiring broker gets the same behavior.

## Reported reproduction

1. Authenticate a daily-expiring broker before 03:00 IST.
2. Leave OpenAlgo, its WebSocket feed, and an API/dashboard client running across `SESSION_EXPIRY_TIME`.
3. Continue calling quote or history endpoints after the boundary.
4. Observe repeated broker credential errors. In the supplied Zerodha log, failures occurred from 06:09 IST, while OpenAlgo logged `Session expired at 2026-08-22 03:00:00+05:30 IST` and performed cache/pool teardown only at 06:10:37.

Do not include the reporter's username, filesystem path, API key, or broker access token in tests or comments.

## Evidence in the current code

- `app.py:500-529` performs expiry cleanup only in a `before_request` hook and explicitly skips `/api/` plus `/auth/session-status`.
- `utils/session.py:83-89` can identify a pre-boundary login after the boundary, but `revoke_user_tokens()` is request-session scoped.
- `services/quotes_service.py:144-154` converts any broker exception into a 500 response.
- `services/history_service.py:139-141` does the same for history calls.
- `blueprints/auth.py:1007-1035` treats a stored token as connected without checking whether its login belongs to the current trading session.
- `services/order_update_service.py:240-246` already documents the same gap: the auth row remains non-revoked until a browser request fires the expiry hook.

## Root-cause hypothesis

The stored credential and its caches have no broker-neutral request-time freshness gate. API routes bypass the Flask-session expiry hook, so a token from the previous trading session remains usable by shared REST services even though the broker has invalidated it.

## Scope

Add a request-independent rollover guard at the shared broker-credential resolution boundary. For non-crypto brokers, it should reject a credential when there has been no successful login in the current trading session, run one idempotent invalidation/teardown transition, and return a machine-readable broker-reconnect result before any broker adapter is called.

Use the existing `has_login_this_trading_session()` check, auth/cache invalidation infrastructure, and `BROKER_SESSION_EXPIRED` response convention where possible. Do not add a new APScheduler/background-thread job; #1441 tracks an eventlet incompatibility in that approach.

## Acceptance criteria

- [ ] API-key-authenticated quote and history calls made after the rollover do not call a broker with the previous session's credential.
- [ ] The shared path returns HTTP 401 with `code: BROKER_SESSION_EXPIRED` and a broker-reconnect message, not HTTP 500.
- [ ] `/auth/session-status` or the dashboard detects the same stale broker state and keeps the app session usable for broker reconnection.
- [ ] The first stale-credential detection revokes/invalidates auth and feed caches and triggers the existing WebSocket/order-adapter teardown across processes.
- [ ] Repeated calls are idempotent: they do not contact the broker, repeat full exception tracebacks, or publish teardown storms.
- [ ] `DISABLE_SESSION_EXPIRY=true` keeps the existing 24/7 crypto behavior.
- [ ] Tests freeze time before, exactly at, and after the configured boundary and need no broker credentials or network access.
- [ ] No API key, access token, feed token, or token prefix is written to logs.

## Verification

```powershell
uv run pytest test/test_broker_session_rollover_guard.py test/test_session_expiry.py test/test_order_adapter_auth_rejection.py -v
uv run ruff check database/auth_db.py utils/session.py services/quotes_service.py services/history_service.py blueprints/auth.py test/test_broker_session_rollover_guard.py
```

## Related issues

- #1400 added the reconnect UX and machine-readable expired-session behavior.
- #1419 covered stale WebSocket state after a new trading-day login.
- #1371 covered Zerodha WebSocket auth-failure retry behavior.
- #1441 is an open scheduler/eventlet failure and is a constraint on the implementation here.

- Difficulty: Medium
- Estimated time: 4-8 hours
- You will learn: timezone boundary testing, shared authentication contracts, Flask service errors, cache invalidation, and broker-neutral regression testing.
- Suggested first step: write the frozen-time test proving that a valid OpenAlgo API key plus a previous-session broker login currently reaches the mocked broker adapter.

Pragitics:
## What this does

Stops OpenAlgo sending a previous trading session's broker credential on the API path, and makes every endpoint report an unresolved credential honestly.

`app.py`'s expiry sweep runs from a `before_request` hook that skips `/api/` and reads a Flask cookie an API-key caller does not have, so an API-authenticated client kept using yesterday's credential until an unrelated browser request happened to fire the hook. In the reported case that gap ran from 03:00 to 06:10, surfacing as repeated HTTP 500s.

## Why

Closes #1858

## The two problems

`database.auth_db.get_auth_token_broker()` answers `(None, None)` for four different situations, and the tuple cannot say which:

```
bad API key                -> (None, None)
token revoked (logout)     -> (None, None)
broker never connected     -> (None, None)
session rolled over (#1858)-> (None, None)
```

Only the first is a problem with the API key. Every consumer reported all four as `403 "Invalid openalgo apikey"`, so after 03:00 every morning the platform told operators to regenerate a key that was fine. Regenerating cannot help, because the verdict never came from the key, and it overwrites the working key on the way past.

Separately, several endpoints ask for a credential they never use, so withholding one takes them down for no reason.

### One decision, inherited by every caller

`utils/credential_errors.credential_error()` answers it once, from `verify_api_key` alone:

| Situation | Answer |
| --- | --- |
| API key not recognised | 403 `Invalid openalgo apikey` |
| key valid, credential missing for any reason | 401 `BROKER_SESSION_EXPIRED` |

31 modules across `services/` and `restx_api/` now route through it, and the hand-rolled 401 branches from the earlier round are gone.

Deriving the answer from the API key rather than from `active_sessions` is also what makes it stable. An earlier revision asked the freshness inference here, which answers unknown for an empty `active_sessions`; once the expiry sweep cleared those rows the honest 401 silently reverted to the misleading 403, while the new documentation still told the reader to re-issue their key. That is fixed by construction rather than by another branch.

### Endpoints that never call a broker resolve identity, not a credential

`ping`, `symbol`, `intervals` and the two analyzer endpoints now use `verify_api_key` plus `get_broker_name`. None of them touch the auth token: `ping_with_auth` returns pong and ignores it, `symbol` reads the local `SymToken` table, `intervals` reads a per-broker map through `import_broker_module(broker)`, and the analyzer drives the sandbox. `/api/v1/ping` is the canonical health check, so gating it turned the daily rollover into a monitoring outage.

`intervals` needed a second fix you caught: exempting it replaced the token with `get_broker_name(api_key)`, which answers `None` for a revoked or absent Auth row, so `import_broker_module(None)` looked for `broker.None.api.data` and returned 404 plus an ERROR log line per call, where main returned 403. It now guards the `None` and answers `credential_error(api_key)`.

### Sandbox is no longer coupled to a live broker session

This is the item you asked me to decide explicitly rather than leave as an undeclared partial. I have applied the early return to all of it, not just `placeorder`.

Thirteen entry points now check the analyze branch before resolving a credential:

```
place_order        place_smart_order   split_order      basket_order
modify_order       cancel_order        cancel_all_order close_position
orderbook          positionbook        tradebook        holdings
funds
```

The analyze branch lives inside each `*_with_auth` function, so resolving first meant a rolled-over broker session rejected a sandbox operation that never reaches a broker. CLAUDE.md documents the sandbox engine as fully isolated from live trading.

### A failed session write is undecidable, not stale

`handle_auth_success` ignores `register_session`'s return value, so a failed write leaves today's live token beside the previous session's rows: readable logins that all predate the boundary, which is a confidently wrong stale verdict. `register_session` now marks the user, and the freshness verdict answers unknown until the next successful login clears the mark.

### The freshness verdict itself

`utils/session.broker_session_freshness()` returns one of three values, and only a confirmed stale reading withholds anything:

| `active_sessions` says | Verdict | Effect |
| --- | --- | --- |
| a login at or after the current session boundary | `fresh` | credential resolves |
| readable logins exist, all predate a boundary that has been crossed | `stale` | credential withheld |
| no rows, no parseable `login_time`, a failed read, or a failed session write | `unknown` | credential resolves, as before |

`_current_session_boundary()` is the underlying bug fix: `_todays_rollover_boundary()` returns today's clock time unconditionally, which between midnight and `SESSION_EXPIRY_TIME` is a future instant, so every healthy overnight session compared as stale. MCX runs to 23:30 IST, so that window holds live positions. This is the same two-sided reasoning `is_session_valid()` already applies:

```python
if now_ist > daily_expiry and login_time < daily_expiry:
```

It is a behaviour change for the two existing read-only consumers, not a no-op: `services/order_update_service.py` will start the order-update adapter during the midnight to 03:00 window where it previously refused, and `/auth/session-status` stops reporting a live session as expired there.

`has_login_this_trading_session()` now also short-circuits on `is_session_expiry_disabled()`, which its sibling already did. Without it a 24/7 Delta Exchange instance whose last login was yesterday read stale and got no order-update adapter at boot.

### Withhold, never destroy

The guard returns `(None, None)` and does not touch the row. Revoking would reach `upsert_auth(username, "", "", revoke=True)`, which blanks the token and broker irreversibly, and `_try_resume_broker_session` then refuses to resume. It is also unnecessary: that function already validates the stored token against the live broker before accepting it, so a dead token cannot be resumed.

## Two deliberate deviations

I would rather name these than have them found.

### 1. Acceptance criterion 4 of #1858 is not implemented

The issue asks that the first stale detection revoke auth and feed caches and trigger the WebSocket and order-adapter teardown across processes. Your first review called that path credential destruction and asked for it to go, and I agree with you rather than with the issue text. The guard now withholds and destroys nothing. The issue's criterion was never updated to match; flagging it here so the unchecked box is not read as an oversight.

### 2. `get_auth_token` and `get_feed_token` on the streaming path stay unguarded

You listed these under "worth fixing while you are here". I have not, and the reason is blast radius rather than effort.

The verdict is an inference over `active_sessions` rows, which is exactly why `SESSION_UNKNOWN` exists. A wrong inference costs very different amounts on the two paths:

```
false stale on REST    -> one call answers 401, the next call re-reads, self-heals
false stale on the feed -> the pooled adapter drops, and one pool serves every
                           device (_POOLED_ADAPTERS), so all of them lose market data
```

Under "what I verified holds" you wrote that a false verdict cannot touch the shared feed because the proxy never calls `get_auth_token_broker`. Guarding these two readers is precisely what would remove that containment, and CLAUDE.md's multi-session invariant exists because #1591 was that failure.

The mechanism also differs. REST re-resolves the credential on every call, which is the loop that produced the reported 500s. The streaming path resolves once at connect, and it already has two pre-existing guards of its own: `services/order_update_service.py:247` refuses to start an adapter at boot without a login this trading session, and `websocket_proxy/base_adapter.py` reacts to a 403 by re-reading the token with the cache bypassed (#765).

I have not watched either of those fire against a live broker, so I am claiming the mechanism differs and the reactive path exists, not that streaming handles the rollover correctly. Happy to be overruled, and happy to do it as a follow-up if you want the symmetry.

### Where the dashboard asks the question

`/auth/dashboard-data` now asks `is_broker_session_stale_for_user` at the call site instead of inside `get_auth_token`, for the same reason. That keeps the containment above while closing the gap you identified, where `Dashboard.tsx:84` reads `BROKER_SESSION_EXPIRED` from an endpoint that resolved through the unguarded reader and so could never emit it. Analyze mode is exempt, on the same logic as the sandbox branch.

## On splitting

You recommended splitting `_current_session_boundary()` out and landing it today, and offered it twice. I have kept it as one PR, which is the option you left open, and the three items from your last review are all in.

I want to put the decision back to you once, now that the work is finished, because the ground has moved. When I declined the split this was 13 files. Completing the consumer sweep you asked for took it to 49. If you would still rather have two PRs, say so and I will carve it; the boundary fix is self-contained and everything else rebases onto it.

The diff is 49 files, +2136/-195. Most of it is the same one-line call-site change repeated.

## Changes

| File | Change |
| --- | --- |
| `utils/credential_errors.py` | New. The single decision |
| 33 `services/*.py` | Route the failure through it; 13 also gained the sandbox early return |
| `restx_api/{ticker,portfolio,sip}.py` | Same answer on the resolve-failure branch |
| `services/{ping,symbol,intervals,analyzer}_service.py` | Resolve identity, not a credential |
| `utils/session.py` | `_current_session_boundary()`, `broker_session_freshness()`, crypto bypass |
| `database/auth_db.py` | The guard, and the failed-session-write marker |
| `blueprints/auth.py` | `/auth/session-status` and `/auth/dashboard-data` check freshness |
| `docs/api/market-data/{quotes,multiquotes,history,depth}.md` | Error-response tables |
| 3 new test files | 80 tests |
| `.github/workflows/ci.yml` | Registers all three |

`restx_api/portfolio.py` and `sip.py` keep their 401 branches written out rather than calling `credential_error()`. They resolve a three-tuple with a feed token, so the call shape differs; what matters is that they now carry `code: BROKER_SESSION_EXPIRED` instead of answering 403, and the static test asserts the marker rather than the helper so either spelling passes.

## How I tested

```
test/test_broker_session_rollover_guard.py     25 passed
test/test_session_status_broker_expiry.py       7 passed
test/test_credential_error_uniformity.py       48 passed

the CI-safe list, verbatim from ci.yml         152 passed
```

**Full suite, diffed against a fresh baseline on current main.** Both runs exclude `test_bot_web.py` and `test_telegram_startup.py`, which fail collection on both sides.

```
main    : 11 failed, 9 errors, 1826 passed
branch  :  9 failed, 9 errors, 1936 passed

new failures on branch: none
```

The branch's 9 failures are a strict subset of main's 11: mstock x4 (needs live broker credentials), telegram_bot, telegram_charts and websocket x3 (`async def` under a non-async pytest). Main additionally failed `test_email_functionality` and `test_telegram_config` on my machine, which are environment-dependent and not something this branch touches.

**Every fix reverted individually and confirmed red**, per fix rather than per file. The last three are the items from your most recent review:

| Reverted | Result |
| --- | --- |
| one service back to the old 403 | 2 failed |
| `ping` re-gated on a credential | 2 failed |
| the failed-session-write marker removed | 1 failed |
| `_current_session_boundary` back to `_todays_rollover_boundary` | 1 failed |
| the day-subtraction removed inside the helper | 2 failed |
| `unknown` collapsed into `stale` | 4 failed |
| the depth `len()` check restored | 2 failed |
| the log throttle removed | 1 failed |
| `service_key()` back to a raw path membership test | 2 failed |
| the `intervals` `None` guard removed | 1 failed, reproducing `broker.None.api.data` |
| the sandbox early return removed from one service | 2 failed |

`test_no_service_hand_rolls_the_credential_answer` is the one that matters over time: it fails for a service added later that resolves a credential and copies the old two-case shape.

No new ruff errors anywhere in the repo compared to main. `ruff format` was run only over lines this branch added in its own two new test files, so pre-existing formatting elsewhere is untouched and the diff stays readable. No non-ASCII in any added line.

### What I have not verified

- No live broker was involved at any point. Every claim about broker behaviour comes from reading the code and from the issue report.
- The pre-boundary window is exercised by monkeypatching the boundary helper, not by running the suite at 01:00 IST.
- I have not observed the streaming path's boot gate or its 403 re-read fire in production; see deviation 2 above.
- `test_session_expiry.py` and `test_order_adapter_auth_rejection.py` are named in #1858's verification block and both pass, but neither is registered in `ci.yml`. That gap predates this branch and I have left it alone rather than widen the diff further. Say the word and I will open a follow-up.
- The `frontend-test` failure from an earlier round is not this change and now passes on this branch with no frontend file modified. The root cause is real: the tests use `screen.getByRole`, which is synchronous and throws on the first miss, so the slow timeout added in `2d8edfd0e` could never have fixed it. `findByRole` retries. Raised separately as #1903 rather than folded in here.

## Checklist

- [x] Conventional commit messages
- [x] Existing tests pass unmodified
- [x] New tests need no broker credentials or network
- [x] All three new test files registered in the CI-safe list
- [x] No API key, access token, feed token, or token prefix written to logs
- [x] `DISABLE_SESSION_EXPIRY=true` behaviour preserved, on both sides now
- [x] No APScheduler or background-thread job added
- [x] No schema change, so no migration
- [x] fd-audit run: bounded `TTLCache`es only, no new descriptors
- [x] ASCII only, in code and in this description


marketcalls:
Thanks for the detailed write-up. The diagnosis is correct and well evidenced: `app.py:510` really does skip `/api/`, `check_session_expiry` (`app.py:501`) really does read the Flask cookie an API-key caller never sends, and `services/order_update_service.py:247` already applies the same rule with a comment describing this exact gap. Gating the shared resolution point is the right place to fix it, and reusing the existing helpers instead of adding a scheduler is the right instinct.

The problem is what the guard does when it fires. I applied the branch and reproduced the following against an in-memory database.

## 1. The freshness helper is wrong, not just misapplied (blocking)

`utils/session.py:129-161` compares each login against `_todays_rollover_boundary()` (`utils/session.py:97-107`), which returns today's clock time unconditionally. Whenever "now" is before that clock time, which with the default `SESSION_EXPIRY_TIME=03:00` is every night from midnight to 03:00 IST, a login made the previous evening reads as pre-boundary and `has_login_this_trading_session()` returns False for a session that is unambiguously alive.

The codebase already knows about this window and handles it correctly in two other places:

- `is_session_valid()` at `utils/session.py:87` guards it two-sidedly: `if now_ist > daily_expiry and login_time < daily_expiry`.
- `get_trading_session_date()` at `utils/session.py:112-124` documents it in its own docstring: "Between midnight and that boundary the session still belongs to the previous calendar date".

`_has_fresher_session` does neither. Reproduced with the boundary set two hours ahead, one `active_sessions` row logged in one hour ago and a non-revoked Auth row: `get_trading_session_date()` correctly returned the previous date, the `is_session_valid` comparison correctly said not expired, and `has_login_this_trading_session()` still returned False. `get_auth_token_broker()` then returned `(None, None)` and left the row `is_revoked=True`, `broker=''`, token blob empty, `active_sessions` count 0.

MCX runs to 23:30 IST, so there are live positions and live pollers in that window. This helper is safe today only because its one consumer, `services/order_update_service.py:247`, is non-destructive: it skips starting an adapter and self-heals at the next login. Wiring it to `revoke_user_tokens()` is what turns a conservative heuristic into data loss.

## 2. An empty `active_sessions` table is not proof of a stale broker session (blocking)

`blueprints/auth.py:707` (reset), `:811` and `:1469` (change password) call `clear_user_sessions(username)` and deliberately leave the broker token alive, so the user can log back in and `_try_resume_broker_session` skips broker OAuth. The only `revoke=True` in that file is the logout path at `:1331`.

Reproduced: seed a live token plus a fresh login today, confirm `get_auth_token_broker(KEY)` returns `('live-broker-token', 'zerodha')`, then clear sessions exactly as `auth.py:811` does. The next call returns `(None, None)` with `is_revoked=True`. Every running strategy gets a 403 mid-market-day and the user must redo full broker OAuth.

## 3. The teardown is irreversible, which the PR description does not reflect (blocking)

`revoke_user_tokens` reaches `upsert_auth(username, "", "", revoke=True)` at `utils/session.py:269`, and `upsert_auth` overwrites `auth_obj.auth` with `encrypt_token("")` and `auth_obj.broker` with `""` at `database/auth_db.py:544-548`. Verified: after one guard firing the row reads `broker=''` and an empty token blob. This is credential destruction, not a reversible flag flip, and `blueprints/auth.py:202` then refuses to resume. Please size the risk of items 1, 2 and 4 against that, not against "the next login re-authenticates anyway".

## 4. The signal fails open to stale (blocking)

`get_active_sessions` (`database/auth_db.py:379-397`) catches `Exception` and returns `[]`, and `_has_fresher_session` catches `Exception` and falls through to `return False`. A transient SQLite lock is therefore indistinguishable from "no login today". Reproduced with a raising `get_active_sessions`: the log shows `Error checking for fresher sessions ... database is locked` immediately followed by the rollover teardown, and the row comes back `is_revoked=True`. CLAUDE.md calls out that SQLite locking is stricter on Windows. Please distinguish "confirmed pre-boundary login" from "could not determine" and fall back to today's non-destructive behaviour on the latter.

## 5. The once-per-trading-session marker can last 5 minutes (blocking)

`_rollover_teardown_done = TTLCache(maxsize=128, ttl=get_session_based_cache_ttl())` is evaluated at import, and `get_session_based_cache_ttl` clamps with `max(300, ...)` at `database/auth_db.py:132`. Measured: a worker started 1 or 4 minutes before `SESSION_EXPIRY_TIME` gets `ttl=300`. Since `upsert_auth`'s skip gate is `if not (token_changed or revoke)` and `revoke=True` always falls through (`database/auth_db.py:581-587`), each repeat runs in full. Instrumented over 4 marker windows of 20 polls each: 8 ZeroMQ `CACHE_INVALIDATE_ALL` publishes, 4 `cleanup_pools_for_user` calls, 4 `stop_order_update_adapter` calls, 4 symbol master cache clears, plus a `force_logout` broadcast each time. A restart just before the 03:00 rollover is exactly when operators schedule one. Key the marker on the trading date with a lifetime that cannot expire mid-session.

## 6. The guard sits above the cache read

`database/auth_db.py:1073-1094` runs before `if cache_key in auth_cache:`. Two consequences, both measured:

- 5 fully warm calls issue 10 SELECTs (5 on `active_sessions`, 5 on `auth`) versus 5 on main. `get_auth_token_broker` is on the order path via `place_order_service.py:320`, `modify_order_service.py:241`, `cancel_order_service.py:239` and around 35 other call sites, and under NullPool each added SELECT opens and closes a fresh SQLite connection.
- `auth_cache[cache_key] = negative_result` at `:1093` is unreachable on the return path. `has_login_this_trading_session` was called 5 times over 5 stale calls, so the PR description's claim that "a poller stops re-querying the database once the answer is known" is not what the code does. Please correct that line.

Move the guard below the cache read, and either drop the negative write or put it where it is consulted.

## 7. Tests

`uv run pytest test/test_broker_session_rollover_guard.py -q` printed, from my machine:

```
Handshake status 403 Forbidden -+-+- {'date': 'Mon, 24 Aug 2026 07:41:37 GMT', ...}
b'{"status":"error","message":"Authentication failed."}'
```

That is a live response from Zerodha's server, which contradicts the checklist item "New tests need no broker credentials or network". `_seed_connected_broker` calls `upsert_auth`, which reaches `start_order_update_adapter` at `database/auth_db.py:621-629`, and `publish_all_cache_invalidation` and `cleanup_pools_for_user` are left unstubbed, so real threads and ZMQ sockets are created per fixture. `test/test_auth_upsert_multisession.py:47-58` shows the pattern to copy.

Please also:

- Drop the three `raising=False` monkeypatches in `TestServiceResponses`, which would silently pass if `is_broker_session_stale` were renamed.
- Add a case that seeds cleared sessions with a live token (the password-change state) and asserts the token survives.
- Add a case that pins the pre-boundary window from item 1: a login one hour ago with the boundary still ahead must resolve, not revoke.
- Register the file in the CI-safe list at `.github/workflows/ci.yml:46-53` once it no longer touches the network. As written it would never run in CI.

## 8. Smaller items

- `services/quotes_service.py:370` (`get_multiquotes`) is in the file you edited and still returns 403 "Invalid openalgo apikey" for the same condition, so `/api/v1/quotes` and `/api/v1/multiquotes` now disagree. Either fix both or return the distinct code from the shared resolution point so all callers inherit it. `docs/api/market-data/quotes.md` and `history.md` carry no error-code table at all today, so if you introduce a new public code please add one.
- `blueprints/auth.py:1013` inherits the item 1 defect read-only: between midnight and `SESSION_EXPIRY_TIME` every browser is told `broker_session_expired: true` and renders Reconnect Broker against a live session, while `is_session_valid()` still says the app session is fine, so nothing self-corrects (`_touch_session_heartbeat` updates `last_seen`, not `login_time`).
- `services/place_order_service.py:319` resolves the credential before `place_order_with_auth` reaches its analyze-mode branch at `:149`, so the guard also blocks sandbox order placement, which never talks to the broker. Worth a decision either way.
- Three blank lines were introduced above `def get_auth_token_broker` in `database/auth_db.py`. Just delete the extra one. Do not run `uv run ruff format .` on that file: it reports a large number of pre-existing reformats this repo has never applied, and applying them would bury the real change.

## Direction

The correct shape is smaller than what is here. Either stamp the token's issue time on the Auth row and compare that (which removes the dependency on `active_sessions` entirely and fixes items 1, 2 and 4 at once), or keep the inference but make it three-valued: fresh, confirmed stale, unknown. Confirmed stale means at least one session row exists and all of them predate a boundary that has actually been crossed. Unknown keeps today's non-destructive behaviour. Whichever you pick, please decouple "stop handing out this credential" from "destroy it": returning `None` without revoking already fixes #1858 for the API path, and the `#1419` resume concern is better handled at the resume site in `blueprints/auth.py:189-215` than by wiping the row from the hot path.

No schema change is involved, so no migration is needed.

One process note: CI has not run on this branch. The only green checks are Snyk reporting no manifest changes and a neutral cubic review, so nothing here has been through lint, the backend test job or the frontend build. Please do not read the check marks as validation.

Pragitics:
Thank you for the detailed review. You were right on all five blocking points, and the fix is now a good deal smaller than what I first pushed. Everything is in `87109ea`.

**The main change: it no longer deletes anything.**

The guard now just refuses to hand out the credential. It does not touch the row, so nothing is destroyed and there is nothing to undo. That single decision removes most of what you flagged: no revoke, so no ZeroMQ invalidation, no pool cleanup, no adapter stop, no force logout, and therefore no once-per-session marker whose TTL could expire early. Items 3 and 5 stopped existing rather than being fixed.

Your point about #1419 was the thing that unblocked this for me. I went and read the resume path, and `_try_resume_broker_session` already calls the broker to check the token before accepting it, and falls back to normal OAuth when that fails. So revoking here was never needed to prevent a dead token being resumed.

**The boundary bug.**

You were right that this was the real defect, not just a misapplication. `_todays_rollover_boundary()` hands back today's 03:00 even when 03:00 has not happened yet, so between midnight and the boundary a login from the previous evening looked stale. There is now a `_current_session_boundary()` that steps back a day when the boundary has not been crossed, which is the same two-sided check `is_session_valid()` already does. Fixing it in the shared helper also cleared up the read-only version of the same bug in `/auth/session-status` that you noticed.

**Three states, as you suggested.**

`broker_session_freshness()` returns fresh, stale, or unknown. Only a confirmed stale reading withholds anything. No session rows, an unreadable timestamp, or a failed read all answer unknown and behave exactly as before. That covers both the password-change case and the locked-database case, since `get_active_sessions` returns `[]` for both and they genuinely cannot be told apart.

I went with this rather than adding a timestamp column to the auth row. A new column needs a migration for every existing install, and there is nothing on an old row to backfill it from, so every existing row would be null. Null means unknown, which means the three-state logic would still be needed, just with a migration on top.

**Something you did not catch, which I should mention.**

Taking out the teardown also took out its accidental rate limit. The rejection was logging once per call. I measured 520 identical lines from 520 calls, which for a strategy polling every few seconds would be roughly 720 log lines an hour with no restart to truncate them. There is now a small per-trading-session gate on that line, so it prints once per rollover. It gates only the log line, so if it expires early the worst case is a duplicate message. I found this while running the fd-audit measurement you would have expected me to run.

**Two places where I did not follow your advice literally.**

First, moving the guard below the cache read was not enough on its own. Even below the cache, a warm call still paid for one `active_sessions` read, so your 10-versus-5 count would have stayed. I added a 30 second cache of the freshness answer as well, cleared by `upsert_auth` so a real login is visible straight away. Warm calls are back to roughly what main does.

Second, I reverted the `username` and `current_session_id` parameters I had added to `revoke_user_tokens()`. Once the teardown was gone nothing passed them, and leaving unused parameters behind seemed worse than removing them. That function no longer appears in the diff against main at all.

**On sandbox.**

I left it, but I want to flag the reasoning rather than have it look like I missed it. A confirmed-stale session now declines a sandbox order that never reaches the broker. An already-revoked token does the same thing today, so this widens existing behaviour rather than introducing it, and moving the analyze branch above credential resolution felt like a separate change. Happy to fold it into this PR if you would rather have it here.

**Tests and CI.**

You were right that the tests were reaching Zerodha. `upsert_auth` was starting a real order-update adapter through the fixture. Those side effects are stubbed now, following `test_auth_upsert_multisession.py`, and the `raising=False` monkeypatches are gone. I added the two cases you asked for, plus ones for an unreadable session store, an unparseable login time, the credential surviving, and the log volume. The file is registered in the CI-safe list.

I also checked each new test by reverting its fix on its own and confirming it goes red. Full suite compared against main by failure list rather than count: identical, no new failures.

CI has now run on this branch for the first time. backend-lint, backend-test, frontend-build, frontend-lint, frontend-test and frontend-e2e have all passed. security-scan is still going as I write this, so please take it as green apart from that one.

The PR description has been rewritten. The claim about no longer re-querying the database is corrected, since you were right that the line was unreachable, and there is now a section listing what I have not verified. I also added the error-code tables to the quotes, multiquotes and history docs, since multiquotes now returns the same 401 as quotes.

marketcalls:
Approving after three small fixes. This is a different patch, not a patched patch, and the score moves from 40 to 80. I re-ran last round's reproductions against `277c12ef` rather than reading the description, and I mutation-tested each fix rather than trusting your revert table.

What holds up:

- **No revoke anywhere.** I seeded a real `active_sessions` row through `register_session` (no monkeypatch) with a login five hours before the boundary. It round-trips as naive IST `2026-08-23T22:00:00`, the verdict is `stale`, `get_auth_token_broker` returns `(None, None)`, and five consecutive rejections leave the row `is_revoked=False`, `broker='zerodha'`, token intact. `revoke_user_tokens` is byte-identical to main: 6571 bytes on both sides, head line 241.
- **Undecidable states.** A raising `get_active_sessions`, the `[]` the real function returns, `clear_user_sessions()` and an unparseable `login_time` all give `unknown` and all still resolve. Collapsing `unknown` into `stale` fails two tests, so that half is genuinely pinned.
- **The guard.** Disabling it fails four tests. Well pinned.
- **Warm path.** I attached a `before_cursor_execute` listener to the head AND to main. Both give `{'auth': 5}` for five warm calls. Exactly main's number, zero `active_sessions` reads on the warm path.
- **Streaming.** The WebSocket path resolves through `get_auth_token(user_id)` (`websocket_proxy/base_adapter.py:446`) and the proxy authenticates clients with `verify_api_key` plus `get_broker_name` (`server.py:835,845`). Neither touches `get_auth_token_broker`, so the shared feed never sees this guard.
- **Hygiene.** The GitHub contents API confirms 0 CR bytes in every blob (a local check with `core.autocrlf=true` gives a false positive, so ignore that). Zero non-ASCII in all 781 added lines; every em dash in the touched files is on a pre-existing context line. `openalgo-charts` still exactly `1.6.0`. Ruff parity with main on both `check` (6 UP035) and `format` (the same 3 files reformat on main). The merge with main is clean, keeps both ci.yml registrations, and the merged CI-safe list runs 71 passed locally. CI is green on all five jobs that can actually fail.

## Where I have to correct you

**The pre-boundary test does not pin the pre-boundary fix.** Your table says reverting `_current_session_boundary` to `_todays_rollover_boundary` gives "1 failed". I did exactly that at the head and got `18 passed in 2.23s`.

The reason is the test's own clock. It sets `SESSION_EXPIRY_TIME` to `(now_ist + timedelta(hours=2)).strftime("%H:%M")`. From 22:00 IST onward that wraps past midnight, so `now_ist.replace(...)` lands earlier today, `now_ist < boundary` is False, the day-subtraction branch is never taken, and the assertion passes with or without your fix. Enumerating half-hour steps over a full IST day, the fixed and unfixed verdicts are identical at 22:00, 22:30, 23:00 and 23:30, which is 16:30 to 18:29 UTC. Roughly one CI run in twelve.

The code fix is right, and I verified it separately. It is only the regression pin that is missing, and it is missing on the one defect that scored CRITICAL last round. Please monkeypatch `_todays_rollover_boundary` (or the IST reader) so the window is chosen rather than inherited from the wall clock.

Two smaller corrections. `test/test_broker_session_rollover_guard.py:137` still has `monkeypatch.delenv("DISABLE_SESSION_EXPIRY", raising=False)`, which is fine, but the three `setattr(..., raising=False)` calls I actually objected to are gone, so the claim "the raising=False monkeypatches are gone" is right in substance and wrong as written. And your Changes table says `_has_fresher_session` and `has_login_this_trading_session` keep "semantics unchanged"; they do not. Both now use the corrected boundary, and `has_login_this_trading_session` gates whether the order-update adapter starts at boot (`services/order_update_service.py:247`). That change looks correct to me, it just contradicts your own boundary section four paragraphs earlier.

## Three things to fix before merge

1. **`services/depth_service.py:150`** checks the tuple length, not the token:

```python
auth_info = get_auth_token_broker(api_key, include_feed_token=True)
if len(auth_info) == 3:
    AUTH_TOKEN, FEED_TOKEN, broker_name = auth_info
```

Your guard returns `(None, None, None)`, so `len(...) == 3` is true, `broker_name` becomes `None`, and `get_depth_with_auth` falls through to `import_broker_module(None)`. Measured directly: `get_depth_with_auth(None, None, None, 'SBIN', 'NSE')` returns `404 "Broker-specific module not found"`. The bug predates you, but until now it only fired on a bad key; this makes it fire every morning on an endpoint sitting in the same `docs/api/market-data/` folder as the three files you documented. I audited all 39 call sites: depth is the only `len()`-based one. Add the `AUTH_TOKEN is None` check plus the same 401 branch and a fourth table in `depth.md`.

2. **The boundary test**, per above.

3. **The PR body**: five em dashes (U+2014), repo style is ASCII only including descriptions. While you are in there, fix the revert table, the "semantics unchanged" line, and "CI has not run on this branch", which is now stale.

## Not blocking

- **`blueprints/auth.py`.** `get_session_status` now calls `is_broker_session_stale_for_user`, but nothing in `frontend/src` reads `broker_session_expired`; the only consumer of the concept is `Dashboard.tsx:84` reading `code === 'BROKER_SESSION_EXPIRED'` from `/auth/dashboard-data`. `AuthSync.tsx` still takes the `logged_in && broker` branch and simply never sets the api key. So the flag is inert today, and the file has no test. Either wire it up or drop it, but please add a test either way, since `/auth/session-status` is in `check_session_expiry`'s skip list and is reachable without the sweep.
- **One fail-closed hole.** `register_session` swallows its exception and returns False, and `handle_auth_success` ignores the return value. If that write fails at login, the user has a live token and a fresh cookie while `active_sessions` holds only yesterday's row: a decidable but wrong `SESSION_STALE`, which refuses a working credential for the rest of the day. Low probability, but it is the exact class of failure your design says it avoids, and nothing logs the divergence. Worth a log line at minimum.
- **Sandbox.** Thank you for raising it. One nuance: before this PR an API-only user's sandbox market orders could not fill either, because `execution_engine._fetch_quote` goes through the same `get_quotes` with the same dead token. What changes is acceptance-then-no-fill becoming outright rejection. Still worth a follow-up that moves the `get_analyze_mode()` branch above credential resolution, just not here.
- **The other callers.** 45 occurrences across 38 files still say "Invalid openalgo apikey". I reproduced it: with a seeded stale session, `quotes` answered 401 while `intervals` and `funds` answered 403. Landing the market-data endpoints first is the right call.
- **`_stale_logged`** has a 3600s TTL, so the line prints once an hour, not once per trading session. Make the name match.

You are behind main by one commit; the trial merge is clean and keeps both test registrations, so just merge main before this lands.

Pragitics:
Thank you, and thank you especially for mutation-testing my revert table instead of reading it. All three fixes are in `6941ae8`, along with the four non-blocking items. Both of your corrections were right.

**The boundary test.**

You were right, and I reproduced it before changing anything rather than taking it on trust. Enumerating half-hour steps across a full IST day with the old test, the fixed and unfixed verdicts agree at exactly 22:00, 22:30, 23:00 and 23:30. The same four windows you named, 4 of 48. The `strftime("%H:%M")` wraps past midnight after 22:00, so the day-subtraction branch is never taken and the assertion holds either way.

It now monkeypatches `_todays_rollover_boundary` so the window is chosen instead of inherited from the wall clock. The same enumeration against the new test gives 0 of 48. I also added a second test that walks all 24 hours through `_current_session_boundary(now_ist=...)` directly, which needs no clock patching at all, since that function already takes its "now" as an argument. Reverting the call site fails one test, and removing the day-subtraction inside the helper fails both.

**depth_service.**

Confirmed and fixed. `include_feed_token=True` always returns a 3-tuple, so `len(auth_info) == 3` was always true and the 403 underneath it was unreachable. It now checks `AUTH_TOKEN is None` like quotes and history, with the 401 branch and a fourth table in `depth.md`. I pinned both branches, not just the new one: there is now a test asserting a bad key gets 403 rather than the 404 from `import_broker_module(None)`. Restoring the length check fails both.

**Both of your corrections to my description are accurate.**

"Semantics unchanged" was wrong. The body now states plainly that `_has_fresher_session` and `has_login_this_trading_session` both inherit the corrected boundary, that `order_update_service.py:247` will therefore start the adapter during the midnight to 03:00 window where it previously refused, and that `/auth/session-status` stops reporting a live session as expired in that window. I also dropped the overstated `raising=False` sentence.

**One place I did not follow your advice literally.**

On `/auth/session-status` you offered wire it up or drop it. I kept it and added tests rather than dropping it, because dropping it puts the endpoint back to reporting a dead broker session as connected, which is the same defect this PR exists to fix, on an endpoint that is in `check_session_expiry`'s skip list. The flag is inert today and I have said so in the description rather than implying a UI reads it. Four tests: confirmed stale reports the flag without downgrading the app session, fresh reports normally, an undecidable reading raises no false alarm, and a missing token still reports reconnect.

Worth mentioning that my first version of that undecidable test stubbed `is_broker_session_stale_for_user` to False, which is the same value the fresh test stubs, so it asserted nothing. It now seeds an empty `active_sessions` and calls the real function. Collapsing `unknown` into `stale` fails it.

**The smaller items.**

`_stale_logged` is now `_stale_log_throttle`, and the comment says at most once an hour per user rather than once per trading session. You were right that the name promised something the TTL does not deliver.

`register_session` now names the consequence in its failure log, and uses `logger.exception` per the repo convention. The point that the caller ignores the return value is the part worth logging, since the symptom otherwise looks like a broker outage rather than a failed session write.

Sandbox: agreed, follow-up rather than here. Thank you for the `_fetch_quote` nuance, I had not traced that path and it changes the framing from "this breaks sandbox" to "acceptance-then-no-fill becomes rejection".

The remaining "Invalid openalgo apikey" callers are untouched.

**Verification.**

Main is merged, the trial merge kept every ci.yml registration, and both new test files are in the CI-safe list. Full suite diffed against a fresh baseline on current main: identical 18-line failure lists, no new failures, 25 more passing. Every new test reverted individually and confirmed red. No non-ASCII in any added line or in the description, and no CR bytes.

CI is green on this commit: backend-lint, backend-test, frontend-build, frontend-lint, frontend-test, frontend-e2e, security-scan and both Snyk checks.

marketcalls:
Not merging yet, and I want to be precise about why: this fails on completeness, not on safety. The security direction is right and the redesign is honest work. Score moves 40 to 80 to **75**, and the small drop is my own earlier review being too narrow, not you regressing.

I ran a four-lens audit on `6941ae8b` and then an adversarial pass over its own findings. That second pass refuted seven of them, including two of mine. I am reporting only what survived.

## What I verified holds

The containment that makes "withhold instead of revoke" safe is real, and I traced it rather than taking the description's word:

- The guard's only effect is returning `(None, None)`. No path from it reaches `upsert_auth`, `revoke_user_tokens`, `cleanup_pools_for_user`, or the ZeroMQ `CACHE_INVALIDATE_ALL` publish.
- The #1591 multi-device invariant holds. Your two cache clears at `database/auth_db.py:584-585` sit above the `if not (token_changed or revoke): return` gate, so they run on every login without arming the teardown.
- A false verdict cannot touch the shared feed: the proxy never calls `get_auth_token_broker`.
- It is not an API-key oracle. `is_broker_session_stale` calls `verify_api_key` first and returns False on failure. I also checked whether the second `verify_api_key` double-counts the invalid-key ban tracker, and it does not: `verify_api_key` consults `invalid_api_key_cache` at step 1, so a probe showed exactly one tracker hit per request. I had this down as a likely finding and it is not one.
- Boundary math is sound. Across all 72 half-hour points of a day the boundary was never in the future and never more than 24 hours behind, zero violations, and `Asia/Kolkata` has no DST so the un-normalized day subtraction is exact.
- Hygiene: bounded `TTLCache`es, no new FDs, no asyncio, `logger.exception` throughout, no emoji or dashes in the 781 added lines, ruff parity with main, 25 tests with both files registered in `ci.yml`.

`_current_session_boundary()` is a genuine bug fix and it is the strongest thing in this PR.

## Blocking

**1. The guard covers 4 of 39 consumers.** `grep` gives 39 `get_auth_token_broker(` call expressions outside `test/` and `auth_db.py`. `is_broker_session_stale` appears in exactly three service modules, covering four endpoints. The other 35 still answer `403 "Invalid openalgo apikey"` for a key that is perfectly valid, which is the precise misdiagnosis this PR exists to remove. `place_order`, `place_smart_order`, `split_order`, `modify_order`, `cancel_order`, `close_position`, `positionbook`, `holdings`, `funds` all still lie about the cause.

**2. Four of the guarded consumers need no broker session at all.** This is the part that turns a labelling change into lost functionality:

- `services/ping_service.py:10-26` ignores the token entirely and returns pong. `/api/v1/ping` is the canonical health check, so external monitoring will report "Invalid openalgo apikey" every morning after 03:00.
- `services/symbol_service.py:109-113` reads the local `SymToken` table.
- `/api/v1/intervals` is a static map.
- The analyzer status and toggle endpoints touch no broker.

**3. Sandbox order placement is blocked by a live-broker verdict.** `services/place_order_service.py:320` resolves the token before `place_order_with_auth` reaches `if get_analyze_mode()` at line 149. I ran the same sandbox order twice in one process against identical DB state, swapping only the resolver: with your guard it returns `403 "Invalid openalgo apikey"` and the sandbox is never reached; with main's resolver the sandbox is reached. CLAUDE.md documents the sandbox engine as fully isolated from live trading, and this couples it to one. To be clear, my audit initially attributed this to main and that was wrong; the reproduction shows it is introduced here.

**4. A failed `register_session` withholds a freshly minted live credential.** `utils/auth_utils.py:395` discards the boolean, then `upsert_auth` runs at line 428 in a separate transaction. `register_session` (`database/auth_db.py:331-378`) rolls back DELETE plus COUNT plus INSERT as a unit, restoring the previous day's row. Result: today's live token, yesterday's session rows, verdict STALE. A transient SQLite lock is enough, and CLAUDE.md notes locking is stricter on Windows. I am calling this MEDIUM rather than HIGH because I confirmed it self-heals: the next successful login restores it, since the same-IP DELETE is idempotent. Forcing SESSION_UNKNOWN for the day when `register_session` returns False would close it.

**5. The honest 401 is transient, and then the docs mis-instruct.** This one no lens caught. `get_auth_token_broker` short-circuits at `auth_db.py:1119-1121` before consulting freshness, and `is_broker_session_stale` returns False when `active_sessions` is empty. So once the rollover sweep runs and clears those rows, the verdict becomes UNKNOWN and the same condition reverts to the misleading 403. Your new tables in `docs/api/market-data/*.md` then tell the reader that 403 means "re-issue the key at /apikey", which cannot change a verdict computed from session rows.

## Not blocking, but worth fixing while you are here

- **REST and streaming disagree.** `get_auth_token` (`auth_db.py:651`) and `get_feed_token` (`:745`) are unguarded, so the WebSocket path still serves the credential the REST path withholds.
- **`broker_session_expired` is write-only.** `grep -rn broker_session_expired frontend/src` returns nothing. The only consumer of a broker-expiry signal is `Dashboard.tsx:84`, reading `code === 'BROKER_SESSION_EXPIRED'` from a 401 on `/auth/dashboard-data`, and that endpoint (`blueprints/auth.py:1227`) resolves through the unguarded `get_auth_token`. Related: the comment you added at `blueprints/auth.py:1028-1030` asserts a recovery contract that neither named endpoint implements for the state this PR creates.
- **The crypto bypass is on one side only.** `is_broker_session_stale_for_user` short-circuits on `is_session_expiry_disabled()`, but `has_login_this_trading_session` (`utils/session.py:222-241`), which you rewrote in the same commit to share the helper, did not get it. On a 24/7 Delta Exchange instance that reads stale.
- `_session_freshness_cache` at ttl=30 serves a dead credential for up to 30 seconds past each rollover. Self-healing and still better than main, so this is a note, not a request.

## What I would actually do with this

Split it. `_current_session_boundary()` is a real, self-contained bug fix that stands on its own merits and I would merge that today. The guard is a bigger change than its diff suggests, because putting it at the shared resolution point means all 39 consumers inherit it, and right now four of them are wrong and 35 report the wrong cause. That is a design pass, not a patch.

If you would rather keep it as one PR, then items 1 to 5 need to land together, and I would want the ping, symbol, intervals and analyzer paths excluded from the guard explicitly rather than by omission.

One process note: scope went from 3 files to 13 and +1043/-24 across three rounds. Some of that is my fault for asking for the docs. But `depth_service`, `quotes_service` and the second test file arrived on top of a PR that closes a single issue, and reviewing a moving target is how item 5 stayed hidden for three rounds.

The work here is good and the reasoning in your last two replies was better than mine in places. I want this in, just not as one 13-file change on the credential path.

marketcalls:
Separate from the code review above, here is a contributor scorecard for how this PR was worked, since you asked for direction rather than just a verdict. This scores your **process across four rounds**, not the current diff. The diff is 75. You are **77**.

Those two numbers being close is a coincidence worth explaining: the artifact is held back by one structural decision, while the way you worked is better than the artifact. If the completeness gap closes, the diff score moves a long way and this one barely moves.

| Dimension | Score | Short version |
| --- | --- | --- |
| Problem diagnosis | 19/20 | Excellent. Best part of your work. |
| Responsiveness to review | 19/20 | You redesign rather than patch. Rare. |
| Verification rigor | 15/20 | Strong now, but one table was asserted not run. |
| Completeness of the change | 11/20 | The gap. 4 of 39 consumers. |
| Scope discipline | 6/10 | 3 files became 13 mid-review. |
| Communication accuracy | 7/10 | Three descriptions needed correcting. |
| **Total** | **77/100** | |

## What you are already doing better than most

**Diagnosis (19/20).** Round 1 was not "I think this is broken". You landed on `app.py:510` skipping `/api/`, `check_session_expiry` reading a cookie an API-key caller never sends, and `order_update_service.py:247` already carrying a comment describing the same gap. You found the right layer to fix on the first attempt. That is the hardest part of contributing to a codebase you did not write, and you did it cold.

**Redesign over patching (19/20).** When I gave you five blocking findings, the obvious move is five fixes. You instead asked what the guard should *do*, decided it should withhold rather than revoke, and three of the five findings stopped existing. Your sentence "items 3 and 5 stopped existing rather than being fixed" is exactly the right instinct. Most contributors patch. Also: you went and read `_try_resume_broker_session` yourself and worked out the revoke was never needed, rather than taking my word for the direction.

**You reported a regression I had missed.** Removing the teardown removed its accidental rate limit, and you measured it: 520 identical log lines from 520 calls, roughly 720 an hour for a polling strategy, in a worker that never restarts. Nobody asked. Volunteering a defect in your own change, against your own interest in getting it merged, is the single strongest signal in this whole thread. Keep doing that.

**You reproduce before changing.** In round 4 you enumerated half-hour steps across a full IST day, found the fixed and unfixed verdicts agree at exactly 22:00, 22:30, 23:00 and 23:30, and confirmed my correction independently before touching anything. Then you found a latent bug in your own new code, the `len(auth_info) == 3` check in `depth_service` that made a 403 unreachable.

## Where to improve, most valuable first

**1. Completeness: follow the change to every consumer (11/20).** This is the one that matters and it is what blocks the PR.

You put the guard at the shared resolution point, which was the right call. But a chokepoint means every caller inherits the behaviour, and you wired the honest 401 into 3 service modules covering 4 endpoints out of 39 call sites. Worse, four of the endpoints you did guard need no broker session at all: `ping` is the canonical health check, `symbol` reads a local table, `intervals` is a static map, and the analyzer toggle touches no broker.

The habit to build: after changing a shared function, run `grep -rn "get_auth_token_broker(" --include="*.py" . | grep -v test/` and read every hit. Not skim. For each, ask whether the new behaviour is correct there. Thirty-nine is a lot, but it is one afternoon, and it is the difference between a change that is safe and a change that is finished.

**2. Verification rigor: never ship a revert table you have not run (15/20).** Your round-2 table said reverting `_current_session_boundary` gives "1 failed". It gave 18 passed. The test derived its own clock from `(now + 2h).strftime("%H:%M")`, which wraps past midnight, so the branch was never exercised.

You clearly can do this work, and round 4 proved it. The gap is that you produced the table from what the test was *meant* to prove rather than from running the revert. A claim in a PR description carries the same weight as a claim in code. If you write "reverting X fails N tests", revert X and run it, every time. When I mutation-test a table and it does not hold, everything else in that description becomes suspect, which costs you a round.

**3. Scope discipline (6/10).** This PR opened at 3 files and is now 13, +1043/-24, for one issue. Some of that is mine, since I asked for the docs. But `depth_service`, `quotes_service` and a second test file arrived on top of a change already under review.

Reviewing a moving target has a concrete cost, and you can see it in the review above: the transient-401 problem survived three rounds partly because the surface kept changing underneath. When a review turns up an adjacent problem, the strong move is a new issue or a follow-up PR, exactly as you did with the `@cross_origin` handlers on a different PR. That was the right instinct; apply it here too.

**4. Communication accuracy (7/10).** Three separate description corrections across four rounds: "semantics unchanged" contradicting your own boundary section, the overstated `raising=False` sentence, and "CI has not run on this branch" left in after CI ran green. Individually trivial. Together they mean a reviewer cannot use your description as a map and has to re-derive everything from the diff, which is slow and is how things get missed.

Before pushing, reread the description against the diff as if you were the reviewer, and delete anything you have not personally confirmed at the current head.

## The one habit that would move you fastest

You are strong at "is this correct?" and weaker at "what else does this touch?".

Concretely, before your next push: pick the function your change modifies, list every caller, and write one line per caller saying what the new behaviour does there. If that list has an entry you cannot justify, the change is not finished. That single exercise would have caught the ping endpoint, the sandbox coupling, and 35 of the 39 misleading 403s before I ever saw the PR.

Do that and you are comfortably in the 90s. The judgement is already there.

Pragitics:
`get_auth_token_broker()` returns `(None, None)` for four different conditions, so every caller reported all four as "Invalid openalgo apikey".

```python
def credential_error(api_key):
    if verify_api_key(api_key):
        return {"status": "error", "code": "BROKER_SESSION_EXPIRED", ...}, 401
    return {"status": "error", "message": "Invalid openalgo apikey"}, 403
```

@marketcalls So now it separates a bad API key from a missing broker session, in one place that every caller inherits.

Pragitics:
Adding the decisions that are not visible in the diff, on top of the note above.

**I kept it as one PR.** You offered that as the alternative provided items 1 to 5 land together, and they do. I want to be straight that this is the weaker choice on your scope-discipline point rather than pretend otherwise. The reason is that 1, 2 and 5 turned out to share one root cause, that the return value cannot carry the reason, so fixing it centrally closed all three at once. Splitting would have landed the boundary fix today and left the misdiagnosis in place for another round. If you would still rather have two, say so and I will carve it.

**Item 5 is gone by construction, not patched.** The verdict now comes from `verify_api_key` alone, so it no longer reads `active_sessions` and cannot flip back to the 403 once the sweep clears those rows. That was the point I would most like you to re-check, since it is the one you said no lens caught.

**Streaming is deliberately untouched.** You listed `get_auth_token` and `get_feed_token` as worth fixing while I was here, and I have not. Under "what I verified holds" you wrote that a false verdict cannot reach the shared feed because the proxy never calls `get_auth_token_broker`. Guarding those two readers would remove exactly that containment. Happy to be overruled, but I did not want to trade it away silently.

**The crypto bypass was one-sided and is fixed** in `237eb1fed`. `has_login_this_trading_session` did not short-circuit on `is_session_expiry_disabled()`, and it gates whether the order-update adapter starts at boot, so a Delta Exchange instance got no adapter. Two tests: a crypto instance with a three-day-old login reads fresh, and a non-crypto instance with the same rows still reads stale, so the exemption cannot leak.

**On the red `frontend-test` leg from last round.** It now passes on this branch, same tests, no frontend file changed between the two runs. The root cause is real though: the tests call `screen.getByRole('tab', { name: 'Greeks' })`, which is synchronous and throws on the first miss instead of retrying, so `SLOW_INTEGRATION_TEST_TIMEOUT` could never have fixed it. That is why `2d8edfd0e` did not settle it. `findByRole` does retry. Raised separately as #1903, one file, six lines, test-only, rather than folding it into this one.

Everything else from your list is in: depth checks the token instead of the tuple length with both branches pinned, the boundary test now monkeypatches the helper so the window is chosen rather than inherited from the wall clock, and the description no longer claims "semantics unchanged" or that CI has not run. 47 tests across three files, all registered in the CI-safe list, and every fix reverted individually and confirmed red rather than asserted.

One gap from the description is now closed. The sandbox ordering was pinned only by a source-order assertion, because `place_order_service` cannot be imported on its own: line 8 imports `restx_api.schemas`, whose package `__init__` pulls in `options_multiorder`, which imports `place_order_service` back. Importing the `restx_api` package first lets it complete, so there is now a functional test that calls `place_order()` in analyze mode with a resolver that raises if consulted. Both it and the ordering assertion fail if the branch moves back.

Still not verified, in case it matters: no live broker was involved anywhere, and the pre-boundary window is exercised by monkeypatching the boundary helper rather than by running the suite at 01:00 IST.

marketcalls:
Please pick this back up when you can. It is one of the two highest-impact PRs still open, because the bug it fixes fires nightly and sits on the shared broker-credential path. Score is **80** and the security posture is sound; what is left is three concrete items, two of them one-liners.

To be clear about the direction of travel: the containment is verified and holds. No path from the guard reaches `upsert_auth`, `revoke_user_tokens`, `cleanup_pools_for_user` or the ZeroMQ invalidation; the #1591 multi-device invariant is intact because your two cache clears sit above the `if not (token_changed or revoke): return` gate; and the WebSocket proxy never calls `get_auth_token_broker`, so a false verdict cannot touch the shared feed. `_current_session_boundary()` is a genuine fix for a real overnight defect. None of that is in question.

## 1. The new static test fails on Windows

`test/test_credential_error_uniformity.py::TestEveryConsumerInherits::test_no_service_hand_rolls_the_credential_answer` fails on my machine: `1 failed, 47 passed`.

`glob.glob("services/*.py")` returns `services\ping_service.py` with a backslash, while `NO_BROKER_SERVICES` holds `services/ping_service.py` with a forward slash, so the skip set never matches and the test reports intervals, ping and symbol as offenders. CI is ubuntu-latest so all three backend-test legs are green and this is invisible there.

One line: normalise with `PurePath(path).as_posix()` before the membership test.

## 2. `/api/v1/intervals` now 404s where it used to 403

This is new this round and it is the same shape as the `depth_service` bug you just fixed, one file over.

Exempting intervals from the freshness guard replaced the token with `get_broker_name(api_key)` at `services/intervals_service.py:110`. That returns `None` for a revoked or absent Auth row, so `import_broker_module(None)` tries `broker.None.api.data` and the endpoint answers 404 "Broker-specific module not found" plus an ERROR log line per call. Main answered 403 for the same state.

Guard the `None` and answer `credential_error(api_key)`, or a 400 with an honest message.

## 3. Decide the rest of the sandbox surface, explicitly

You fixed `place_order`, and I mutation-verified both tests that pin it. But `cancel_order`, `modify_order`, `close_position`, `split_order`, `basket_order`, `place_smart_order`, `cancel_all_orders`, `orderbook`, `positionbook`, `tradebook`, `holdings` and `funds` all still resolve the live credential in the entry function before reaching their analyze branch inside `*_with_auth`. After the rollover those sandbox operations answer 401 where main let them through, and CLAUDE.md documents the sandbox engine as fully isolated from live trading.

Either apply the same early return to the rest, or say in the body that `placeorder` is the only one being fixed here and open a follow-up. I do not mind which, but the current state is an undeclared partial.

## On splitting

My last review suggested splitting `_current_session_boundary()` out as its own PR, and I still think that is the faster route: it is a small self-contained fix for a real bug and I would merge it today. The guard is the part that needs the consumer sweep. Your call, and if you would rather keep it as one PR then the three items above are what it needs.

Two smaller things, not blocking: the `broker_session_expired` flag still has no reader in `frontend/src`, and `has_login_this_trading_session` did not get the `is_session_expiry_disabled()` bypass that its sibling got, so a 24/7 Delta Exchange instance reads stale.

Pragitics:
All three items are in, each one reverted and confirmed red rather than asserted. The description is rewritten against the current head.

**1. The Windows path.** Reproduced first: `glob()` yields `services\ping_service.py`, the skip set holds `services/ping_service.py`, so the membership test never matched and intervals, ping and symbol were reported as offenders. There is now a `service_key()` that normalises separators before the split, plus a test that feeds it a literal `C:\repo\services\ping_service.py` so the rule is checked on the ubuntu runner where the bug is otherwise invisible. Reverting it fails 2.

**2. The intervals 404.** Confirmed, and it was the same shape as the `depth_service` one. `get_broker_name()` answers `None` for a revoked or absent Auth row, `import_broker_module(None)` looks for `broker.None.api.data`, and the endpoint answered 404 plus an ERROR line per call where main answered 403. It now guards the `None` and returns `credential_error(api_key)`. Reverting it fails 1, and the mutation reproduces your `broker.None.api.data` in the log.

**3. The sandbox surface.** I took the first option: applied, not declared. Thirteen entry points now check the analyze branch before resolving a credential, up from one. `place_order`, `place_smart_order`, `split_order`, `basket_order`, `modify_order`, `cancel_order`, `cancel_all_order`, `close_position`, `orderbook`, `positionbook`, `tradebook`, `holdings`, `funds`. Reverting the early return from any one of them fails 2.

Both smaller items are in as well. `has_login_this_trading_session` now short-circuits on `is_session_expiry_disabled()` like its sibling, so a 24/7 Delta Exchange instance no longer reads stale and gets no adapter at boot. And `broker_session_expired` has a reader: `/auth/dashboard-data` now asks the freshness question at the call site, which is what `Dashboard.tsx:84` was already listening for but could never receive, since that route resolves through the unguarded `get_auth_token`.

**Two things I want on the record rather than found.**

The streaming readers are still unguarded, deliberately. My reasoning is blast radius: a false stale on REST costs one 401 that self-heals on the next call, while a false stale on `get_auth_token` drops the pooled adapter that serves every device at once. Your own audit listed "a false verdict cannot touch the shared feed" under what holds, and guarding these two is exactly what would remove it. The mechanism differs too, since REST re-resolves per call and streaming resolves once at connect, and the streaming path already has its own boot gate and its own 403 re-read. I have not watched either fire against a live broker, so I am claiming the mechanism differs, not that streaming is correct. If you would rather have the symmetry, I will do it as a follow-up.

Acceptance criterion 4 of #1858 asks the guard to revoke and trigger teardown. It does not, because your first review asked for that path to go and I think you were right. The issue text was never updated, so I have said so in the description rather than leave an unchecked box looking like an oversight.

**On splitting.** Putting it back to you once, now that the work is done rather than in the middle of it. When I declined the split this was 13 files; completing the consumer sweep took it to 49. That is the direction your concern pointed, not away from it. If you want two PRs, say so and I will carve it: the boundary fix is self-contained and the rest rebases onto it. Otherwise this is ready.

One thing I have left alone on purpose: `test_session_expiry.py` and `test_order_adapter_auth_rejection.py` are named in #1858's verification block, both pass, and neither is registered in `ci.yml`. That gap is on main and adding it here is the scope creep you marked me down for, so it is a follow-up if you want one.

Full suite diffed against a fresh baseline on current main: 9 failures on the branch against 11 on main, the branch's being a strict subset, and 1936 passed against 1826. No new failures. CI-safe list is 152 passed.

marketcalls (commented):
Review score: 84/100 at `7f1397f5bbf79b63ca9dc40945f38a836369fef5`. Changes are needed before approval and merge.

- `services/cancel_gtt_order_service.py:166`: The rollover guard still blocks several sandbox operations before their sandbox branch. After rollover, a valid API key cannot cancel or inspect sandbox GTTs or read sandbox openposition/orderstatus despite those operations not using a broker. Ordinary sandbox orders were exempted, leaving inconsistent behavior and unmanaged sandbox triggers. Apply the identity-validated sandbox-first dispatch consistently to GTT create/modify/cancel/book, openposition and orderstatus, and test each after the guard returns no live credential.

- `database/auth_db.py:1`: PR conflicts with current main. This head cannot be merged safely in its current form. Rebase onto current main, resolve conflicts while preserving both fixes, and rerun affected checks.

PR #1870, opened 2026-08-23
fyres env

SandeepSid:
## Summary by cubic
Targets the sample environment to Fyers by setting a Fyers-specific OAuth callback and swapping generic API placeholders for example values. This clarifies Fyers testing but makes the sample non-generic.

- Old redirect: http://127.0.0.1:5000/<broker>/callback; new redirect: http://127.0.0.1:5000/fyers/callback. If you are not using Fyers, change the route to your broker.
- Replace the sample API key/secret with your own before running; do not commit real credentials.
- The header comment now includes “#Sandeep S#tra123”; confirm this is intentional or remove it.

<sup>Written for commit 4afbec6dabfe33706d21fb9ae21a197c8d1adace. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1870?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


SandeepSid (commented):
Fyre ev

marketcalls:
One item first, because it is time sensitive.

## Revoke the Fyers app credentials now

`.sample.env` lines 10 and 11 in this PR contain:

```
BROKER_API_KEY = '0L5ODX...-100'
BROKER_API_SECRET = '9JXBT.....'
```

Those match the Fyers app-id and secret-id shapes that `broker/fyers/api/auth_api.py` lines 46 to 50 hash together into `appIdHash`. They are already public in this PR diff and in your fork's git history, so closing the PR does not undo the exposure. Please delete or regenerate that app at the Fyers developer portal today. The SEBI static-IP requirement limits what someone else can do with them from another machine, but it is not a substitute for revocation.

## On the PR itself

What is right: the diff is small and trivially revertible, and you correctly left `ENV_CONFIG_VERSION` and the `OPENALGO_PLACEHOLDER_APP_KEY` / `OPENALGO_PLACEHOLDER_API_KEY_PEPPER` values alone, so the first-boot secret rotation in `utils/env_check.py` still works. Everything else in the change has to be reverted, for the reasons below.

**1. `.sample.env` lines 10 to 11: restore `YOUR_BROKER_API_KEY` and `YOUR_BROKER_API_SECRET`.** Those literals are a contract, not decoration. They are the substitution targets in `install/install.sh:784-785`, `install/install-docker.sh:270-271`, `install/install-multi.sh:364-365`, `install/install-docker-multi-custom-ssl.sh:834-835`, `install/docker-run.sh:272-273` and `282-283`, and `install/docker-run.bat:225-226`. With the literals gone, every one of those matches nothing, `sed` and the PowerShell `-replace` both exit successfully, and the finished `.env` carries your Fyers app id instead of the key the user typed at the install prompt. Two guards also stop firing: `docker-build.sh:55` (`grep -q "YOUR_BROKER_API_KEY" .env`) no longer warns about an unconfigured file, and `broker/tradejini/api/auth_api.py:18-20` and `41-45` treat any value beginning with `YOUR_` as not configured, so a Tradejini user gets a bare 401 from the broker instead of the clear "API key not configured" message.

**2. `.sample.env` line 18: restore `REDIRECT_URL = 'http://127.0.0.1:5000/<broker>/callback'`.** `<broker>` is both a substitution target (`install/install.sh:796`, `install/install-docker.sh:273`, `install/install-multi.sh:343`, `install/docker-run.sh:274` and `284`, `install/docker-run.bat:229`, `install/install-docker-multi-custom-ssl.sh:837`) and a deliberate startup guard: `utils/env_check.py:1301-1322` exits on both the exact default value and on any remaining `<broker>`, specifically so that a plain `cp .sample.env .env` cannot start the app. Hardcoding `/fyers/callback` removes that guard, and the broker name is parsed back out of `REDIRECT_URL` in three more places you may not have seen:

- `blueprints/auth.py:124-133` decides which broker login button the SPA renders, so a Zerodha user installing from this template is shown a Fyers login.
- `blueprints/broker_credentials.py:108-116` sets `current_broker` on the in-app credentials page.
- `utils/env_check.py:1218-1265` picks which broker-specific key-format check runs, so the `a:::b` compound-key validation for flattrade, dhan and 5paisa silently stops running.

**3. `.sample.env` line 1: remove `#Sandeep S#tra123`.** It stays inside the comment so it breaks no parser, but `tra123` reads like a password fragment. If it is one, change that password too.

**4. Why the urgency is higher than a normal template edit.** `install/docker-run.sh:38` and `install/docker-run.bat:32` fetch `https://raw.githubusercontent.com/marketcalls/openalgo/main/.sample.env` at install time. A merge to `main` would push those credentials to every Docker install run worldwide immediately, with no `git pull` involved. `install/update.sh:357-361` and `install/update.bat:210-212` also copy `.sample.env` straight to `.env` when `.env` is missing, substituting only `APP_KEY` and `API_KEY_PEPPER`, never the broker key.

**5. Docs would contradict the file.** `docs/design/28-environment-config/README.md:39-47` reproduces the placeholders verbatim, and `docs/userguide/06-broker-connection/README.md:29` plus `docs/test/BROKER_INTEGRATION_TESTING_GUIDE.md:29` both tell users the callback ends in `/<broker>/callback`. The fix is to revert `.sample.env`, not to edit the docs.

## Process notes

- Never put real values in `.sample.env`. Copy it to `.env`, which is gitignored, and configure there. The in-app broker credentials page (`blueprints/broker_credentials.py`) will write `.env` for you.
- CI has not run on this branch. The only green checks are Snyk, which reported no manifest changes, and the cubic summarizer. No lint, no backend test and no frontend build has looked at it. Even a full CI run would not have caught this: nothing in `test/` asserts the placeholder contract. If you want to contribute something genuinely useful here, a small `test/test_sample_env_placeholders.py` asserting that `.sample.env` still contains `YOUR_BROKER_API_KEY`, `YOUR_BROKER_API_SECRET` and `<broker>` would close that gap, and it must be added to the pytest list in `.github/workflows/ci.yml` (around line 47) or it will never run in CI.
- Commit and PR titles follow Conventional Commits per `CLAUDE.md`. "fyres env" is neither conventional nor the broker's spelling.

## What to do with this PR

Please close it. There is no fix in it to salvage. If there is a real Fyers option-chain defect behind the branch name `fyers_oc_fix`, open a fresh branch off current `main` containing only that change, with a `fix:` title, a description of the defect, and how you verified it. I will review that one.

marketcalls:
Closing per the review above. Nothing in this diff is salvageable, and the Fyers app credentials in it need revoking at the Fyers developer portal regardless of this PR being closed, since they are already public in the diff and in the fork's history.

If there is a real Fyers option-chain defect behind the branch name fyers_oc_fix, please open a fresh branch off current main containing only that change, with a fix: title and a description of the defect. I will review that one.

Issue #1857, opened 2026-08-22
security(docker): Pin base images by digest


marketcalls:
## Summary

The multi-stage Dockerfile selects base images only by mutable tags.

## Evidence

- `Dockerfile:2`, `Dockerfile:17`, and `Dockerfile:26` use tag-only image references.
- The stages use official Node and Python images but do not record immutable digests.

## Why this matters

A mutable base tag can resolve to different content between builds, weakening reproducibility and supply-chain review. Digest pins make the exact base content explicit.

## Scope

Pin every base image to a verified digest while retaining a readable tag. Confirm multi-stage aliases and supported architectures still build, and document the safe update procedure.

## Acceptance criteria

- [ ] All `FROM` references include `tag@sha256:digest`.
- [ ] Digests are taken from the official image registry/manifests.
- [ ] The production target builds for the project's supported architecture(s).
- [ ] A short comment or maintainer note explains how pins are refreshed.
- [ ] No credentials or registry tokens are committed.

## Verification

```powershell
rg -n "^FROM " Dockerfile
docker build --target production -t openalgo-security-check .
```

## Security guidance

Do not post or use real API keys, access tokens, OTPs, passwords, or broker credentials. Use obvious test values such as `test-token-not-real` and keep regression tests fully local.

## First-time contributor notes

- Difficulty: Medium
- Estimated time: 2-4 hours
- Broker credentials required: No
- You will learn: container reproducibility, OCI digests, and supply-chain maintenance.
- Please comment on this issue before starting so contributors can avoid duplicate work.

srajbr:
I am picking this up

PR #1872, opened 2026-08-23
chore: pin docker base images to immutable digests

srajbr:
Pins all three `FROM` base images in the multi-stage `Dockerfile` to an immutable `sha256` digest, alongside the existing tag, so a build always resolves the exact same base image content instead of whatever a mutable tag currently points to.

- `python:3.12-bullseye` (python-builder stage) -> pinned
- `node:22-bullseye-slim` (frontend-builder stage) -> pinned
- `python:3.12-slim-bullseye` (production stage) -> pinned

Each digest was taken directly from the official image's live manifest via `docker buildx imagetools inspect <tag>` and matched against the correct tag (the production stage keeps the smaller `slim-bullseye` variant it always used, only the digest is new).

Also adds a short comment above the builder stage documenting how to refresh a pin after a deliberate base-image version bump.

## Related Issues

Closes GitHub issue #1857

## Testing

- `rg -n "^FROM " Dockerfile` - all three lines are `tag@sha256:digest`
- Verified each digest against the registry with `docker buildx imagetools inspect <tag>`
- `docker build --target production -t openalgo-security-check .` - builds successfully
- `docker compose up --build -d` - container reaches `healthy`; `curl http://127.0.0.1:5000/auth/check-setup` and `/` both return `200`; startup logs show all migrations completing with no errors
- `docker compose down` - clean teardown

## Checklist

- [x] Application starts without errors (verified via `docker compose up`)
- [x] All existing features unaffected (no application code changed, Dockerfile only)
- [x] No linting errors applicable (no Python/TS code changed)
- [x] No credentials or registry tokens committed
- [x] Production target builds and runs for the project's supported architecture
- [x] New test passes locally and is registered in the CI test list
- [x] Dependabot config is valid YAML


marketcalls:
Thanks for picking this up. The digest work itself is correct, and I checked it rather than taking the description on trust.

I resolved all three tags against registry-1.docker.io. Each digest in the patch is the current resolution of its own tag, each is an OCI image index rather than a platform-specific manifest, and each carries both linux/amd64 and linux/arm64 entries. That last point is load bearing: ci.yml builds both arches natively (ubuntu-latest and ubuntu-24.04-arm), and a platform-manifest digest would have broken the arm64 leg. The slim and non-slim python variants are not cross-wired. Because each digest equals what its tag resolves to today, a build from this branch is byte identical at the base layer to a build from main today, so the change cannot regress the image.

Four things before I merge.

1. Documentation. `docs/design/11-docker/README.md` lines 88, 103 and 111 embed a verbatim copy of the Dockerfile, and those three FROM lines were byte exact against `Dockerfile:2`, `:17` and `:26` on main. They are wrong after this PR. `docs/docker/DOCKER_BUILD_GUIDE.md` lines 70, 76 and 81 name the same three tags. CLAUDE.md treats `docs/` as the single source of truth, so please either update those lines or, better, replace the embedded Dockerfile block in the design doc with a pointer to the real file. That block has already drifted in other ways (it still shows `uv pip install gunicorn eventlet>=0.40.3` where the Dockerfile has `"gunicorn>=25.0,<26" eventlet`, and it omits the chromium and fonts-liberation install), so a pointer is the durable fix.

2. Refresh path. `.github/dependabot.yml` declares pip, npm at `/`, npm at `/frontend` and github-actions. There is no docker ecosystem and no renovate config anywhere in the repo, so nothing will ever bump these digests. A pin with no refresh path is a permanent maintenance liability. Please add it in this PR:

```yaml
  - package-ecosystem: "docker"
    directory: "/"
    schedule:
      interval: "weekly"
      day: "monday"
    commit-message:
      prefix: "deps(docker):"
```

Dependabot's docker ecosystem updates `FROM image:tag@sha256:...` and moves tag and digest together, which also closes point 3.

3. A guard against tag and digest desync. Once a digest is present, BuildKit resolves by digest and the tag is decorative. If someone later bumps the production stage to a bookworm tag and forgets the digest, the build silently keeps using the old bullseye image and nothing notices. There is no hadolint hook in `.pre-commit-config.yaml`, no Dockerfile lint in either workflow, and no test in `test/` reads the Dockerfile. Please add `test/test_dockerfile_pins.py` asserting that every line matching `^FROM ` carries `@sha256:` followed by 64 hex characters, and register it in the `Run CI-safe tests` step of the `backend-test` job in `.github/workflows/ci.yml`. It needs no credentials and runs in milliseconds.

4. Description text. Please strip the em dashes and arrow glyphs from the PR body. I counted five U+2014 and three U+2192. CLAUDE.md bans em dashes, en dashes and icons in PR descriptions as well as in source. The cubic block at the bottom also embeds SVG button imagery and calls this "Linear #1857"; it is GitHub issue marketcalls/openalgo#1857 in this repo. The lines you added to the Dockerfile are clean ASCII, so this is only the description.

On validation, so it is on the record: CI has not run on this branch. Fork PRs need maintainer approval to run workflows here, and separately the `docker-build` job is gated at `ci.yml:204` on `github.event.pull_request.head.repo.full_name == github.repository`, so it is skipped for a fork PR even after approval. The one job that builds the Dockerfile will first run on the post-merge push to main. The green Snyk checks reported no manifest changes and never looked at the Dockerfile. I am comfortable on the build because of the digest equivalence above, but nothing automated has checked this branch.

One thing for a separate issue, not for this PR. The runtime base you pinned was last built upstream on 2025-07-22 (org.opencontainers.image.created in the index annotations) and is Python 3.12.11 on Debian 11. Bullseye variants are no longer rebuilt, so the pin freezes a base that is already more than a year stale, and `apt-get update` at `Dockerfile:31` will start failing once bullseye moves to archive.debian.org, which breaks the build outright. The mutable tag was the smaller half of the supply-chain problem here. Please open a follow-up issue for a bookworm base bump so it does not get lost behind the hash.

srajbr:
Follow-ups from this review, none of which belong in this PR's diff since they are not caused by or entangled with the digest pinning change. Listing them here so they do not get lost.

## 1. docs/docker/docker.md describes a different, older Dockerfile

This file's embedded Dockerfile (line 12 onward) is `python:3.11-slim` with plain `pip install -r requirements-nginx.txt`, no `uv`, and no multi-stage build. It predates the current setup entirely and is unrelated to this PR. Needs its own rewrite, not a small edit.

## 2. docker-compose (v1) vs docker compose (v2) command syntax

`docs/design/11-docker/README.md` (Docker Compose Commands section, lines 252-261) and `docs/docker/DOCKER_BUILD_GUIDE.md` (around 40 occurrences throughout) both use the hyphenated `docker-compose` v1 standalone binary syntax. Modern Docker ships `docker compose` as an Engine/Desktop plugin. Not functionally broken on systems that still carry the v1 compatibility shim, but stale style across a large surface. Suggest a dedicated pass rather than folding it into this PR.

## 3. Read-only .env mount guidance is a previously-fixed critical bug, still present in docs

`docs/releases/version-2.0.0.8-released.md` documents that mounting `.env` as `:ro` broke the app's key-rotation write (atomic `.env.tmp`, `EACCES`, gunicorn crash loop) and that the fix dropped `:ro` from every `.env` mount, including `docker-compose.yaml`. That guidance was never cleaned out of the docs. It is still present in:

- `docs/design/11-docker/README.md:235` (Build Commands docker run example)
- `docs/design/11-docker/README.md:352` (Security Considerations table)
- `docs/docker/DOCKER_BUILD_GUIDE.md:58` (Manual Build with Docker CLI example)
- `docs/docker/DOCKER_BUILD_GUIDE.md:483` (Runtime Security list)
- `docs/docker/DOCKER_SCRIPTS_ANALYSIS.md` (multiple examples)
- `docs/docker/README.md:135`
- `docs/docker/DOCKER_NUMBA_FIX.md:118`
- `docs/docker/UPDATES_SUMMARY.md` (multiple examples)

Anyone copying these commands today reintroduces a bug that was already fixed in code. Flagging this as higher priority than the other items here since it is an active regression risk, not just staleness.

## 4. start.sh documentation is out of date

`docs/design/11-docker/README.md` describes `start.sh` as a "246-line entrypoint" (lines 129, 137, 148) with a 6-step overview and a comparison table (line 212). The real file is now 341 lines with 7 sections, including a `PRE-FLIGHT: COMPROMISED-KEY DETECTION` step that is not mentioned anywhere in the doc.

## Question: bookworm base image bump, issue now or wait for Dependabot?

For the earlier note about the `python:3.12-slim-bullseye` base being stale upstream (last rebuilt 2025-07-22, Debian 11 past its rebuild cutoff), should we open the follow-up issue now, or wait and let the new Dependabot docker ecosystem entry surface it?

Our read is that Dependabot will not solve this on its own. Its docker ecosystem refreshes the digest for the same tag, it does not propose switching `python:3.12-slim-bullseye` to `python:3.12-slim-bookworm`, since that is a different tag family rather than a version bump within one. Left alone, Dependabot will keep the digest current on a codename that is already past its rebuild cutoff without addressing the actual problem, `apt-get update` eventually failing once bullseye moves to archive.debian.org. We are leaning toward opening the issue now rather than waiting. Let us know if that matches what you had in mind.

marketcalls:
Close, but not merging yet. Three of the four items are in good shape. The guard itself needs a two line fix first, and there is one edit from my last comment that was skipped.

Everything below was checked against the blobs at 4cd2fe15 from the contents API, not the patch.

Docs. Confirmed. The embedded Dockerfile block is gone from `docs/design/11-docker/README.md` and replaced with a pointer, the duplicated Compose block went with it, and both links resolve: `../../../Dockerfile` and `../../../docker-compose.yaml` from `docs/design/11-docker/`, `../../Dockerfile` from `docs/docker/`. Both targets exist at the repo root. `git grep -nE "^FROM (python|node)" -- docs/` at your head returns only `docs/docker/docker.md:12`, the unrelated older guide you correctly scoped out. Your three diagram edits are all 79 columns, matching their neighbours, so nothing is knocked out of alignment.

Dependabot. Confirmed. `docker` at `/`, weekly Monday, prefix `deps(docker):`. Parses, and it is the only delta on that file against main.

Test registration. Confirmed, and it really runs. I pulled the backend-test log for run 32747885113: both cases appear as PASSED at 74 and 76 percent, 55 passed overall. The merge kept main's `test_multi_option_greeks_regression.py` and `test_history_format.py` registrations, so no other PR's entry was lost.

The problem is the guard itself. I mutated the Dockerfile rather than reading the test, six times:

1. Drop the node digest: FAILS with `['node:22-bullseye-slim']`. Good, that is the case you tested.
2. Retarget production to `python:3.12-slim-bookworm` while leaving the bullseye digest attached: PASSES. That is verbatim the scenario your module docstring says the test guards, "someone retargeting a stage to a new tag without updating the digest, which would leave the build silently pinned to the old, stale image". It does not.
3. Append an unpinned stage written as lowercase `from python:3.13-slim AS extra`: PASSES. Dockerfile instruction names are not case sensitive, so that is a real unpinned base waved straight through.
4. The same stage indented two spaces: PASSES. Leading whitespace before an instruction is also accepted by the parser.
5. `FROM production AS smoketest`: FAILS with `['production']`. A legitimate stage to stage FROM breaks the test.
6. `FROM --platform=$BUILDPLATFORM node:...@sha256:<valid digest>`: FAILS with `['--platform=$BUILDPLATFORM']`.

3 and 4 are the ones that matter. A supply chain guard that silently passes an unpinned base is worse than no guard, because it creates the belief the case is covered. One line fixes all four:

```python
FROM_LINE_RE = re.compile(r"^[ \t]*FROM\s+(?:--\S+\s+)*(\S+)", re.MULTILINE | re.IGNORECASE)
```

then drop any captured ref that matches a stage alias declared by an earlier `AS <name>`, which clears 5 and 6. For 2, either add the check or correct the docstring so it claims only what the code asserts. While you are in the file, add `encoding="utf-8"` to the `open()`. That Dockerfile is not ASCII decodable, six stray dash characters in comments that predate you, so the read depends on the locale. I tested it: a plain `LC_ALL=C` run on Linux still succeeds because CPython coerces C to C.UTF-8, so it is not urgent, but it costs nothing.

Body. The em dashes, the arrows and the SVG imagery are all gone, so that part is done. Two U+2011 non breaking hyphens remain in the cubic block; those are hyphens rather than dashes and I am not going to chase them. But line 36 still reads "(Linear #1857)". That was an explicit ask last round. It is GitHub issue marketcalls/openalgo#1857 in this repo.

One thing I am flagging rather than blocking on. The `DOCKER_BUILD_GUIDE.md` commit also rewrote the image size table with figures described as measured via `docker history`. That was not requested, and nobody can reproduce it, because `docker-build` is skipped for fork PRs so no image has ever been built from this branch. The one component I could check independently is wrong: I pulled the per arch manifests for the production digest you pinned, downloaded every layer and decompressed them, and they total 127 MB on amd64 and 120 MB on arm64, not the ~115 MB the table now states. Please either drop that hunk or soften it to an estimate with no measurement claim.

A correction to something I would otherwise have told you. I had assumed a Dependabot docker PR would be the first Dependabot PR to reach `docker-build` and would fall over because `secrets.DOCKERHUB_USERNAME` is empty for Dependabot runs. That is wrong. PR #1719 was a Dependabot actions bump, it reached `docker-build`, the metadata step logged `images: /openalgo` with the empty secret, `docker/metadata-action` accepted it and emitted `/openalgo:pr-1719`, and both matrix legs succeeded. So your new docker ecosystem entry will get a real amd64 and arm64 build on every digest bump. That is a bonus, not a hazard.

Digests re-checked today: all three still equal exactly what their tags resolve to right now, and all three are multi arch OCI indexes with a `linux/arm64` entry, so a build from this branch is still byte identical at the base layer to a build from main. The branch is 5 ahead and 0 behind, the two dot diff against main is exactly your six files, and every touched blob is LF with no CR bytes.

On bookworm you are right and I was loose. Dependabot's docker ecosystem refreshes the digest for the tag that is written; it will not move `3.12-slim-bullseye` to `3.12-slim-bookworm`. Left alone it keeps a dead codename permanently current. Open that issue now. The real deadline is `apt-get update` at `Dockerfile:31` failing once bullseye moves to archive.debian.org, which breaks the build outright.

Your other three follow-ups are correctly scoped out. Number 3, the `:ro` `.env` mount guidance, is the highest priority of them, since anyone copying those commands reintroduces a bug already fixed in code. Please open that one separately as well.

Fix the regex and the docstring, fix the Linear attribution, and soften or drop the size table, and I will merge. Thanks for the fast turnaround on the rest of it.

srajbr:
Thanks for the detailed re-check. All four blocking items are addressed.

**Test guard regex.** Rewrote `FROM_LINE_RE` to `r"^[ \t]*FROM\s+(?:--\S+\s+)*(\S+)(?:\s+AS\s+(\S+))?"` with `re.IGNORECASE`, and added stage-alias tracking so a `FROM <earlier-stage> AS <name>` reference is excluded from the digest check. Re-ran all six of your mutations against the fixed regex in a scratch copy: the two real gaps (lowercase `from`, indented `FROM`) now correctly fail, the two false positives (stage-to-stage `FROM`, `--platform=...`) now correctly pass, and the case you'd already confirmed working (dropped digest) still fails as before.

**Module docstring.** Narrowed it to only claim what the code actually asserts: presence of a digest, not that the digest matches what the tag currently resolves to on the registry. Verifying that would need a network call, which an offline CI-safe test shouldn't make. Also added `encoding="utf-8"` to the `open()` call.

**Linear attribution.** Fixed in the description directly. For context on why it was there: it was inside the "Summary by cubic" section Cubic auto-appends below the description (marked with an HTML comment), not something either of us wrote by hand. It used a Linear-style backlink even though this repo links issues directly on GitHub.

**Image size table.** Rolled back to the original figures rather than softened, since neither of us can currently produce a reproducible measurement (docker-build skips fork PRs). Diffed against main to confirm the section is now byte-identical to what it was before this PR touched it. For what it's worth, the size increase I was seeing locally does look real, not a measurement artifact: adding `chromium` and `fonts-liberation` for Kaleido chart export is what's driving it, on the order of several hundred MB by itself. The old figures predate that install and never accounted for it. Opening a separate issue to get an accurate, reproducible number and update the table properly, noted below.

Also merged main in twice more since your last pass to pick up the CI matrix changes and new test registrations, most recently at the current head. Full CI-safe test list passes locally, 74 tests.

Three follow-up issues from this review are being opened separately, not part of this PR:

- Bookworm base image bump, since Dependabot's docker ecosystem will keep bullseye's digest current without ever moving off a codename that's past its rebuild cutoff: #1900
- The `:ro` `.env` mount guidance still present across several docs, reintroducing a bug already fixed in v2.0.0.8: #1901
- The image size table itself, to get an accurate, reproducible number and update `DOCKER_BUILD_GUIDE.md` properly: #1902

The other follow-ups from earlier (docs/docker/docker.md describing an older pre-uv Dockerfile, and the docker-compose v1/v2 command syntax used throughout the docs) are still correctly scoped out of this PR per your last review, not forgotten, just not urgent enough for their own issue yet.

marketcalls (commented):
Review score: 82/100 at `4144090e7770bc80cf2d5e20925807d6cd91769f`. Changes are needed before approval and merge.

- `Dockerfile:5`: The three immutable pins still target Bullseye and conflict with the Trixie Docker fix now on main. Resolving the conflict by retaining these FROM lines would discard the base-image correction from merged #2004. Rebase and pin the current Trixie builder/frontend/runtime images to verified multi-platform digests; preserve amd64/arm64 support and rerun both image builds and startup smoke checks.

Validation: Both new offline digest-format tests passed; they do not validate image contents or build availability. Full Docker, Dependabot, workflow and documentation diff reviewed. Current head conflicts with main.

srajbr:
merge conflict resolved

PR #1873, opened 2026-08-23
fix: validate option offsets in order schemas

### LINKED ISSUE #1820: python: Reuse option-offset validation in order request schemas
## Summary

Option offset validation is applied to one request family but omitted from order schemas accepting the same field, so invalid values pass some API boundaries.

## Evidence

- `restx_api/data_schemas.py:22` defines the reusable validator.
- `restx_api/data_schemas.py:176` applies it.
- `restx_api/schemas.py:214` and `restx_api/schemas.py:253` declare comparable fields without it.

## Scope

Apply the existing validator to every matching order request field and add focused schema tests. Do not duplicate validation logic.

## Acceptance criteria

- [ ] All relevant schemas use the shared validator.
- [ ] Valid boundary values remain accepted.
- [ ] Invalid values produce a clear schema error.
- [ ] Tests cover every updated schema.

## Verification

```powershell
uv run pytest test/test_order_schema_validation.py -v
uv run ruff check restx_api/data_schemas.py restx_api/schemas.py test/test_order_schema_validation.py
```

- Difficulty: Easy
- Estimated time: 1-2 hours
- You will learn: Marshmallow validation reuse and API contracts.

siddharthg2309:
## Summary

- Reuse the shared option-offset validator in single options orders and multi-order legs.
- Reject invalid offsets at the schema boundary before symbol resolution or order processing.
- Preserve valid lowercase offsets such as `atm` and `itm1`.
- Add focused coverage for single orders, multi-order legs, and nested multi-order validation errors.
- Register the backend regression test in the CI-safe test suite.

## Related issue

Closes #1820

## Behavior change

Offsets outside `ATM`, `ITM1-ITM50`, and `OTM1-OTM50` are now rejected with a `400` validation error before strike resolution. This prevents malformed or out-of-range offsets from resolving to an unintended strike.

## Verification

- `env -u LOG_FORMAT uv run pytest test/test_order_schema_validation.py -v --timeout=60`, 23 passed
- Current CI-safe backend test suite, 95 passed
- `uv run ruff check restx_api/data_schemas.py restx_api/schemas.py test/test_order_schema_validation.py`, passed
- `uv run ruff format --check test/test_order_schema_validation.py`, passed
- `git diff --check`, passed

## Contribution checklist

- [x] One focused bug fix
- [x] Focused regression coverage for every updated schema
- [x] Backend test is registered in CI
- [x] Ruff lint passes for changed Python files
- [x] No UI changes, screenshots are not applicable
- [x] No credentials, generated artifacts, or unrelated changes included


marketcalls:
Thanks for this. The bug is real and I reproduced it on main. `restx_api/schemas.py:218` and `:253` declare `offset` as a bare `fields.Str(required=True)`, so `ITM0` reaches `services/option_symbol_service.py:526`, resolves to `atm_index - 0` which is the ATM strike, and a live order goes out at a strike the caller did not ask for. `ITM-3` inverts the direction the same way, and `ITMX` blows up in `int(offset[3:])` and comes back as a 400 echoing `invalid literal for int() with base 10` through the `except ValueError` at `services/option_symbol_service.py:833`. Reusing `validate_option_offset` is the right fix: it matches what `OptionSymbolSchema` already enforces at `restx_api/data_schemas.py:177`, what the Flow executor enforces at `services/flow_executor_service.py:784`, and what `docs/api/order-management/optionsorder.md:176` already claims we do.

I applied the branch locally. `uv run ruff check` is clean on both files, the 20 tests pass with the fix and 10 of them fail without it, and the whole file runs in 1.2 seconds. I also walked every internal caller that bypasses the REST schema (`services/flow_openalgo_client.py:445`, `services/pending_order_execution_service.py:148`, `services/action_center_service.py`, `mcp/mcpserver.py:684`, `frontend/src/components/flow/panels/ConfigPanel.tsx:2699`) and none of them produces an offset outside ATM / ITM1-50 / OTM1-50, so nothing existing breaks.

Four things before I merge.

1. `.github/workflows/ci.yml`, the `backend-test` step (lines 43-50). Add `test/test_order_schema_validation.py` to the CI-safe list. As it stands the regression test never executes in CI, so the exact regression it guards against could ship green. Worth knowing: CI has not run on this branch at all, since fork workflows need maintainer approval, so my local run is currently the only thing that has validated it.

2. `test/test_order_schema_validation.py`. Add a case that loads `OptionsMultiOrderSchema`, not only the leg schema. `restx_api/schemas.py:302` nests the leg schema inside a `fields.List`, so a bad leg surfaces to the client as `{"legs": {1: {"offset": [...]}}}` keyed by the failing leg index, which `restx_api/options_multiorder.py:156` then wraps as `{"status": "error", "message": "Validation error", "errors": {...}}`. The flat `{"offset": [...]}` your test asserts is never what an API consumer parses, so the error contract that actually matters is unpinned.

3. `test/test_order_schema_validation.py`. Add a lowercase case (`atm`, `itm1`). `restx_api/data_schemas.py:26` uppercases only for comparison and neither schema normalizes, so the raw lowercase string is what gets stored and echoed back at `services/place_options_order_service.py:300`. That behaviour is fine, it just needs to be locked in.

4. PR description. Two em dashes in the Verification section. The repo rule is no em or en dashes in any generated text, PR descriptions included. Please replace them with commas or full stops. The diff itself is clean ASCII, so this is only the body.

Two notes that are not blockers.

The shared validator uses `re.match` with `$` (`restx_api/data_schemas.py:31-32`), and in Python `$` also matches before a single trailing newline. I confirmed that `offset="ITM1\n"` is accepted and stored verbatim while `offset="ATM\n"` is rejected, so the two branches of the rule disagree. It is harmless downstream (`int("1\n")` is 1, and everything re-uppercases), but `restx_api/options_order.py:119` logs the raw offset into a single-line text log (`utils/logging.py:353`), so the stray newline reaches the log file. The Flow copy of the same rule uses `re.fullmatch` and does not have this hole. Changing `$` to `\Z` closes it. That file is outside your diff, so a separate PR if you want it. While you are there, `underlying` and `strategy` on that same log line are still unvalidated `fields.Str`, so this PR narrows but does not close the log-forging surface on that endpoint.

Worth adding to the PR body: with an explicit `strike_int`, `services/option_symbol_service.py:790` uses `calculate_offset_strike`, whose `int(offset[3:])` has no upper bound, so an offset above 50 previously resolved to a real strike and placed a real order. After this change it is a 400 at the schema. The tightening is correct and matches the documented contract, but it is a behaviour change for anyone scripting deep offsets and should go in the release notes.

marketcalls (commented):
Reviewed head `b49b96432d8d1a4e0f9a88d5a8042f483dea4c97`: 87/100. Please address the following before merge:

- `.github/workflows/ci.yml:50`: Required frontend-test (22) check is failing. The PR is not merge-ready despite focused backend tests passing. Rerun the failed check; if reproducible, fix the unrelated StrategyBuilder test failure before merge. Do not change valid offset logic to address a frontend-only failure. Evidence: https://github.com/marketcalls/openalgo/actions/runs/32832502298/job/97754087706; StrategyBuilder.test.tsx:1495 expects no Margin Req. element after identity reset.

PR #1874, opened 2026-08-23
fix: validate history intervals per broker

### LINKED ISSUE #1818: python: Align history validation with broker-advertised intervals
## Summary

The supported-intervals endpoint can advertise canonical intervals that the history request schema rejects before the request reaches a broker.

## Evidence

- `services/intervals_service.py:53` classifies broker map keys ending in `m` or `h` as supported intervals.
- `broker/zerodha/api/data.py:158` advertises `60m`, and `broker/deltaexchange/api/data.py:217` advertises `6h`.
- `restx_api/data_schemas.py:72` currently accepts intervals only through `30m` and `4h`.

## Scope

Make the history schema accept the broker-advertised canonical values `60m` and `6h`, or derive the accepted set from a single canonical interval definition. Add a focused contract test. Do not change broker-specific interval mappings.

## Acceptance criteria

- [ ] `60m` and `6h` pass `HistorySchema` validation.
- [ ] All currently valid interval values remain valid.
- [ ] An unsupported interval still produces a schema validation error.
- [ ] A credential-free contract test proves that advertised canonical intervals are requestable.

## Verification

```powershell
uv run pytest test/test_intervals_history_contract.py -v
uv run ruff check restx_api/data_schemas.py services/intervals_service.py test/test_intervals_history_contract.py
```

- Difficulty: Easy
- Estimated time: 1-2 hours
- You will learn: Marshmallow schemas, API/service contracts, and parameterized pytest coverage.

siddharthg2309:
## Summary

Validate history intervals against the selected broker’s supported `timeframe_map`.

This allows intervals such as `60m` or `6h` only when that broker actually supports them. Unsupported intervals now return a clear `400` before a broker provider is called.

Database history requests (`source="db"`) keep their existing behavior.

## Related issue

Closes #1818

## Changes

- Move interval validation from the global request schema to the broker-aware history service.
- Return a clear error listing the selected broker’s supported intervals.
- Add contract coverage for every broker’s declared timeframe map.
- Cover unsupported broker intervals and the database-history path.
- Update the history-service test stub with its supported `5m` interval.

## Verification

- `env -u LOG_FORMAT uv run pytest ...` CI-safe backend suite: 138 passed
- Focused history and interval tests: 73 passed
- Ruff, formatting, and whitespace checks passed


marketcalls:
Thanks for taking on #1818. The diagnosis is right as far as it goes, and I verified the premise on main: ten brokers put `60m` in `timeframe_map` (zerodha, upstox, iiflcapital, compositedge, iifl, ibulls, wisdom, jainamxts, fivepaisaxts, rmoney) and deltaexchange puts `6h` there, while `restx_api/data_schemas.py:62-95` rejects both. The mismatch is user visible too: `frontend/src/pages/CustomStraddle.tsx:521-532`, `IVChart.tsx:319` and `StraddleChart.tsx:507` populate their interval dropdowns straight from `/api/v1/intervals`, so a Zerodha user really can pick a value the history API then answers with 400. Your contract test does fail before the patch and pass after it, and the diff is ruff clean.

The problem is the layer the fix sits at. `HistorySchema` is a single global allowlist, but "which intervals exist" is a per-broker fact. Widening the global list grants `60m` and `6h` on the roughly two dozen brokers that do not support them, and on two of them that is a regression, not a no-op.

1. `broker/groww/api/data.py:336-340`. `get_history` reads `if timeframe in self.timeframe_map: ... else: logger.warning("Unrecognized timeframe ..., defaulting to daily"); interval_minutes = "1440"`. Groww's map (lines 112-124) has neither `60m` nor `6h`. So after this patch `POST /api/v1/history` with `interval=60m` on Groww stops returning 400 and starts returning 200 with daily candles, with nothing in the payload saying the interval was substituted. A strategy that asked for hourly bars then computes on daily bars. This is the same failure the project already fixed once and documented at `broker/fivepaisa/api/data.py:114-120`: "'3m' used to be advertised here; 5Paisa has no 3-minute candle, and map_interval quietly fell back to daily, so a 3m request returned daily bars."

2. `broker/definedge/api/data.py:640-648` returns an empty DataFrame for an unknown interval, so the same request becomes 200 with `"data": []`. `broker/hdfcsecurities/api/data.py:450-456` states the house position on that: "Raises rather than returning an empty DataFrame: an empty frame reads as 'no trades in that window' to every caller downstream (backtests, indicators, charts), which would silently produce wrong results instead of a visible, actionable failure."

3. On the remaining brokers (angel, fyers, aliceblue, flattrade, indmoney, dhan, fivepaisa and the rest) the request now reaches the adapter and comes back as 500 from `services/history_service.py:138-140`, with a full traceback written to `log/errors.jsonl` on every call. That is a worse client contract than the 400 it replaces.

What I would rather see: validate the interval against the resolved broker instead of against a global list. In `services/history_service.py:get_history_with_auth`, immediately after `data_handler` is constructed (around line 125), reject an interval that is not a key of `data_handler.timeframe_map` and return 400 naming that broker's supported keys. Every broker in the tree defines `timeframe_map`, and the five with no history API define it as `{}` on purpose (`broker/hdfcsecurities/api/data.py:67`, `broker/kotak/api/data.py:34`, `broker/motilal/api/data.py:76`, `broker/paytm/api/data.py:66`, `broker/pocketful/api/data.py:112`), so the check is uniform and needs no special cases. Apply it only on the `source="api"` path: `services/history_service.py:252-260` already routes `source="db"` to Historify, and `database/historify_db.py:parse_interval` computes arbitrary intervals from stored 1m and D bars, so that path must stay unconstrained. That closes #1818 by construction, advertised equals requestable for every broker, and it also removes the Groww and Definedge silent-substitution exposure for the intervals that can already reach them today.

If you would rather keep the allowlist approach, then these have to change as well.

- `restx_api/data_schemas.py`: `25m` is advertised by `broker/dhan/api/data.py:152` and `broker/dhan_sandbox/api/data.py:133`, and `4m` by `broker/indmoney/api/data.py:228`. `services/intervals_service.py:69-71` buckets every key ending in `m` into the advertised minutes list, so these are the identical defect and are still rejected after your patch. Either add them or say why they are out of scope for #1818.
- `services/indicator_service.py`: `6h` is missing from `_BARS_PER_DAY` (line 295) and `_INTERVAL_MINUTES` (line 475), and from `mcp/mcpserver.py:_BARS_PER_DAY` (line 1775). The comment at `services/indicator_service.py:314-317` spells out the consequence of a missing entry: the window falls back to 75 bars per day and every indicator returns null. Accepting `6h` at the API boundary without those entries reproduces that.
- Also worth reading before you settle on the shape: `broker/fivepaisa/api/data.py:658-672` explains why `1h` is the canonical public spelling for hourly here and why accepting `60m` too means advertising one timeframe twice, once under minutes and once under hours. Zerodha (`broker/zerodha/api/data.py:150-162`), Upstox (`broker/upstox/api/data.py:53-73`) and IIFL Capital (`broker/iiflcapital/api/data.py:439-450`) already carry both keys, so for them `1h` works today and `60m` adds only a second name. The brokers that genuinely need help are the XTS family (compositedge, fivepaisaxts, ibulls, iifl, jainamxts, rmoney, wisdom), which carry `60m` and no `1h`; adding `"1h": "3600"` to those maps would fix hourly there under the canonical name. `6h` on deltaexchange is the one case with no canonical alternative.

On tests:

- `test/test_intervals_history_contract.py` stubs a fabricated broker (`_BrokerData.timeframe_map` hardcoded to 30m, 60m, 4h, 6h, D) and monkeypatches `intervals_service.import_broker_module`, so it can only assert the values typed into the fixture. It passes today even though dhan advertises `25m` and indmoney advertises `4m`. Iterate the real `broker/*/api/data.py` `timeframe_map` dicts so the test fails whenever a broker advertises something the API rejects.
- Add a case for the direction this patch actually changes: an interval the resolved broker does not support must still fail cleanly, and must not come back as 200 with substituted or empty data.
- The file is not in the CI-safe list at `.github/workflows/ci.yml:44-53`, so it will never run. Add it there.

Two smaller items:

- `services/intervals_service.py`: please drop the `typing` import cleanup. `F401` is in the ruff ignore list at `pyproject.toml:190`, so it was not lint driven, and the checklist on this PR says no unrelated changes are included.
- The PR body uses em dashes. House style is plain text with no em or en dashes anywhere, PR descriptions included. A comma, colon, parentheses or a full stop instead.

No migration is needed, nothing here touches a schema.

One last thing on the checks: CI has not run on this branch. The only green marks are Snyk and the cubic reviewer, so no ruff run, no backend tests and no frontend build have validated it. Please run the CI-safe pytest list plus your new file locally before the next push.

marketcalls:
Please prioritise this one. Of everything still open in this batch it has the highest real impact, because the failure mode is wrong data returned as a success rather than an error.

Score is **75**, up from 66, and CI is finally green on all 11 checks. The redesign is right: moving interval validation from a global allowlist to a per-broker `timeframe_map` check is the correct layer, and it fixes the Groww problem I raised, where `60m` silently substituted daily candles. Two things block it.

## 1. Deleting the allowlist removed the only interval validation on a second endpoint

This is the blocker. I verified it against your head `0fd6c25e` rather than inferring it.

On main, `restx_api/data_schemas.py:62` reads:

```python
    interval = fields.Str(
        required=True,
        validate=validate.OneOf([ ... ]),
```

At your head it reads:

```python
    interval = fields.Str(required=True)  # Validated against the resolved broker's timeframe_map
```

That comment is true for `/api/v1/history`, which you did wire to the per-broker check. It is **not** true for `/api/v1/ticker/<symbol>`, which reuses the same schema and then goes straight to the broker:

- `restx_api/ticker.py:167-171`: `from .data_schemas import HistorySchema` / `history_schema = HistorySchema()` / `history_data = history_schema.load(ticker_data)`
- `restx_api/ticker.py:211-214`: `data_handler = broker_module.BrokerData(AUTH_TOKEN)` then `df = data_handler.get_history(` with no `timeframe_map` lookup anywhere in between.

So `GET /api/v1/ticker/<symbol>?interval=60m` on Groww now returns HTTP 200 carrying daily candles where it previously returned a clean 400. That is strictly worse than the bug this PR set out to fix: before, one endpoint returned wrong data; now a second one does too, and this one is silent.

Fix either way round: apply the same per-broker resolution in `ticker.py` before calling `get_history`, or give ticker its own schema that keeps the `OneOf`. The first is more consistent with your design.

## 2. `read_text()` without an encoding

`test/test_intervals_history_contract.py:37` is `tree = ast.parse(data_file.read_text())`. On a cp1252 default locale that raises `UnicodeDecodeError` reading `broker/deltaexchange/api/data.py`, which contains UTF-8 arrows. Because the call runs at module import, it takes the whole module down rather than failing one test.

CI is ubuntu so this is invisible there, but `uv run pytest test/ -v` is the documented contributor workflow and it breaks on Windows. Pass `encoding="utf-8"`. Worth doing everywhere you read a repo file.

## Suggested check before you push

Run `grep -rn "HistorySchema" --include="*.py" .` and confirm every consumer is still covered. That is the habit that would have caught `ticker.py` before review: when you delete a validator, the question is not whether the new check is better, it is which callers were relying on the old one.

Ping me when it is up and I will re-review straight away. The layer fix is good work and I want it in.

marketcalls (commented):
Reviewed head `4e18273fc1cf2ff1398b0dac6f8a47bca7613a12`: 89/100. The revised design covers both HistorySchema consumers, and all 74 focused history/ticker/contract tests passed. The previous ticker validation omission and Windows encoding issue are fixed. Rebase the current conflict and rerun the backend workflow on the resulting head; the cached latest-head checks do not include a backend test run.

PR #1875, opened 2026-08-23
fix: reject unsupported history sources

### LINKED ISSUE #1826: python: Reject unsupported direct history-service sources
## Summary

The history service treats `db` specially and routes every other source string to the API path, silently accepting typos.

## Evidence

- `services/history_service.py:219` branches for `db`.
- Other source values fall through to API-backed behavior.

## Scope

Accept only `api` and `db` before selecting a provider. Keep the default unchanged and use mocked tests.

## Acceptance criteria

- [ ] `api` and `db` retain current behavior.
- [ ] Unknown, empty, and wrong-type values return a client error.
- [ ] Invalid values call neither provider.
- [ ] Tests require no credentials.

## Verification

```powershell
uv run pytest test/test_history_service_source_validation.py -v
uv run ruff check services/history_service.py test/test_history_service_source_validation.py
```

- Difficulty: Easy
- Estimated time: 1-2 hours
- You will learn: service contracts and mocked dependencies.

siddharthg2309:
## Summary

Validate the `source` argument in `get_history()` before provider selection.

Only `api` and `db` are accepted. Unknown, empty, and non-string values now return a `400` client error without calling a provider.

## Related issue

Closes #1826

## Changes

- Preserve `source="api"` and the existing default API behavior.
- Preserve `source="db"` and the local database behavior.
- Reject unsupported sources before the API provider, database provider, or rate limiter can run.
- Add focused service tests for valid, default, and invalid source values.
- Remove unused legacy type imports in the same service file as a lint-only cleanup.

## Testing

- [x] `env -u LOG_FORMAT uv run pytest test/test_history_service_source_validation.py -v`
  - 9 passed
- [x] `uv run ruff check services/history_service.py test/test_history_service_source_validation.py`
  - passed
- [x] `git diff --check`
  - passed

## Full-suite note

`env -u LOG_FORMAT uv run pytest test/ -v` was also attempted. Collection is currently blocked by two unrelated existing environment/repository issues:

- `test/test_bot_web.py` imports `get_telegram_bot`, which is not present in `services.telegram_bot_service`.
- `test/test_telegram_startup.py` requires the unavailable `eventlet` package.

Neither failure is in this PR's changed files.

## Checklist

- [x] The change is covered by focused pytest tests.
- [x] Python linting passes for the changed files.
- [x] No credentials or sensitive data are included.

---
## Summary by cubic
Rejects unsupported history sources in get_history to return clear client errors and avoid unintended provider calls. Old: unknown or non-string sources could reach provider selection; new: only 'api' (default) and 'db' are accepted, others return 400 with "Source must be either 'api' or 'db'."

- Invalid sources short-circuit before the API provider, DB provider, and rate limiter run.
- Adds focused tests for valid/default/invalid sources and runs them in CI.

**Migration**
- If a client sends any source other than 'api' or 'db', update it to use 'api' or 'db' to avoid 400 responses.

<sup>Written for commit 5a97bdcb39393caa94e46339f735e4989b8902ac. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1875?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,6 +45,7 @@ jobs:
       - name: Run CI-safe tests
         run: |
           uv run pytest test/test_log_location.py test/test_navigation_update.py \
+            test/test_history_service_source_validation.py \
             test/test_python_editor.py test/test_rate_limits_simple.py \
             test/test_logout_csrf.py test/test_auth_logout.py \
             test/test_auth_resume.py test/test_auth_upsert_multisession.py \
--- a/services/history_service.py
+++ b/services/history_service.py
@@ -242,13 +242,16 @@ def get_history(
         auth_token: Direct broker authentication token (for internal calls)
         feed_token: Direct broker feed token (for internal calls)
         broker: Direct broker name (for internal calls)
-        source: Data source - 'api' (broker, default) or 'db' (DuckDB/Historify)
+        source: Data source - 'api' (broker, default) or 'db' (DuckDB/Historify).
+            Unsupported values return 400 before a provider is called.
 
     Returns:
         Tuple containing:
         - Success status (bool)
         - Response data (dict)
         - HTTP status code (int)
+
+        Unsupported source values return a 400 error before either provider is called.
     """
     if not isinstance(source, str) or source not in {"api", "db"}:
         return (
--- a/services/history_service.py
+++ b/services/history_service.py
@@ -1,6 +1,6 @@
 import importlib
 import time
-from typing import Any, Dict, List, Optional, Tuple, Union
+from typing import Any
 
 import pandas as pd
 
@@ -250,6 +250,13 @@ def get_history(
         - Response data (dict)
         - HTTP status code (int)
     """
+    if not isinstance(source, str) or source not in {"api", "db"}:
+        return (
+            False,
+            {"status": "error", "message": "Source must be either 'api' or 'db'."},
+            400,
+        )
+
     # Source: 'db' - Fetch from DuckDB/Historify database
     if source == "db":
         return get_history_from_db(
--- a/test/test_history_service_source_validation.py
+++ b/test/test_history_service_source_validation.py
@@ -1,0 +1,114 @@
+import pytest
+
+from services import history_service
+
+
+def _history_request() -> dict:
+    return {
+        "symbol": "NIFTY",
+        "exchange": "NSE_INDEX",
+        "interval": "1m",
+        "start_date": "2026-08-01",
+        "end_date": "2026-08-22",
+        "auth_token": "test-token",
+        "feed_token": "test-feed-token",
+        "broker": "test-broker",
+    }
+
+
+def test_api_source_calls_api_provider(monkeypatch):
+    provider_calls = []
+    monkeypatch.setattr(history_service, "_enforce_rate_limit", lambda: None)
+    monkeypatch.setattr(
+        history_service,
+        "get_history_with_auth",
+        lambda *args: provider_calls.append(args) or (True, {"provider": "api"}, 200),
+    )
+
+    result = history_service.get_history(**_history_request(), source="api")
+
+    assert result == (True, {"provider": "api"}, 200)
+    assert provider_calls == [
+        (
+            "test-token",
+            "test-feed-token",
+            "test-broker",
+            "NIFTY",
+            "NSE_INDEX",
+            "1m",
+            "2026-08-01",
+            "2026-08-22",
+        )
+    ]
+
+
+def test_default_source_calls_api_provider(monkeypatch):
+    monkeypatch.setattr(history_service, "_enforce_rate_limit", lambda: None)
+    monkeypatch.setattr(
+        history_service,
+        "get_history_with_auth",
+        lambda *args: (True, {"provider": "api"}, 200),
+    )
+
+    assert history_service.get_history(**_history_request()) == (
+        True,
+        {"provider": "api"},
+        200,
+    )
+
+
+def test_db_source_calls_database_provider(monkeypatch):
+    provider_calls = []
+
+    def fake_database_provider(**kwargs):
+        provider_calls.append(kwargs)
+        return True, {"provider": "db"}, 200
+
+    monkeypatch.setattr(
+        history_service,
+        "get_history_from_db",
+        fake_database_provider,
+    )
+
+    result = history_service.get_history(**_history_request(), source="db")
+
+    assert result == (True, {"provider": "db"}, 200)
+    assert provider_calls == [
+        {
+            "symbol": "NIFTY",
+            "exchange": "NSE_INDEX",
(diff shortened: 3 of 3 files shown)


marketcalls:
Reviewed and verified locally. Decision: merge after one small fix, adding the new test file to the CI-safe pytest list. The code change itself is good as it stands.

## What I verified against main (a1a2dc9)

- The patch applies clean. I applied it, ran `uv run pytest test/test_history_service_source_validation.py -v` (9 passed in 0.51s), then `uv run ruff check` and `uv run ruff format --check` on both changed files (clean), then reverted.
- The guard order is right. `not isinstance(source, str)` has to come first, because `[] in {"api", "db"}` raises TypeError on an unhashable value. Your parametrized `[]` and `{}` cases actually exercise that, so the type check is not cosmetic.
- The guard is non-breaking. I traced every in-process caller of `get_history`: restx_api/history.py:41, portfolio/data.py:347, services/flow_openalgo_client.py:295, services/historify_service.py:322, blueprints/chart_test.py:95, blueprints/scalping.py:187, blueprints/pnltracker.py:499/667/766/893, plus custom_straddle_service, iv_chart_service and multi_strike_oi_service. None can produce a value outside {"api", "db"} today and none passes `source=None`, so nothing that works now starts returning 400.
- Your full-suite note is accurate. test/test_bot_web.py:10 imports `get_telegram_bot`, which does not exist in services/telegram_bot_service.py, and eventlet is listed in requirements-nginx.txt but not in pyproject.toml, so `uv sync` does not install it. Both are pre-existing and unrelated to this PR.

## Required before merge

1. `.github/workflows/ci.yml` lines 47 to 52. Add `test/test_history_service_source_validation.py` to the backend-test allowlist. That step runs an explicit list of files, not `pytest test/`, so a new backend test that is not listed never executes in CI and this guard ships with no regression protection. test/conftest.py already assigns API_KEY_PEPPER, APP_KEY, LOG_DIR and the test database URLs, so the file runs there as is. (Frontend tests are different: the frontend-test job runs the whole vitest suite, so those need no registration. Backend tests do.) If pushing a workflow change from your fork is blocked on your side, say so and I will add the line on merge.

## Optional, fine to fold into the same push

2. services/history_service.py:245 and the Returns block at 247 to 251. The docstring still describes `source` as "'api' (broker, default) or 'db' (DuckDB/Historify)" and lists no error condition. One line saying an unsupported source returns 400 without calling a provider makes the new contract discoverable.

## Optional follow-up, not this PR

3. The allowed set is now spelled out in three places: restx_api/data_schemas.py:99 (`validate.OneOf(["api", "db"])`), portfolio/data.py:311, and your new literal in services/history_service.py. The repo's pattern for this is a shared constant: `VALID_EXCHANGES` in utils/constants.py, already imported by this very file at line 9. A `VALID_HISTORY_SOURCES` consumed by all three would stop them drifting. Separate PR is fine.

## Two things I am explicitly not asking for

- Keep the typing import cleanup on line 3. `uv run ruff check services/history_service.py` on current main reports exactly 3 errors, all UP035 on that one line (typing.Dict, typing.List, typing.Tuple), and your change clears all three. It is a lint fix inside the file you are already editing, not stray scope. Just call it out in the commit body.
- No rebase against PR #1866, the only other open PR touching services/history_service.py. Its hunk starts at line 4 and changes line 7 (the `database.auth_db` import), yours changes line 3, and its second hunk lands inside the api-key branch below your guard. I built both post-patch versions from main's base and ran a three-way merge: exit 0, zero conflict markers. The two compose in either order, so neither PR blocks the other.

## One framing note for the description, not a change request

Issue #1826 scopes this to the direct service call, which is the right scope, but it is worth being explicit that `/api/v1/history` already rejects unknown sources: restx_api/data_schemas.py:99 declares `source` with `validate.OneOf(["api", "db"])`, restx_api/history.py:30 loads through that schema, and the ValidationError handler at line 53 returns 400. portfolio/data.py:311 raises DataError for the same. The real value of your guard is the in-process path with no schema in front of it: the Flow executor reads `source` from node data at services/flow_executor_service.py:1769, 1831 and 1997 via `get_str`, which interpolates `{{var}}` templates, and hands it to services/flow_openalgo_client.py:295. The Flow UI constrains that field to a two-option select (frontend/src/components/flow/panels/ConfigPanel.tsx:2228), but a template variable is not constrained, and today a typo there silently falls through to the broker API instead of erroring. Say that plainly, rather than letting the cubic auto-summary's "Migration" line imply API clients need to change.

## On the checks you see

CI has not run on this branch. Fork PRs wait on maintainer approval here, so the only green marks are Snyk (which reports no manifest changes) and the cubic reviewer. No Ruff, no pytest and no frontend build has touched your commit. That is why I ran the tests and Ruff locally, and it is the other reason item 1 matters: once the file is on the allowlist it runs on every future push.

Push the ci.yml line and I will merge this.

siddharthg2309:
I’ve added the CI test registration and the documentation clarification. The shared-constant suggestion is left for a separate follow-up, as suggested.

marketcalls:
Merged as `c654d85e`, and #1826 is closed.

You did the two things that made this straightforward to merge, and both are worth naming because they are the ones most PRs in this batch got wrong.

**You registered the test.** `test/test_history_service_source_validation.py` is in the CI-safe list, and I confirmed on merged main it sits at position 3 of 18 and runs 9 passed. Three other PRs this week added a backend test without that line and shipped files that execute nowhere.

**You dropped the unrelated typing-import cleanup when I asked.** That was the specific thing that would have collided with #1866 on the same import block in `services/history_service.py`. Removing it kept the diff to the three files the change actually needs.

I verified the merge against current main rather than trusting the `CLEAN` label, because main moved three times while this sat open. The computed merge tree keeps the Python 3.12/3.13/3.14 matrix intact and carries every entry that landed in the meantime, including `test_market_calendar_validation.py` and `test_admin_client_error_url.py`, with yours inserted correctly. No conflict, nothing reverted.

## The scope note, restated in the merge commit

This is defence in depth, not a user-visible API fix, and you accepted that framing when I raised it rather than arguing the point. `/api/v1/history` already rejects an unknown `source` with a 400 at the marshmallow layer via the `OneOf` on `HistorySchema` at `restx_api/data_schemas.py:99`. The guard you added protects the in-process callers, principally the Flow execution path, which bypass that schema entirely.

I put that in the squash body so the merge record is honest about what the change does and does not cover. Issue #1826 asked for exactly this, so the scope is right; it is just worth the reader knowing the API surface was never exposed.

Clean PR. Nothing outstanding.

PR #1878, opened 2026-08-23
fix: validate option expiry codes

### LINKED ISSUE #1821: python: Reject malformed option expiry codes at the API boundary
## Summary

Several option endpoints accept an unconstrained expiry string although downstream code expects a calendar-valid `DDMMMYY` value.

## Evidence

- `restx_api/data_schemas.py:168` declares an unconstrained expiry.
- `restx_api/schemas.py:212` and `restx_api/schemas.py:267` expose the same input contract.

## Scope

Create one reusable validator for exact `DDMMMYY` form and calendar validity, then apply it to the matching schemas.

## Acceptance criteria

- [ ] `28AUG26` is accepted.
- [ ] Invalid dates, casing, length, and separators are rejected.
- [ ] All matching schemas reuse one validator.
- [ ] Tests include leap-day and malformed cases.

## Verification

```powershell
uv run pytest test/test_option_expiry_validation.py -v
uv run ruff check restx_api/data_schemas.py restx_api/schemas.py test/test_option_expiry_validation.py
```

- Difficulty: Easy
- Estimated time: 1-2 hours
- You will learn: date parsing, shared validators, and parameterized tests.

siddharthg2309:
## Summary

Add one shared validator for option expiry codes at the API boundary.

The validator requires the exact uppercase `DDMMMYY` format and rejects dates that do not exist on the calendar.

## Related issue

Closes #1821

## Changes

- Add `validate_option_expiry()` in `restx_api/data_schemas.py`.
- Reuse the validator for the Option Symbol, Single Option Order, and Multi Option Order leg expiry fields.
- Add focused schema-validation tests.

## Coverage

- Valid expiry: `28AUG26`
- Valid leap day: `29FEB24`
- Invalid calendar dates: `29FEB25`, `31APR26`
- Invalid casing: `28aug26`
- Invalid length and separators: `28AUG2026`, `1AUG26`, `28-AUG-26`
- Shared validator reuse across all three updated schemas

## Testing

- [x] `env -u LOG_FORMAT uv run pytest test/test_option_expiry_validation.py -v`
  - 21 passed
- [x] `uv run ruff check restx_api/data_schemas.py restx_api/schemas.py test/test_option_expiry_validation.py`
  - passed
- [x] `git diff --check`
  - passed

## Local environment note

This machine sets `LOG_FORMAT=json`, which conflicts with the repository logging setup during test collection. The focused test was run with that local-only environment variable removed; no project configuration was changed.

## Checklist

- [x] One reusable validator is used by every schema changed in this PR.
- [x] Valid, leap-day, malformed, and calendar-invalid inputs are covered.
- [x] No broker credentials or sensitive data are included.

---
## Summary by cubic
Enforces strict validation for option expiry codes. Previously `expiry_date` accepted any string; now only uppercase DDMMMYY (e.g., 28AUG26) that map to real calendar dates are accepted, otherwise the API returns a schema validation error.

**Reviewer notes**
- Applies `validate_option_expiry()` to `OptionSymbolSchema`, `OptionsOrderSchema`, `OptionsMultiOrderLegSchema`, and `OptionsMultiOrderSchema`.
- Rejects impossible dates and format errors (e.g., 29FEB25, 28aug26, 28-AUG-26); allows leap day (29FEB24).
- Keeps `expiry_date` optional and allows empty string when the underlying embeds expiry.
- Adds `test/test_option_expiry_validation.py` and runs it in the CI-safe test job.

<sup>Written for commit 9407d015efe039384d3f69a2cc4507c6b80816b4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1878?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,6 +45,7 @@ jobs:
       - name: Run CI-safe tests
         run: |
           uv run pytest test/test_log_location.py test/test_navigation_update.py \
+            test/test_option_expiry_validation.py \
             test/test_python_editor.py test/test_rate_limits_simple.py \
             test/test_logout_csrf.py test/test_auth_logout.py \
             test/test_auth_resume.py test/test_auth_upsert_multisession.py \
--- a/restx_api/data_schemas.py
+++ b/restx_api/data_schemas.py
@@ -40,6 +40,9 @@ def validate_option_offset(data: str) -> bool:
 
 def validate_option_expiry(data: str) -> None:
     """Validate an option expiry date in the exact DDMMMYY format."""
+    if data == "":
+        return
+
     if not isinstance(data, str) or not re.fullmatch(r"\d{2}[A-Z]{3}\d{2}", data):
         raise ValidationError("Expiry date must use the DDMMMYY format (for example, 28AUG26).")
 
--- a/restx_api/data_schemas.py
+++ b/restx_api/data_schemas.py
@@ -1,4 +1,5 @@
 import re
+from datetime import datetime
 
 from marshmallow import Schema, ValidationError, fields, validate
 
@@ -35,6 +36,17 @@ def validate_option_offset(data: str) -> bool:
         raise ValidationError("Offset must be ATM, ITM1-ITM50, or OTM1-OTM50")
 
     return True
+
+
+def validate_option_expiry(data: str) -> None:
+    """Validate an option expiry date in the exact DDMMMYY format."""
+    if not isinstance(data, str) or not re.fullmatch(r"\d{2}[A-Z]{3}\d{2}", data):
+        raise ValidationError("Expiry date must use the DDMMMYY format (for example, 28AUG26).")
+
+    try:
+        datetime.strptime(data, "%d%b%y")
+    except ValueError as error:
+        raise ValidationError("Expiry date must be a valid calendar date.") from error
 
 
 class QuotesSchema(Schema):
@@ -167,7 +179,7 @@ class OptionSymbolSchema(Schema):
     underlying = fields.Str(required=True)  # Underlying symbol (NIFTY, RELIANCE, NIFTY28OCT25FUT)
     exchange = fields.Str(required=True, validate=validate.OneOf(VALID_EXCHANGES))  # Exchange (NSE_INDEX, NSE, NFO)
     expiry_date = fields.Str(
-        required=False
+        required=False, validate=validate_option_expiry
     )  # Expiry date in DDMMMYY format (e.g., 28OCT25). Optional if underlying includes expiry
     strike_int = fields.Int(
         required=False, validate=validate.Range(min=1), allow_none=True
--- a/restx_api/schemas.py
+++ b/restx_api/schemas.py
@@ -294,7 +294,7 @@ class OptionsMultiOrderSchema(Schema):
     underlying = fields.Str(required=True)  # Underlying symbol (NIFTY, BANKNIFTY, RELIANCE)
     exchange = fields.Str(required=True, validate=validate.OneOf(VALID_EXCHANGES))  # Exchange (NSE_INDEX, NSE, BSE_INDEX, BSE)
     expiry_date = fields.Str(
-        required=False
+        required=False, validate=validate_option_expiry
     )  # Optional if underlying includes expiry (DDMMMYY format)
     strike_int = fields.Int(
         required=False, validate=validate.Range(min=1), allow_none=True
--- a/restx_api/schemas.py
+++ b/restx_api/schemas.py
@@ -1,5 +1,6 @@
 from marshmallow import EXCLUDE, Schema, ValidationError, fields, post_load, pre_load, validate
 
+from restx_api.data_schemas import validate_option_expiry
 from utils.constants import CRYPTO_EXCHANGES, VALID_EXCHANGES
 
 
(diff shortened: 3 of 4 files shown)


marketcalls:
Thanks for a clean, tightly scoped patch. This is the right shape for #1821: one shared `validate_option_expiry()` in `restx_api/data_schemas.py`, reused by exactly the three fields the issue named, a linear `re.fullmatch` with no backtracking risk, and 21 parameterized tests. I applied it on a fresh checkout of `main` and verified: it applies cleanly, `ruff check` passes on all three files, `ruff format --check` shows no new drift (both schema files already deviate identically on unpatched main), there is no import cycle from `restx_api/schemas.py` importing `restx_api/data_schemas.py`, and all 21 tests pass.

Rejecting lowercase is correct and is what the issue asked for, and it costs nothing today: `services/option_symbol_service.py:793` passes the expiry verbatim into `construct_option_symbol`, and the lookup at `services/option_symbol_service.py:316` is a case sensitive `==` with no NOCASE collation, so `28oct25` already returns a confusing 404. Your patch turns that into a clear 400.

Four things before merge.

1. Stop rejecting the empty string, in `restx_api/data_schemas.py`. The service layer deliberately reads a falsy expiry as "not supplied": `services/option_symbol_service.py:631` is `final_expiry = expiry_date or embedded_expiry`, and the Flow executor encodes the same rule at `services/flow_executor_service.py:517` (`has_explicit_expiry = bool(expiry_date)`), where an empty expiry is fine whenever the underlying carries one. Return early from the validator when the value is empty, and drop the `""` case from the reject list at `test/test_option_expiry_validation.py:50`. Omitting the field entirely already works, and I confirmed your patch keeps that path intact.

2. Cover the common expiry on the multi order endpoint, `restx_api/schemas.py:295` (`OptionsMultiOrderSchema.expiry_date`). Right now one request body validates the per leg override but not the value every leg falls back to: `services/options_multiorder_service.py:225` and `:476` both do `leg.get("expiry_date") or common_data.get("expiry_date")`. I loaded both payloads against your patch: `28-AUG-26` inside a leg is rejected, the identical string at the top level is accepted. The issue did not list that line, so this is not a miss on your part, but it is one line and it makes the endpoint self consistent.

3. Register the test in CI. `.github/workflows/ci.yml:45-53` enumerates the CI safe tests explicitly, and `test/test_option_expiry_validation.py` is not in that list, so it will never run in GitHub Actions. It is pure schema validation with no DB and no network, so please add it there. Worth knowing that CI has not run on this branch at all yet (fork PRs need maintainer approval to run workflows), so local runs are currently the only evidence either of us has.

4. Drop the private internals assertion at `test/test_option_expiry_validation.py:59`. `_declared_fields` and `Field.validate` are marshmallow internals; the very next line already asserts the real behaviour through `schema_class().load(...)`, which is the one worth keeping. While you are in the file, add a case for the behaviour the PR promises to preserve: `expiry_date` omitted with an underlying such as `NIFTY28OCT25FUT`.

Two smaller points, take or leave.

- `datetime.strptime(data, "%d%b%y")` makes the month name locale dependent, since `%b` resolves through `LC_TIME`. This repo already has a locale independent DDMMMYY parser for exactly this: `MONTH_MAP` at `services/option_greeks_service.py:69`, used by `get_expiry_datetime` at line 155. Reusing that avoids a fourth slightly different DDMMMYY parser (the others are `services/flow_executor_service.py:91` and the `re.match` checks in `blueprints/gex.py:59`, `blueprints/ivsmile.py:59`, `blueprints/oiprofile.py:66`, `blueprints/gamma_density.py:63`). Related: `get_expiry_datetime` uppercases before matching, so after this PR `/api/v1/optiongreeks` accepts `28aug26` while `/api/v1/optionsymbol` rejects it. Not blocking, but it is the reason a shared helper would be better than a fourth copy.
- `28XXX26` currently reports "Expiry date must be a valid calendar date" when the real fault is an unknown month token. One combined message would read better.

`OptionChainSchema.expiry_date` (`restx_api/data_schemas.py:223`) and `SyntheticFutureSchema.expiry_date` (`restx_api/schemas.py:314`) are both required and both still unvalidated, but they sit outside the scope #1821 defined, so please leave them out of this PR. I will open a follow up issue for them.

Heads up on merge order rather than on the code: PR #1873 adds `from restx_api.data_schemas import validate_option_offset` at the same import line in `restx_api/schemas.py`, so whichever of the two lands second will need a one line rebase.

Finally, your note about `LOG_FORMAT=json` is a real repo bug, not a local quirk. `utils/logging.py:353` reads `LOG_FORMAT` as a percent style format string, so any non-format value aborts pytest collection with `ValueError: Invalid format 'json' for '%' style`. I reproduced it. Please open that as its own issue rather than folding it into this PR.

siddharthg2309:
Thanks for the review. I’ve addressed the requested changes: empty expiry values retain the existing underlying fallback, the common multi-order expiry is now validated, and the updated tests are included in the CI-safe suite.

marketcalls (commented):
Reviewed head `9407d015efe039384d3f69a2cc4507c6b80816b4`: 89/100. The requested schema behavior and follow-up fixes are covered; all 27 focused tests passed and CI is green. No new code defect was found. Rebase the current conflict with main, preserving the shared expiry validator and common multi-order expiry validation, then rerun the focused suite and required CI.

PR #1879, opened 2026-08-23
fix: label transaction cost controls

### LINKED ISSUE #1832: frontend: Associate transaction-cost controls with programmatic labels
## Summary

Several transaction-cost controls display labels visually but are not programmatically associated with their form fields, making screen-reader navigation ambiguous.

## Evidence

- `frontend/src/components/portfolio/ChargeControls.tsx:42`, `:68`, and `:89-110` render labeled controls.
- `frontend/src/components/portfolio/ChargeControls.tsx:127-145` has the same pattern.
- The labels lack matching `htmlFor` and input `id` values.

## Scope

Give every visible label a stable association with its control. Preserve component styling and state behavior, and add an accessible-name test.

## Acceptance criteria

- [ ] Every input/select has a unique stable id.
- [ ] Every visible label references the matching control.
- [ ] Controls can be queried by their visible label text.
- [ ] Existing values and change handlers still work.

## Verification

Run from `frontend/`:

```powershell
npm run test:run -- src/components/portfolio/ChargeControls.test.tsx
npm run lint
```

- Difficulty: Easy
- Estimated time: 1-2 hours
- You will learn: accessible form labeling and React Testing Library queries.

siddharthg2309:
## Summary

Associate every transaction-cost input with its visible label so screen readers can identify each control correctly.

## Related issue

Closes #1832

## Changes

- Add stable, unique IDs to all transaction-cost inputs.
- Connect visible labels to their matching inputs with `htmlFor`.
- Use semantic named groups for the Exchange and Brokerage button controls.
- Preserve existing styling, values, and state behavior.
- Add focused accessibility and interaction coverage.

## Coverage

- Inputs can be queried by their visible labels.
- Default Brokerage input and all standard charge inputs have unique stable IDs.
- Conditional percent Brokerage inputs are also labelled and uniquely identified.
- Exchange and Brokerage controls are exposed as named groups.
- Changing the exchange still updates the transaction fee.
- Changing a charge found by its visible label still updates its value.

## Testing

- [x] `npm run test:run -- src/components/portfolio/ChargeControls.test.tsx`
  - 4 passed
- [x] `npm run lint`
  - passed
- [x] `git diff --check`
  - passed

## Checklist

- [x] This PR contains one focused accessibility fix.
- [x] Every transaction-cost input has a stable unique ID.
- [x] Every visible numeric label is associated with its matching control.
- [x] Existing change-handler behavior is covered.
- [x] No backend or calculation behavior changed.
- [x] No UI redesign is included; screenshots are not applicable.

---
## Summary by cubic
Makes transaction-cost controls fully accessible to screen readers. Old: labels were unbound spans and toggle state wasn’t exposed; new: each control is programmatically labeled and groups are named; no visual or calculation changes.

- Assigns stable IDs and connects visible labels with htmlFor for all numeric inputs.
- Groups Exchange and Brokerage options in fieldsets with accessible names.
- Exposes the active option via aria-pressed on Exchange and Brokerage buttons.
- Adds a11y tests, including `jest-axe`, for label associations, group roles, and violations.

<sup>Written for commit e0c8f3520576397dd9bd8070beb7183fb665b35b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1879?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/src/components/portfolio/ChargeControls.test.tsx
+++ b/frontend/src/components/portfolio/ChargeControls.test.tsx
@@ -21,6 +21,31 @@ function Harness() {
 }
 
 describe('ChargeControls', () => {
+  it('associates every visible charge label with a uniquely identified control', () => {
+    render(<Harness />)
+
+    expect(screen.getByRole('group', { name: 'Exchange' })).toBeInTheDocument()
+    expect(screen.getByRole('group', { name: 'Brokerage' })).toBeInTheDocument()
+
+    const labelledControls = [
+      ['Per order (₹)', 'portfolio-charge-brokerage-flat'],
+      ['STT (%)', 'portfolio-charge-stt'],
+      ['Exchange txn (%)', 'portfolio-charge-exchangeTxn'],
+      ['Stamp duty (%)', 'portfolio-charge-stampDuty'],
+      ['GST (%)', 'portfolio-charge-gst'],
+      ['SEBI (₹/crore)', 'portfolio-charge-sebiPerCrore'],
+      ['Slippage (%)', 'portfolio-charge-slippage'],
+    ]
+
+    const ids = labelledControls.map(([label, id]) => {
+      const control = screen.getByLabelText(label)
+      expect(control).toHaveAttribute('id', id)
+      return control.id
+    })
+
+    expect(new Set(ids)).toHaveLength(ids.length)
+  })
+
   it('updates the exchange transaction rate with the venue', async () => {
     const user = userEvent.setup()
     render(<Harness />)
@@ -29,5 +54,32 @@ describe('ChargeControls', () => {
     await user.click(screen.getByRole('button', { name: 'BSE' }))
 
     expect(screen.getByDisplayValue('0.00375')).toBeInTheDocument()
+  })
+
+  it('associates the percent brokerage controls with their visible labels', async () => {
+    const user = userEvent.setup()
+    render(<Harness />)
+
+    await user.click(screen.getByRole('button', { name: '% of order' }))
+
+    expect(screen.getByLabelText('Rate (%)')).toHaveAttribute(
+      'id',
+      'portfolio-charge-brokerage-pct'
+    )
+    expect(screen.getByLabelText('Cap / order (₹)')).toHaveAttribute(
+      'id',
+      'portfolio-charge-brokerage-cap'
+    )
+  })
+
+  it('updates a charge control found by its visible label', async () => {
+    const user = userEvent.setup()
+    render(<Harness />)
+
+    const stt = screen.getByLabelText('STT (%)')
+    await user.clear(stt)
+    await user.type(stt, '0.2')
+
+    expect(stt).toHaveValue(0.2)
   })
 })
--- a/frontend/src/components/portfolio/ChargeControls.test.tsx
+++ b/frontend/src/components/portfolio/ChargeControls.test.tsx
@@ -1,4 +1,5 @@
 import { useState } from 'react'
+import { axe, toHaveNoViolations } from 'jest-axe'
 import { describe, expect, it } from 'vitest'
 import { render, screen, userEvent } from '@/test/test-utils'
 import {
@@ -6,6 +7,8 @@ import {
   type ChargeState,
 } from '@/lib/portfolioRequest'
 import { ChargeControls } from './ChargeControls'
+
+expect.extend(toHaveNoViolations)
 
 function Harness() {
   const [value, setValue] = useState<ChargeState>(DEFAULT_CHARGES)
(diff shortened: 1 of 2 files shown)


marketcalls:
Thanks for this. The defect is real and the core of the fix is right, so this is a **merge after one required change**, not a rework.

## What I verified

Applied your head commit f2f3df86 on top of current main and ran it here:

- Patch applies cleanly to main.
- `npx vitest run src/components/portfolio/ChargeControls.test.tsx`: 4 passed.
- `npx tsc -b`: exit 0.
- `npx biome lint` on both changed files: clean.
- The defect on main is exactly as you describe: `ChargeControls.tsx` lines 88, 99, 109 and 136 are bare `<span>` elements sitting in front of each `<Input>`, and lines 42 and 68 are `<Label>` with no `htmlFor` and no nested control. None of the numeric inputs has an accessible name today.

Worth knowing: CI has not run on this branch. Fork PRs wait on maintainer approval, so the only green checks you see are Snyk and cubic. When it does run, the `frontend-test` job executes `npm run test:run` over the whole vitest suite (`.github/workflows/ci.yml:106`), so your three new cases will genuinely gate the merge rather than sit dormant. Nothing backend is touched, so no migration is in play.

## Blocking

**1. Revert the two grouping wrappers from `<fieldset>` plus visible `<legend>` back to `<div>`. `ChargeControls.tsx:41-42` and `:67-68`.**

A fieldset's rendered legend is painted over the block-start border edge, and the border is not drawn behind it. I did not take this from the spec alone: I rebuilt the exact markup with the Tailwind v4 preflight and utility rules extracted from `frontend/dist/assets/index-21IofHHz.css` and rendered both variants in Chrome. With `<fieldset className="m-0 min-w-0 space-y-2 border-0 border-t pt-2">` and `<legend>Brokerage</legend>`, the `border-t` divider above Brokerage is cut away for the full width of the word and the heading rides on the line, instead of a solid full-width divider with the heading `pt-2` below it as on main.

The bare `<legend>` also drops `font-medium` and `leading-none`, which `components/ui/label.tsx:11` applies to every `<Label>`, so both the Exchange and Brokerage headings drop from weight 500 to 400 with normal line height. That change is visible in the same render.

Together that contradicts "Preserve existing styling, values, and state behavior" in your own PR body, and makes the checklist item "No UI redesign is included; screenshots are not applicable" inaccurate as the branch stands.

It would also be the first visible legend in the codebase. The only `<legend>` under `frontend/src` today is screen-reader-only (`components/flow/panels/MarginPositionsFields.tsx:416`), and every visible toggle group uses `<fieldset aria-label=...>` with the visible heading left outside it: `components/option-chain/ViewModeToggle.tsx:20-22`, `components/strategy-builder/ManualLegBuilder.tsx:427`, `components/trading/ExecuteBasketDialog.tsx:277`.

Either shape keeps the accessible group name and the current pixels:

```tsx
<div
  className="space-y-2 border-t pt-2"
  role="group"
  aria-labelledby="portfolio-charge-brokerage-heading"
>
  <Label id="portfolio-charge-brokerage-heading" className="text-[11px]">
    Brokerage
  </Label>
```

or leave the `<div>` and `<Label>` exactly as they are on main and wrap only the button row in `<fieldset aria-label="Brokerage">`, matching ViewModeToggle. Your existing `getByRole('group', { name: 'Brokerage' })` assertions pass either way, so the tests do not need to change.

Everything else in the patch is correct and I want it as is: the ids and the `htmlFor` pairs at 89/95, 106/112, 122/128 and 155/161 are precisely the fix.

## Please fold into the same push (not blocking on their own)

**2. `aria-pressed` on the two button groups. `ChargeControls.tsx:45` and `:71`.**

Now that the buttons carry a group name, the missing piece is which option is live. Selected state is conveyed only by the `border-primary/50 bg-primary/15 text-primary` colour swap, so a screen reader announces "Exchange group, NSE button, BSE button" with no current venue. Add `aria-pressed={exchange === ex}` and `aria-pressed={value.brokerageMode === m}`. Every other toggle group in the repo already does this: `ViewModeToggle.tsx:33`, `ManualLegBuilder.tsx:435`, `ExecuteBasketDialog.tsx:285`.

**3. Test naming and the uniqueness assertion. `ChargeControls.test.tsx`.**

`frontend/package.json:20` defines `test:a11y` as `vitest run --testNamePattern='accessibility'`. None of your three new test names contains "accessibility", so the dedicated a11y run skips all of them. Wrap them in `describe('accessibility', ...)` and add the repo's axe check, as in `components/ui/button.test.tsx:141-146` and `components/layout/MobileBottomNav.test.tsx:110-115`.

Separately, change line 46 from `expect(new Set(ids)).toHaveLength(ids.length)` to `expect(new Set(ids).size).toBe(ids.length)`. The current form passes only because Vitest's `toHaveLength` delegates straight to chai's `have.length` (`node_modules/@vitest/expect/dist/index.js:1388`), which special-cases `Set.size`. It states the uniqueness check indirectly and would break under a stricter matcher.

## Optional, take or leave

**4. Drop `m-0`, `p-0` and `border-0`** if any fieldset survives the rewrite. Tailwind v4 preflight already emits `*,:after,:before,::backdrop{box-sizing:border-box;border:0 solid;margin:0;padding:0}`, so all three are no-ops. Only `min-w-0` is load-bearing, since it overrides the UA `min-inline-size: min-content` on fieldset. To be clear, I checked the emitted order in the built CSS and `.border-t` comes after `.border-0`, so that pair is redundant rather than dangerous.

**5. Leave the hardcoded ids alone.** They are stable and unique, which is what you set out to deliver, and `ChargeControls` is mounted only from `PortfolioBacktester.tsx:394` and `SipBacktester.tsx:332`, which are separate routes, so nothing can collide. `useId()` would be marginally more robust but it is not the house pattern here: exactly two files under `frontend/src` use it, against 65 literal `htmlFor="..."` associations.

## Ordering against the other open accessibility PRs

This is one of four open a11y PRs (#1869, #1877, #1879, #1880). There is no file overlap between any of them, so merge order is free and no rebase is required for conflict reasons.

One coupling to be aware of: `pages/PortfolioBacktester.tsx:394` renders `ChargeControls`, and #1877 adds `pages/PortfolioBacktester.test.tsx`, so that page test will render your component. I checked for accessible-name collisions between the two and found none. `getByLabelText` defaults to an exact match, so #1877's "Risk-free rate (%)" does not pick up your "Rate (%)", and "Rate (%)" only renders in percent brokerage mode, which is not the default. Whichever of the two lands second, rebase onto main and run the full `npm run test:run` rather than just the one file.

Push the grouping revert and I will merge.

siddharthg2309:
Thanks for treview. I restored the visible div and Label layout, used named fieldsets only around the toggle rows, and added aria-pressed to expose the selected options.

The accessibility tests now run under the accessibility suite, include axe coverage, and use the direct Set.size uniqueness check.

Kalaiviswa:
Thanks for the accessibility improvements. The numeric controls and toggle states are covered well. Please address two remaining points before merge:

  1. Generate the control IDs from React useId() instead of hardcoding them. The current IDs collide if ChargeControls is rendered more than once in the same document.

  2. Use the visible Exchange and Brokerage text as <legendelements inside their respective <fieldset> elements. The current <Labelelements are unassociated, while the fieldsets receive separate duplicated names through aria-label.

  Please retain the aria-pressed state and add a test that renders two component instances and verifies document-wide ID uniqueness.

PR #1881, opened 2026-08-23
fix: redact API keys from core logs

### LINKED ISSUE #1855: security: Remove API-key material from core service logs
## Summary

Core services log a full API key or the first ten characters of an invalid key.

## Evidence

- `services/websocket_client.py:577-581` includes the dictionary key (the API key) in a close-error message.
- `restx_api/option_greeks.py:116` logs the first ten characters of an invalid key.
- `restx_api/multi_option_greeks.py:106` repeats the same prefix logging.

## Why this matters

API keys should never be recoverable from logs. Prefixes can still be sensitive and can help correlate or enumerate credentials.

## Scope

Remove key content from these messages. If correlation is necessary, use a short one-way diagnostic fingerprint with a documented domain separator, or log only that a key was present.

## Acceptance criteria

- [ ] No full or partial API key is interpolated into a log message.
- [ ] Different client-close failures remain distinguishable without key material.
- [ ] Invalid-key requests keep their current HTTP response.
- [ ] Tests capture logs and assert that dummy key content never appears.

## Verification

```powershell
uv run pytest test/test_core_credential_logging.py -v
uv run ruff check services/websocket_client.py restx_api/option_greeks.py restx_api/multi_option_greeks.py test/test_core_credential_logging.py
```

## Security guidance

- Difficulty: Easy
- Estimated time: 1-3 hours
- You will learn: secure logging, log capture tests, and safe diagnostic correlation.

mikemikimike:
Closes #1855

Removes full and partial API keys from WebSocket close errors and both option Greeks invalid-key warnings. WebSocket cleanup now uses `logger.exception` with a message that contains no key or exception text, while preserving the traceback for diagnostics. Adds focused regression tests for HTTP 401 preservation and credential-free logging, including an exception message containing a test API key.

Tests:
- `python -m pytest test/test_core_credential_logging.py -q` - 3 passed
- `python -m compileall` on changed files - passed
- `python -m ruff check test/test_core_credential_logging.py` - passed
- `git diff --check` - passed

CI ran the new tests successfully. Four pre-existing Ruff warnings remain in `websocket_client.py` and are unrelated to this change.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,6 +52,7 @@ jobs:
             test/test_telegram_api_contract.py \
             test/test_multi_option_greeks_regression.py \
             test/test_history_format.py \
+            test/test_core_credential_logging.py \
             -v --timeout=60
 
   # Frontend linting
--- a/restx_api/multi_option_greeks.py
+++ b/restx_api/multi_option_greeks.py
@@ -103,7 +103,7 @@ class MultiOptionGreeks(Resource):
 
             # Verify API key
             if not verify_api_key(api_key):
-                logger.warning(f"Invalid API key used for multi option greeks: {api_key[:10]}...")
+                logger.warning("Invalid API key used for multi option greeks")
                 return make_response(
                     jsonify({"status": "error", "message": "Invalid openalgo apikey"}), 401
                 )
--- a/restx_api/option_greeks.py
+++ b/restx_api/option_greeks.py
@@ -113,7 +113,7 @@ class OptionGreeks(Resource):
 
             # Verify API key
             if not verify_api_key(api_key):
-                logger.warning(f"Invalid API key used for option greeks: {api_key[:10]}...")
+                logger.warning("Invalid API key used for option greeks")
                 return make_response(
                     jsonify({"status": "error", "message": "Invalid openalgo apikey"}), 401
                 )
--- a/services/websocket_client.py
+++ b/services/websocket_client.py
@@ -574,9 +574,9 @@ def get_websocket_client(
 def close_all_clients():
     """Close all WebSocket client connections"""
     with _client_lock:
-        for api_key, client in _client_instances.items():
+        for _api_key, client in _client_instances.items():
             try:
                 client.disconnect()
-            except Exception as e:
-                logger.exception(f"Error closing client for API key {api_key}: {e}")
+            except Exception:
+                logger.exception("Error closing WebSocket client")
         _client_instances.clear()
--- a/test/test_core_credential_logging.py
+++ b/test/test_core_credential_logging.py
@@ -12,6 +12,7 @@ def request_context():
 class RecordingLogger:
     def __init__(self):
         self.messages = []
+        self.error_messages = []
 
     def warning(self, message, *args):
         self.messages.append(message % args if args else message)
@@ -19,7 +20,8 @@ class RecordingLogger:
     def exception(self, message, *args):
         self.messages.append(message % args if args else message)
 
-    error = exception
+    def error(self, message, *args):
+        self.error_messages.append(message % args if args else message)
 
 
 def test_option_greeks_invalid_key_is_not_logged(monkeypatch):
@@ -81,4 +83,5 @@ def test_websocket_close_error_does_not_log_api_key(monkeypatch):
     websocket_client.close_all_clients()
 
     assert logger.messages == ["Error closing WebSocket client"]
+    assert logger.error_messages == []
     assert TEST_KEY not in " ".join(logger.messages)
--- a/test/test_core_credential_logging.py
+++ b/test/test_core_credential_logging.py
@@ -1,0 +1,84 @@
+"""Regression tests for credential-safe core service logging."""
+
+from flask import Flask
+
+TEST_KEY = "test-token-not-real-1234567890"
+
+
+def request_context():
+    return Flask(__name__).test_request_context("/", method="POST", json={"apikey": TEST_KEY})
+
+
+class RecordingLogger:
+    def __init__(self):
+        self.messages = []
+
+    def warning(self, message, *args):
+        self.messages.append(message % args if args else message)
+
+    def exception(self, message, *args):
+        self.messages.append(message % args if args else message)
+
+    error = exception
+
+
+def test_option_greeks_invalid_key_is_not_logged(monkeypatch):
+    from restx_api import option_greeks
+
+    logger = RecordingLogger()
+    monkeypatch.setattr(option_greeks, "logger", logger)
+    monkeypatch.setattr(
+        option_greeks.option_greeks_schema,
+        "load",
+        lambda data: {"apikey": TEST_KEY, "symbol": "NIFTY", "exchange": "NFO"},
+    )
+    monkeypatch.setattr(option_greeks, "verify_api_key", lambda api_key: False)
+
+    with request_context():
+        response = option_greeks.OptionGreeks().post()
+
+    assert response.status_code == 401
+    assert logger.messages == ["Invalid API key used for option greeks"]
+    assert TEST_KEY not in " ".join(logger.messages)
+
+
+def test_multi_option_greeks_invalid_key_is_not_logged(monkeypatch):
+    from restx_api import multi_option_greeks
+
+    logger = RecordingLogger()
+    monkeypatch.setattr(multi_option_greeks, "logger", logger)
+    monkeypatch.setattr(
+        multi_option_greeks.multi_option_greeks_schema,
+        "load",
+        lambda data: {"apikey": TEST_KEY, "symbols": []},
+    )
+    monkeypatch.setattr(multi_option_greeks, "verify_api_key", lambda api_key: False)
+
+    with request_context():
+        response = multi_option_greeks.MultiOptionGreeks().post()
+
+    assert response.status_code == 401
+    assert logger.messages == ["Invalid API key used for multi option greeks"]
+    assert TEST_KEY not in " ".join(logger.messages)
+
+
+def test_websocket_close_error_does_not_log_api_key(monkeypatch):
+    from services import websocket_client
+
+    logger = RecordingLogger()
+    monkeypatch.setattr(websocket_client, "logger", logger)
+
+    class FailingClient:
+        def disconnect(self):
+            raise RuntimeError(f"disconnect failed for {TEST_KEY}")
+
+    monkeypatch.setattr(
+        websocket_client,
+        "_client_instances",
+        {TEST_KEY: FailingClient()},
+    )
+
+    websocket_client.close_all_clients()
+
+    assert logger.messages == ["Error closing WebSocket client"]
+    assert TEST_KEY not in " ".join(logger.messages)


marketcalls:
Thanks for this. The two Greeks call sites and the WebSocket cleanup line are real leaks, and I reproduced your tests: all three fail on current main and pass once the three Python hunks are applied. The redaction itself is the right change.

Blocking items, in order.

**1. frontend/package.json and frontend/package-lock.json must come out of this PR.**

The diff moves `"openalgo-charts"` from the exact pin `1.6.0` back to `^1.9.0`, and the lockfile entry from 1.6.0 to 1.9.0. Both hunks apply cleanly on top of HEAD (the diff base blobs `ffc9843fba` and `48c62931e3` are exactly HEAD's blobs), so merging this reverts commit `a1a2dc926`. That pin is deliberate: 1.7.0 through 1.9.0 are deprecated, and 1.6.0 is the release that stops a stop-limit drag sending `price: 0` to the broker (`950974639`). `frontend/dist` is tracked on main and rebuilt by CI after every push, so the next build would ship 1.9.0 chart code to everyone who runs `git pull`.

Your branch also carries `e6f810f9`, `29fcaaac` and `0e863016` (the 1.7.0, 1.8.0 and 1.9.0 bumps), which are no longer ancestors of main. That is why the merge in `dbb1131d` resolved the wrong way. Please rebase onto current `upstream/main` rather than merging, and drop both frontend files from the diff.

**2. services/websocket_client.py: keep logger.exception.**

CLAUDE.md is explicit that error logging is always `logger.exception()`. `logger.exception("Error closing WebSocket client")` gives you exactly what you want: no interpolated key, no interpolated exception text, traceback preserved.

One correction to the reasoning in the PR body. `logger.exception` would not have leaked anything here in the first place, because the only thing that can raise is `WebSocketClient.disconnect()` (services/websocket_client.py:128-142), which sets flags, schedules a coroutine and joins a thread. Nothing in that path can put an API key in the exception message. Worth knowing for a follow up though: `SensitiveDataFilter` in utils/logging.py rewrites only `record.msg` and `record.args`, and the JSON handler writes `traceback.format_exception(*record.exc_info)` straight to `log/errors.jsonl` (utils/logging.py:310-312). Tracebacks are never redacted anywhere in the codebase. That, not three call sites, is where #1855 actually gets closed.

**3. test/test_core_credential_logging.py must be registered in CI.**

Add it to the CI-safe pytest list in `.github/workflows/ci.yml` (the `Run CI-safe tests` step in `backend-test`). As it stands the file never executes in CI and cannot guard against a regression. Note that CI has not run on this branch at all yet, so treat the green checks as meaningless here. Snyk in particular reported no manifest changes while the PR changes both package.json and package-lock.json.

**4. test/test_core_credential_logging.py: run `uv run ruff format`.**

`uv run ruff format --check --diff` wants `request_context()` collapsed onto one line. Also drop the three `monkeypatch.setenv("LOG_FORMAT", ...)` calls: utils/logging.py calls `setup_logging()` at import and reads LOG_FORMAT at line 353, so setting it inside a test is a no-op, and the tests replace the module logger anyway.

Non blocking, but please consider.

- `close_all_clients()` has no callers anywhere in the repo (only its definition at services/websocket_client.py:574 and a mention in docs/prompt/services_documentation.md:629). It is also the only site in this PR that leaked a complete key. Either wire it into shutdown or delete it, otherwise the highest value part of the fix is unreachable.
- The tests assert a literal log string against a stub logger, so they pin wording rather than behaviour and never exercise the real redaction path. Asserting through `caplog` on the real logger would prove the property the PR title claims.
- Still leaking after this PR: broker/flattrade/api/auth_api.py:36 logs `Hash Input: {BROKER_API_KEY}{code}{BROKER_API_SECRET}` at DEBUG, which is the broker secret in plaintext, plus `api_key[:10]` in the compositedge, fivepaisaxts, ibulls, iifl, jainamxts, rmoney and wisdom streaming adapters. A widened `SENSITIVE_PATTERNS` entry would cover those, though it would not cover the two Greeks messages you fixed, because there the value trails the sentence ("... option greeks: <value>") rather than the credential word.
- The PR description uses em dashes. The repo writing rules do not allow them; a comma or a full stop is fine.

Once the two frontend files are out, `logger.exception` is back, and the test file is format clean and listed in ci.yml, this is good to merge.

marketcalls:
Not merging yet. The code is ready. The PR text is not, and one item reported as partially handled is not handled at all.

What I re-verified independently, all confirmed:

1. The frontend files are out, and the situation is better than described. I fetched the head ref rather than working from a patch file. The merge base of your branch and main is a1a2dc926, which is current main HEAD, so `git diff HEAD <head>` is the entire change and merging is a fast forward: 5 files, +90/-5. frontend/package.json and frontend/package-lock.json are byte identical to main (blobs ffc9843fba and 48c62931e3), the pin at line 60 reads "openalgo-charts": "1.6.0", the lock resolves 1.6.0 with the 1.6.0 integrity hash, and frontend/dist is untouched.

2. services/websocket_client.py:580-581 is `except Exception:` then `logger.exception("Error closing WebSocket client")`. No key, no interpolated exception text, traceback preserved.

3. ci.yml:53 adds exactly one line, 12 space indent matching its neighbours, no trailing whitespace, no CRLF, inside the existing pytest invocation. The workflow diff is +1/-0, so no job, trigger, permissions or continue-on-error moved. I did not take the green check on trust: I pulled the backend-test job log for run 32706295439 and the runner echoed the new path and ran all three new tests, which passed at 08:37:20 inside a 43 passed run.

4. The test file is ruff format clean, request_context() is one line at :9, and grep for LOG_FORMAT in the file returns nothing.

I also confirmed the tests are not vacuous, in both directions, in a clean worktree at the head: reverting the three Python hunks fails all three with the leaked strings visible, restoring them passes. And restx_api/option_greeks.py:116 and multi_option_greeks.py:106 are the only api_key[:N] log sites outside broker/, so the "core logs" scope is genuinely complete.

Two things stop me from clicking merge, both in the description.

First, the em dashes are not partially removed, they are not removed. The body still carries four U+2014 characters, bytes e2 80 94, on the four Tests lines, for example "test/test_core_credential_logging.py -q [em dash] 3 passed". That was an explicit request and it is a repo wide writing rule.

Second, and this is new since my last pass: fixing item 2 correctly invalidated the description, and the description did not follow. The body still says "WebSocket cleanup now logs only the exception type, avoiding exception text and traceback content that could contain credentials." That was true of the previous head, which used logger.error with type(e).__name__. It is false of the code you are asking me to merge. Line 581 is logger.exception, which attaches exc_info, and utils/logging.py:310-312 writes traceback.format_exception straight into log/errors.jsonl with no redaction, because SensitiveDataFilter only rewrites record.msg and record.args at lines 162-177. On a PR that closes a credential-leak issue, the merge record would carry an inverted claim about the very property being fixed.

Smaller, same paragraph: the body says "Five pre-existing Ruff warnings remain in websocket_client.py". It is four, UP035 twice at line 15 and UP041 at 202 and 270. I confirmed all four are pre-existing on main and untouched here. The "full uv sync was not run" note is also stale now that CI is green.

Please update the description only, no code change needed: drop the four em dashes, rewrite the WebSocket sentence to say what the code actually does (no key and no exception text in the message, traceback preserved via logger.exception), and fix the ruff count. I will squash on merge, since the branch still carries e6f810f9, 29fcaaac and 0e863016, the 1.7.0, 1.8.0 and 1.9.0 bumps, none of which are ancestors of main.

For the record, I ran the supply chain check this deserves as a fork PR that edits ci.yml and adds contributor-authored code to the CI execution path. It is clean: the trigger is pull_request rather than pull_request_target, so fork runs get no repository secrets and a read only token, docker-build, docker-manifest and commit-dist were all SKIPPED, and the new test file makes no network, subprocess, environment or filesystem access.

One note for the follow up rather than for this PR: test_websocket_close_error_does_not_log_api_key asserts only against the stub logger's message list, so it says nothing about exc_info. Also worth knowing, in the two greeks tests the trailing `assert TEST_KEY not in " ".join(...)` passed even on unfixed main, because TEST_KEY is 30 characters and the old code logged only api_key[:10]. The equality assert is doing all the work. The property the title claims gets closed by redacting tracebacks in utils/logging.py, not by these three call sites.

marketcalls:
Not merging yet. One item that was satisfied last round has come undone, and I found one thing in the test file that three rounds of review including my own missed. Both fixes are small and belong in the same push.

**1. The rebase dropped the ci.yml registration.**

At the previous head 8e805b79, `.github/workflows/ci.yml` carried `test/test_core_credential_logging.py \` at line 53. At f45650f8 it is gone. I checked this by object id rather than by eye: `git rev-parse` on ci.yml at your head and at main both return 84875dfefcea1bae16b94beb5fad325045cd87f7, so the file is byte identical to main. The list now ends `test_telegram_api_contract.py`, `test_multi_option_greeks_regression.py`, `test_history_format.py`, `-v --timeout=60`, and ci.yml is not among the four files in the compare against main.

This is not a flaky check. backend-test is green because it never runs your tests. I pulled both backend-test logs for this SHA, jobs 97484414527 and 97483246994, and `grep -c credential_logging` returns 0 in each, with a summary line of `53 passed`. Your previous head's log, job 97369351947, has four occurrences and prints all three tests PASSED inside `43 passed`. backend-test has no continue-on-error, so its green really does mean the suite it ran was fine, and your tests were not in it.

The cause is not carelessness. main moved under you: `test_multi_option_greeks_regression.py` and `test_history_format.py` landed in that same continuation block, and rebasing resolved the block to main's version and took your line with it.

```
            test/test_history_format.py \
            test/test_core_credential_logging.py \
            -v --timeout=60
```

I ran exactly that list plus your file at your head: 56 passed, no collision with the multi_option_greeks regression test main just gained.

**2. The test stub cannot see the difference between logger.exception and logger.error.**

`RecordingLogger` ends with `error = exception` at test/test_core_credential_logging.py:23. Because both methods append to the same list, the assertions only pin the message text, never which method produced it. I checked what that costs: at your head I changed line 581 to `logger.error("Error closing WebSocket client")`, message untouched, and pytest reported 3 passed.

That matters because restoring `logger.exception` was blocking item 2, it is a CLAUDE.md rule that error logging is always `logger.exception()`, and it is the one thing your first submission got wrong. So the suite guards the redaction and leaves the invariant that was actually at risk uncovered. Give `error` its own list and assert it stays empty, or record the method name alongside the message. Two lines.

Everything else holds up, and I re-verified it at this head rather than carrying it over.

- The charts pin is untouched. frontend/package.json and frontend/package-lock.json are absent from the compare, and their blob ids at your head are ffc9843fba and 48c62931e3, the identical objects main has, so the trees cannot differ. Line 60 reads `"openalgo-charts": "1.6.0"`.
- services/websocket_client.py:577-581 is `for _api_key, client in ...`, `except Exception:`, `logger.exception("Error closing WebSocket client")`. No key, no interpolated exception text, traceback preserved. Both greeks warnings redacted with the 401 intact.
- The body is pure ASCII. All 780 bytes, and `grep '[^ -~]'` over the whole thing matches nothing, so no U+2014 and no U+2013. The title is clean too. Thank you.
- The WebSocket sentence is now accurate and the merge record no longer inverts the property this PR fixes. Ruff count corrected to four, and I confirmed all four are pre-existing on main.
- You went further than I asked on history. I said I would squash; instead you rebased to a single commit whose parent is 0f99ab6c, current main HEAD. mergeStateStatus is CLEAN, so this is a fast forward and needs no special handling.
- Tests are non-vacuous in both directions: 3 pass at your head, and reverting the three Python hunks fails all three with the key visible in every assertion diff. LF endings throughout, `git diff --check` clean, `ruff check` on the test file passes.

One consequence to fix while you are in there: the body says "CI ran the new tests successfully". That was true at 8e805b79 and is false at f45650f8. Restoring the ci.yml line makes it true again.

Push the ci.yml line, the stub fix, and the body edit, and I will merge.

marketcalls:
Closing this because #1855 is now fixed on `main` in 9a4a1d0c, via #1914.

I want to be straight with you about what happened, because you were here first. You opened this on 23 August and #1914 arrived on the 26th, so yours was the earlier submission and it got superseded. That is a poor outcome for you and it was not a judgement on the quality of the work.

The substantive reason the other one was taken comes down to two of the issue's own criteria rather than anything wrong here:

- The issue's Scope asked for "a short one-way diagnostic fingerprint with a documented domain separator, or log only that a key was present". #1914 used a `sha256:` fingerprint, so an operator can still correlate repeated invalid-key attempts. Yours removes the identifier entirely, which is the other option the issue allowed, but leaves nothing to correlate on.
- Acceptance criterion 2 says "different client-close failures remain distinguishable without key material". Your version logs `"Error closing WebSocket client"` with no `{e}`, so two different failures read identically. #1914 keeps the exception text, which preserves that distinction.

Where your version is genuinely stronger: dropping `{e}` is the safer default, because an exception message can carry a credential the format string never mentions. `httpx.HTTPStatusError` embeds the full request URL, which is how `broker/definedge/api/auth_api.py` was leaking its API key on every 4xx. I checked that specific risk for the merged version before approving it: `ws_url` there is `ws://{host}:{port}` with no credential and `disconnect()` cannot raise carrying the key, so it is safe as merged. Your instinct was right in general even though it does not bite in this particular spot.

Your tests were good too, asserting on captured messages rather than re-implementing the check.

Sorry for the wasted effort, and thank you for it. If you would like another, the `good first issue` and `security` labels have open items, and commenting on one before starting is the reliable way to avoid a collision like this.

PR #1882, opened 2026-08-23
fix: reject cross-origin referrers in CSRF and admin rate-limit redirects

### LINKED ISSUE #1852: security: Restrict referrer-based redirects to local URLs
## Summary

Two error handlers redirect directly to the request referrer without checking whether it belongs to the OpenAlgo origin.

## Evidence

- `app.py:566` redirects to `request.referrer` after a CSRF error.
- `blueprints/admin.py:44` does the same for the admin rate-limit handler.

## Why this matters

An attacker-controlled external referrer can turn an application error path into an open redirect, which can support phishing or confusing post-error navigation.

## Scope

Add or reuse a single safe-local-redirect helper. Accept same-origin relative/local targets, reject external and malformed targets, and fall back to the intended internal route.

## Acceptance criteria

- [ ] External `http`, `https`, scheme-relative, and malformed referrers use the local fallback.
- [ ] Valid same-origin/local referrers remain usable.
- [ ] Both handlers share the same validation logic.
- [ ] Tests cover encoded and scheme-relative examples without making network calls.

## Verification

```powershell
uv run pytest test/test_safe_error_redirects.py -v
uv run ruff check app.py blueprints/admin.py test/test_safe_error_redirects.py
```

## Security guidance

- Difficulty: Easy
- Estimated time: 2-4 hours
- You will learn: open-redirect prevention, URL validation, and Flask handler tests.

lebertmcdonald-afk:
## What changed

Two error handlers redirected straight to `request.referrer` with no validation:

- `app.py:566` (CSRF error handler) — `redirect(request.referrer or url_for("auth.login"))`
- `blueprints/admin.py:44` (admin rate-limit handler) — `redirect(request.referrer or url_for("admin_bp.index"))`

`Referer` is an ordinary, attacker-controllable HTTP header on any request the attacker crafts directly (it isn't limited to same-origin browser navigation), so an external referrer turned these error paths into an open redirect — useful for phishing or for confusing post-error navigation off the site.

This adds a single shared helper, `utils/safe_redirect.safe_local_redirect_target(candidate, fallback, host)`, and both handlers now route their target through it instead of using the referrer directly. It:

- Accepts a same-origin absolute URL (what a real browser's `Referer` actually looks like) or a bare same-origin path, and returns just the local `path[?query][#fragment]`.
- Rejects any other host, `javascript:`/`data:`/other non-http(s) schemes, and malformed input, falling back to the handler's intended internal route instead.
- Rejects protocol-relative targets (`//evil.com`), backslash-disguised ones (`/\evil.com`, still normalized to `//evil.com` by some browsers), and percent-encoded forms of either (`%2F%2Fevil.com`), since a naive check that only inspects the raw string is bypassable through any of these.
- Strips CR/LF and other control characters, closing off header-injection via a crafted referrer.

While wiring the admin handler, I also found its existing fallback, `url_for("admin_bp.index")`, was itself broken — that endpoint was removed when the Jinja admin routes were migrated to React, so calling it raises a `BuildError`. That made the "safe" fallback unreachable in exactly the case this fix needed it (an unsafe or missing referrer), so I pointed it at the current route, `react.react_admin_index`, instead.

## Why I prioritized this

It's a live open-redirect vulnerability (CWE-601) in a self-hosted trading platform's error paths — exactly the kind of issue where a plausible-looking phishing link (`https://your-openalgo-instance/...` triggering a CSRF error with a malicious `Referer`) can redirect a logged-in user off the real site. It's also small and self-contained (one new utility module, two call sites), matching the repo's one-fix-per-PR contribution policy, and doesn't touch broker integrations or require broker credentials to verify.

## Testing

`test/test_safe_error_redirects.py` covers:

- The helper directly: external hosts, protocol-relative, backslash, and percent-encoded variants, `javascript:`/`data:`/`ftp:` schemes, userinfo tricks (`http://host@evil.com`), suffix tricks (`app.example.com.evil.com`), CRLF injection, and — on the accept side — same-origin absolute URLs (including query/fragment) and bare relative paths.
- An end-to-end pass through the actual registered `admin_bp` `429` error handler (a throwaway route triggers `abort(429)`), checking the real `Location` header for an external referrer, a protocol-relative referrer, a same-origin referrer, and a missing referrer.

```
uv run pytest test/test_safe_error_redirects.py -v   # 41 passed
uv run ruff check app.py blueprints/admin.py utils/safe_redirect.py test/test_safe_error_redirects.py   # all checks passed
uv run pytest test/ -q   # no regressions vs. main (same pre-existing failures/errors on both, from missing optional deps - eventlet, pytest-asyncio - unrelated to this change)
```

Closes #1852

---
## Summary by cubic
Rejects cross-origin referrers in CSRF and admin rate-limit error redirects to close an open-redirect vulnerability. Previously both handlers redirected to request.referrer; now they validate and only allow same-origin paths, falling back to the intended internal route.

- Add `utils/safe_redirect.safe_local_redirect_target(candidate, fallback, host)` to normalize and validate targets, returning a local path.
- Use the helper in the CSRF handler and the admin rate-limit handler instead of the raw referrer.
- Reject cross-origin hosts, non-http(s) schemes, protocol-relative (`//...`), backslash/percent-encoded variants, and control chars; accept same-origin absolute URLs and local paths.
- Fix the admin fallback: replace `url_for("admin_bp.index")` (removed endpoint) with `url_for("react.react_admin_index")` to avoid `BuildError`.
- Add tests for helper edge cases and end-to-end admin handler behavior.

<sup>Written for commit 2bf6d7a1910c2536c02706035c51047adbf1461d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1882?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app.py
+++ b/app.py
@@ -548,6 +548,8 @@ def create_app():
         """Custom handler for CSRF errors (400 Bad Request)"""
         from flask import flash, jsonify, redirect, request, url_for
 
+        from utils.safe_redirect import safe_local_redirect_target
+
         error_description = str(error)
 
         logger.warning(f"CSRF Error on {request.path}: {error_description}")
@@ -563,7 +565,9 @@ def create_app():
                 ), 400
             else:
                 flash("Security token expired. Please try again.", "error")
-                return redirect(request.referrer or url_for("auth.login"))
+                fallback = url_for("auth.login")
+                target = safe_local_redirect_target(request.referrer, fallback, request.host)
+                return redirect(target)
 
         # For other 400 errors
         return str(error), 400
--- a/blueprints/admin.py
+++ b/blueprints/admin.py
@@ -27,6 +27,7 @@ from database.qty_freeze_db import (
 from database.qty_freeze_db import db_session as freeze_db_session
 from limiter import limiter
 from utils.logging import get_logger
+from utils.safe_redirect import safe_local_redirect_target
 from utils.session import check_session_validity
 
 logger = get_logger(__name__)
@@ -41,7 +42,11 @@ admin_bp = Blueprint("admin_bp", __name__, url_prefix="/admin")
 def ratelimit_handler(e):
     """Handle rate limit exceeded errors"""
     flash("Rate limit exceeded. Please try again later.", "error")
-    return redirect(request.referrer or url_for("admin_bp.index"))
+    # "admin_bp.index" doesn't exist: the Jinja admin routes were migrated to
+    # React, and the admin dashboard is now served by react.react_admin_index.
+    fallback = url_for("react.react_admin_index")
+    target = safe_local_redirect_target(request.referrer, fallback, request.host)
+    return redirect(target)
 
 
 # ============================================================================
--- a/test/test_safe_error_redirects.py
+++ b/test/test_safe_error_redirects.py
@@ -1,0 +1,147 @@
+"""Regression tests for the CSRF and admin rate-limit error redirects.
+
+Both handlers used to redirect straight to ``request.referrer``. The Referer
+header is attacker-controlled on any request the attacker crafts directly, so
+an external, protocol-relative, or backslash-disguised referrer turned the
+error path into an open redirect. Both handlers now resolve their target
+through ``utils.safe_redirect.safe_local_redirect_target`` instead.
+"""
+
+import os
+import sys
+
+import pytest
+from flask import Flask, abort
+
+sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+
+import blueprints.admin as admin_bp_module  # noqa: E402
+import blueprints.react_app as react_app_module  # noqa: E402
+from utils.safe_redirect import safe_local_redirect_target  # noqa: E402
+
+HOST = "app.example.com"
+
+
+# --- safe_local_redirect_target: unit coverage -----------------------------
+
+
+@pytest.mark.parametrize(
+    "candidate",
+    [
+        None,
+        "",
+        "   ",
+        "http://evil.com",
+        "https://evil.com/phish",
+        "http://evil.com/app.example.com",  # host is only a path segment here
+        "HTTP://EVIL.COM",
+        "http://app.example.com.evil.com/phish",  # suffix trick, not a subdomain match
+        "http://app.example.com@evil.com/phish",  # userinfo trick
+        "//evil.com",
+        "//evil.com/phish",
+        "//app.example.com/phish",  # protocol-relative is never a real Referer
+        "/\\evil.com",
+        "\\/evil.com",
+        "\\\\evil.com",
+        "%2F%2Fevil.com",
+        "%2f%2Fevil.com/phish",
+        "/%5c%5cevil.com",
+        "javascript:alert(1)",
+        "data:text/html,<script>alert(1)</script>",
+        "ftp://app.example.com/phish",
+        "https:evil.com",
+        "https:///evil.com",
+        "relative/without/leading/slash",
+        "/redirect\r\nSet-Cookie: pwned=1",
+        "/redirect\nLocation: http://evil.com",
+    ],
+)
+def test_unsafe_targets_fall_back(candidate):
+    assert safe_local_redirect_target(candidate, "/fallback", HOST) == "/fallback"
+
+
+@pytest.mark.parametrize(
+    ("candidate", "expected"),
+    [
+        ("/dashboard", "/dashboard"),
+        ("/settings/profile", "/settings/profile"),
+        ("/search?q=orders", "/search?q=orders"),
+        ("/search?q=hello%20world", "/search?q=hello world"),
+        ("/orders#history", "/orders#history"),
+        ("  /dashboard  ", "/dashboard"),
+        # Real browsers send an absolute URL, not a bare path.
+        ("http://app.example.com/dashboard", "/dashboard"),
+        ("https://app.example.com/settings/profile?tab=security", "/settings/profile?tab=security"),
+        ("HTTP://APP.EXAMPLE.COM/dashboard", "/dashboard"),
+        ("http://app.example.com", "/"),
+    ],
+)
+def test_safe_same_origin_targets_are_preserved(candidate, expected):
+    assert safe_local_redirect_target(candidate, "/fallback", HOST) == expected
+
+
+def test_missing_referrer_uses_fallback():
+    assert safe_local_redirect_target(None, "/admin", HOST) == "/admin"
+
+
+# --- admin_bp rate-limit handler: end-to-end wiring -------------------------
+
+
+# admin_bp has no active route left that raises 429 directly (they were all
+# migrated to React), so add a throwaway one purely to exercise the
+# blueprint's own @admin_bp.errorhandler(429) through a real request. This
+# has to happen once at import time: Flask locks a blueprint against further
+# route additions after its first registration to an app.
+@admin_bp_module.admin_bp.route("/_test/trigger-429")
+def _trigger_429():
+    abort(429)
+
+
+def _admin_app():
+    app = Flask(__name__)
+    app.secret_key = "test-secret"
+    app.register_blueprint(admin_bp_module.admin_bp)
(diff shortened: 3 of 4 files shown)


marketcalls:
Thanks for the report and the patch. The underlying issue is real and I want the fix. `app.py:566` and `blueprints/admin.py:44` are the only two `request.referrer` redirects in the tree, both are unvalidated, and the CSRF one is genuinely browser reachable: a cross-site auto-submitting form POST to any non-`/api` route fails CSRF, falls into the `else` branch at `app.py:565`, and 302s the victim's browser back to the attacker's origin. You are also right that `admin_bp.index` no longer exists (it is commented out at `blueprints/admin.py:54-63`), so the old fallback raised `BuildError`. And your backslash normalization is a real improvement over the textbook `urljoin` check, which happily treats `/\evil.com` as safe.

Here is what has to change before I can merge.

**1. `utils/safe_redirect.py` still emits a protocol-relative target.** That is the exact class the docstring and the PR body say it rejects. Running the shipped module:

```
safe_local_redirect_target("http://app.example.com//evil.com", "/fallback", "app.example.com")
  -> "//evil.com"
safe_local_redirect_target("https://app.example.com//evil.com/phish", "/fallback", "app.example.com")
  -> "//evil.com/phish"
```

The `normalized.startswith("//")` guard on line 55 only inspects the whole candidate string. Once the same-origin absolute branch is taken you rebuild the target from `parts.path`, and that path is never re-checked. Werkzeug passes it through verbatim, I confirmed `redirect("//evil.com")` emits `Location: //evil.com`, which a browser reads as a network-path reference (RFC 3986 4.2) and navigates to `http://evil.com`. Note that `///evil.com` is not affected, Werkzeug collapses that one to `/evil.com`, so the guard you need is specifically on a rebuilt path beginning with `//`.

**2. Drop the blanket `unquote()` on line 46.** Decoding in order to validate is correct, returning the decoded string is not:

```
"/search?q=a%26b=c"  ->  "/search?q=a&b=c"    (one query parameter becomes two)
"/orders%23history"  ->  "/orders#history"    (a literal '#' becomes a fragment)
```

Werkzeug re-encodes a bare space on the way out, so the `%20` case in your test happens to be harmless on the wire, but reserved characters are not re-encoded and the meaning of the URL changes.

For both of the above, the shape I would rather see: keep your backslash pre-check, then resolve with `urljoin(request.host_url, candidate)`, compare `urlsplit(...).hostname` against the request host, and redirect to the resolved absolute URL instead of rebuilding a path by hand. I tested that pattern against your own table and it rejects `//evil.com`, `https://evil.com/phish`, `javascript:`, and the `host@evil.com` userinfo trick, while leaving percent-encoding untouched. It does miss `/\evil.com`, which is why your backslash step stays.

**3. `blueprints/admin.py:47` turns a rate limit into a 500 on installs without `frontend/dist`.** `url_for("react.react_admin_index")` is evaluated eagerly on every 429, but `react_bp` is registered conditionally at `app.py:265-270` behind `is_react_frontend_available()`, and a missing build is a supported, warn-only state (notably on a feature branch CI has not built). I reproduced it: with `react_bp` unregistered the handler raises `werkzeug.routing.exceptions.BuildError` and returns 500 for both a same-origin `Referer` and a missing one. To be fair, the old code also 500'd when the `Referer` was absent, so on a normal install this is still a net improvement, but for a same-origin `Referer` on a dist-less install it is a regression. Use the literal `"/admin"`, or wrap the `url_for` in `try/except BuildError`.

**4. While you are in that handler, fix the response shape.** Every route left on `admin_bp` is `/admin/api/*` and is consumed by the typed JSON client in `frontend/src/api/admin.ts`. Returning a 302 to `/admin` hands those callers an HTML body and a parse error instead of a rate-limit signal. `app.py:558` already shows the pattern I want: `if request.is_json or request.path.startswith("/api"): return jsonify(...), 429`. Without that branch, this handler stays broken for the only routes that can reach it, which undercuts the reason you gave for touching it. (Related and not your job in this PR: that `startswith("/api")` check does not match `/admin/api/...`.)

**5. `utils/safe_redirect.py` has no logger.** Repo convention is `logger = get_logger(__name__)` from `utils/logging.py` in every module. A validator that silently swallows its input is hard to diagnose, and there is a real config where it fires on legitimate traffic: our nginx templates set `proxy_set_header Host $host` (`install/change-domain.sh:552`), and nginx's `$host` drops a non-default port, so an install served on a non-standard port sees `Referer` netloc `example.com:8443` against `request.host` `example.com` and silently bounces the user to the login page. A `logger.debug` on rejection is enough. Comparing hostnames rather than netlocs would remove the cause.

**6. `test/test_safe_error_redirects.py`:**

- The unsafe-candidate table has no case of the form `http://app.example.com//evil.com`. That is why 41 tests pass while the property they exist to certify does not hold. Add that and `https://app.example.com//evil.com/phish`.
- Line 69 asserts `"/search?q=hello%20world"` maps to `"/search?q=hello world"`. That bakes the decoding bug in as expected behaviour. Replace it with a case that round-trips `%26` unchanged.
- Lines 95 to 99 add a route to the shared `blueprints.admin.admin_bp` object at import time. Flask 3.1.3 raises `AssertionError: The setup method 'route' can no longer be called on the blueprint 'admin_bp'. It has already been registered at least once` the moment anything else registers it, so the first test that imports `app.py` will break collection of this module. `test/test_logout_csrf.py:33-38` is the pattern to follow: fresh app per test, never mutate the shared blueprint. Put the trigger route on a throwaway local Blueprint.
- The file is not in the CI-safe list in `.github/workflows/ci.yml` (lines 47 to 52), so none of it runs in CI. I checked that it can run there: it needs no broker credentials, and `test/conftest.py` already assigns `DATABASE_URL`, so the `blueprints.admin` import (which builds SQLAlchemy engines at module scope) is fine. Please add it.
- Add one end-to-end test for the `app.py` CSRF 400 handler. That is the reachable open redirect of the two. The admin 429 handler is XHR-only, so its open-redirect value is close to zero.

On process: CI has not actually run on this branch. Fork PRs need maintainer approval to run workflows here, so the green checks you are seeing are Snyk reporting no manifest changes plus the cubic reviewer. Ruff, the backend test job and the frontend build have all been skipped, so treat nothing on this branch as machine-validated yet.

No migration is needed for this change, and it does not touch any of the eventlet, ZeroMQ, NullPool or FD-hygiene invariants, so the list above is the whole gate.

marketcalls (commented):
Reviewed head `2bf6d7a1910c2536c02706035c51047adbf1461d`: 73/100. Please address the following before merge:

- `utils/safe_redirect.py:78`: Rebuilt path is not checked for a leading double slash. A same-host URL passes the host check, but stripping its origin creates an external Location; the sanitizer violates the property it was added to enforce. Validate the final emitted path against // and backslash authority forms, or emit a validated same-origin absolute URL. Add same-host absolute // and percent-encoded slash regression cases. Evidence: Exact head helper returns //evil.example/phish for https://app.example.com//evil.example/phish; Flask redirect emits that exact external Location.

- `utils/safe_redirect.py:46`: Unquote changes the URL that is returned. Reserved escaped characters change query/path meaning: /search?q=a%26b=c becomes /search?q=a&b=c, and /orders%23history becomes /orders#history. Use decoding only for rejection checks and preserve the original percent-encoded URL when constructing the returned redirect; assert reserved characters round-trip. Evidence: unquote(candidate.strip()) is parsed and parts.path/query are directly used for the returned value.

- `blueprints/admin.py:47`: Fallback url_for is evaluated even when the React blueprint is unavailable. An otherwise valid same-origin Referer now raises BuildError on supported installations without frontend/dist, turning the admin 429 handler into 500. Use a stable local /admin fallback or handle missing React route registration before resolving the target. Add a minimal app test without react_bp. Evidence: app.py conditionally registers react_bp only if is_react_frontend_available(); the new eager url_for("react.react_admin_index") precedes referrer validation.

PR #1884, opened 2026-08-24
fix: validate SIP inputs before loading prices

### LINKED ISSUE #1823: python: Validate SIP service inputs before loading prices
## Summary

The SIP service converts investment amount without a controlled error and loads prices before finishing SIP-day validation.

## Evidence

- `services/sip_service.py:171` calls `float()` without a local validation path.
- `services/sip_service.py:219` loads prices before completing day validation.

## Scope

Validate amount and SIP day before any price lookup. Return the normal client-error result and test with a mocked loader.

## Acceptance criteria

- [ ] Malformed, non-finite, zero, and negative amounts return client errors.
- [ ] Invalid days do not call the price loader.
- [ ] Valid requests retain current behavior.
- [ ] Boundary cases are covered.

## Verification

```powershell
uv run pytest test/test_sip_service_validation.py -v
uv run ruff check services/sip_service.py test/test_sip_service_validation.py
```

- Difficulty: Easy
- Estimated time: 1-3 hours
- You will learn: fail-fast validation and dependency mocking.

siddharthg2309:
## Summary

Validate SIP investment amounts and SIP days before loading historical prices.

## Related issue

Closes #1823

## Changes

- Safely convert the SIP amount once at the service boundary.
- Reject malformed, non-finite, zero, negative, and overflow amounts with a `400` client error.
- Validate monthly and quarterly SIP days before the price loader runs.
- Require a monthly or quarterly SIP day to be an integer from `1` through `28`.
- Reuse the validated amount throughout the remaining calculation.

## Coverage

- Invalid amounts do not reach the price loader.
- Invalid monthly and quarterly SIP days do not reach the price loader.
- Amount coverage includes malformed input, `NaN`, infinities, overflow, zero, and negatives.
- SIP-day coverage includes both boundaries (`1`, `28`), out-of-range days, and wrong input types.
- Valid boundary requests load prices and complete successfully.

## Testing

- [x] `env -u LOG_FORMAT uv run pytest test/test_sip_service_validation.py -v`
  - 19 passed
- [x] `uv run ruff check services/sip_service.py test/test_sip_service_validation.py`
  - passed
- [x] `git diff --check`
  - passed

## Local environment note

This machine sets `LOG_FORMAT=json`, which conflicts with the repository logging setup during test collection. The focused test was run with that local-only environment variable removed; no project configuration was changed.

## Checklist

- [x] Amount and SIP-day validation occur before any price lookup.
- [x] Invalid calls are covered with a mocked price loader.
- [x] Valid boundary requests retain existing behavior.
- [x] No broker credentials or sensitive data are included.

---
## Summary by cubic
Validate SIP amount and SIP day at the service boundary before any price load. Previously, malformed amounts or SIP days could reach the price loader; now the service returns 400 with clear messages and skips price fetches.

- Parse and validate the amount once; reject non-numeric, non-finite (NaN/±Inf), overflow, zero, and negative values with 400 "amount must be a positive number"; reuse the validated amount across results, grids, and benchmarks.
- For "monthly" and "quarterly", require `day_of_month` to be an integer 1–28; explicitly reject bool; return 400 "day_of_month must be an integer between 1 and 28".
- Bound SIP-day validation using `sip.schedule.MAX_DAY_OF_MONTH` to keep behavior consistent with scheduling rules.
- Tests verify the price loader is not called on invalid inputs and that boundary days (1, 28) succeed with no behavior changes.
- Client requirement: send a positive numeric `amount` and, for monthly/quarterly SIPs, an integer `day_of_month` in 1–28.

<sup>Written for commit 25490b2006ff6ce93f99669461e8f74d8d0fc55c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1884?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/services/sip_service.py
+++ b/services/sip_service.py
@@ -48,6 +48,7 @@ from utils.logging import get_logger
 logger = get_logger(__name__)
 
 MAX_SIP_YEARS = 30
+INVALID_AMOUNT_MESSAGE = "amount must be a positive number"
 
 
 def _error(message: str, status: int = 400) -> tuple[bool, dict[str, Any], int]:
@@ -145,9 +146,12 @@ def run_sip_backtest(
             (the base symbol -- ``INFY``, ``SBIN``, ``TATAMOTORS``).
         start_date, end_date: ``YYYY-MM-DD``. ``start_date`` is the date the
             SIP begins; the first installment lands on or after it.
-        amount: the installment before any step-up.
+        amount: the installment before any step-up. Direct callers must provide
+            a finite, positive number or receive a 400 response.
         frequency: weekly, fortnightly, monthly or quarterly.
-        day_of_month: target day for monthly and quarterly SIPs, 1-28.
+        day_of_month: target day for monthly and quarterly SIPs. Direct callers
+            must provide an integer from 1 through 28 or receive a 400 response
+            before prices are loaded.
         step_up_percent: annual increase applied on each anniversary.
         brokerage_percent, brokerage_flat: charged per installment.
         benchmark, benchmark_exchange: an index to run the *same schedule*
@@ -172,9 +176,9 @@ def run_sip_backtest(
     try:
         investment_amount = float(amount)
     except (TypeError, ValueError, OverflowError):
-        return _error("amount must be a positive number")
+        return _error(INVALID_AMOUNT_MESSAGE)
     if not math.isfinite(investment_amount) or investment_amount <= 0:
-        return _error("amount must be a positive number")
+        return _error(INVALID_AMOUNT_MESSAGE)
     if frequency not in FREQUENCIES:
         return _error(f"frequency must be one of {', '.join(FREQUENCIES)}")
     if frequency in ("monthly", "quarterly") and (
--- a/services/sip_service.py
+++ b/services/sip_service.py
@@ -15,6 +15,7 @@ for the same question.
 from __future__ import annotations
 
 import dataclasses
+import math
 from datetime import date, datetime
 from typing import Any
 
@@ -40,7 +41,7 @@ from sip.analytics import (
 )
 from sip.crisis import crisis_sip_analysis, crisis_summary
 from sip.engine import SipError, run_sip
-from sip.schedule import FREQUENCIES, ScheduleError
+from sip.schedule import FREQUENCIES, MAX_DAY_OF_MONTH, ScheduleError
 from sip.xirr import xirr_or_none
 from utils.logging import get_logger
 
@@ -168,10 +169,20 @@ def run_sip_backtest(
         return _error("end_date must be after start_date")
     if (end - start).days > MAX_SIP_YEARS * 366:
         return _error(f"SIP window cannot exceed {MAX_SIP_YEARS} years")
-    if amount is None or float(amount) <= 0:
+    try:
+        investment_amount = float(amount)
+    except (TypeError, ValueError, OverflowError):
+        return _error("amount must be a positive number")
+    if not math.isfinite(investment_amount) or investment_amount <= 0:
         return _error("amount must be a positive number")
     if frequency not in FREQUENCIES:
         return _error(f"frequency must be one of {', '.join(FREQUENCIES)}")
+    if frequency in ("monthly", "quarterly") and (
+        isinstance(day_of_month, bool)
+        or not isinstance(day_of_month, int)
+        or not 1 <= day_of_month <= MAX_DAY_OF_MONTH
+    ):
+        return _error(f"day_of_month must be an integer between 1 and {MAX_DAY_OF_MONTH}")
     exchange = (exchange or "NSE").upper()
     if exchange not in SUPPORTED_EXCHANGES:
         return _error(
@@ -225,7 +236,7 @@ def run_sip_backtest(
 
     try:
         result = run_sip(
-            closes, symbol, exchange, start, end, float(amount),
+            closes, symbol, exchange, start, end, investment_amount,
             warnings=[str(w) for w in warnings], **sip_kw,
         )
     except (SipError, ScheduleError) as exc:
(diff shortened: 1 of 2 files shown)


marketcalls:
**Decision: merge after three small fixes.** The change is correct, the tests are real, and it implements issue #1823 line for line. I want the happy-path assertion loosened and the duplicated error string collapsed before it lands. Thanks for the clean, focused diff.

## What I verified

CI has not run on this branch. Fork PRs wait on maintainer approval, so the only green checks you see are Snyk (reporting "no manifest changes detected") and cubic. No Ruff, no pytest, no frontend build has touched this branch, so I ran it locally against `main` at `a1a2dc9`:

* Patch applies cleanly.
* `uv run pytest test/test_sip_service_validation.py` gives 19 passed.
* Reverting only `services/sip_service.py` and rerunning gives 13 failed, 6 passed, so the tests genuinely exercise the new guards rather than passing either way.
* `uv run ruff check services/sip_service.py test/test_sip_service_validation.py` is clean.
* The diff is pure ASCII, zero non-ASCII bytes. (The two en dashes in "1-28" live in cubic's auto-generated block in the PR body, not in your code.)
* The defect is real at the service boundary: on `main`, `services/sip_service.py:171` calls `float(amount)` bare, so `float("invalid")` escapes as an uncaught `ValueError`, and an out-of-range `day_of_month` reaches `load_prices` and only surfaces later as a 422 from `ScheduleError` at `services/sip_service.py:231-232`.

One scope note so we are both honest about the blast radius. The only production caller is `restx_api/sip.py:174`, and its marshmallow schema already rejects every case you guard. I loaded `SipBacktestSchema` under marshmallow 3.26.2 and confirmed that `"invalid"`, `None`, `"nan"`, `"inf"`, `float("nan")`, `float("inf")`, `float("-inf")` and `10**1000` are all rejected by `amount = fields.Float(...)` at `restx_api/sip.py:64-66`, and `day_of_month` 0 and 29 by `fields.Int(..., validate.Range(min=1, max=MAX_DAY_OF_MONTH))` at `restx_api/sip.py:72-74`. `test/test_sip_api_security.py:126-129` already asserts 400 with `called == []` for `day_of_month` 31 and 0 and `amount` 0 and -100, and the UI goes through the same endpoint (`frontend/src/api/sip.ts:250`). So this is defense-in-depth on a public service function, not a reachable bug in a deployed instance. That is what issue #1823 asked for, and the overstatement is in my issue text, not yours. The change stays.

## Required before merge

**1. `test/test_sip_service_validation.py:105-107`, loosen the happy-path assertion.**

```python
    assert calls == [
        ((["INFY"], ["NSE"], "2024-01-01", "2024-04-30"), {"source": "db", "api_key": None, "auth_token": None, "feed_token": None, "broker": None})
    ]
```

This mirrors the creds dict built at `services/sip_service.py:185-190` verbatim. Adding one credential field there fails a test whose stated subject is input validation, which is a maintenance cost I do not want to carry for a guard test. Assert on what the test is actually about:

```python
    assert len(calls) == 1
    args, kwargs = calls[0]
    assert args == (["INFY"], ["NSE"], "2024-01-01", "2024-04-30")
    assert kwargs["source"] == "db"
```

**2. `services/sip_service.py:175` and `:177`, one message, one place.**

Both lines return the literal `"amount must be a positive number"`, and `test/test_sip_service_validation.py:57` asserts on that exact string. Two copies four lines apart will drift. Pull it into a module constant next to `MAX_SIP_YEARS` and use it from both returns.

**3. `services/sip_service.py:149`, update the docstring.**

The `day_of_month` Args entry still reads "target day for monthly and quarterly SIPs, 1-28" and says nothing about rejection. For a direct caller the status code changed: an out-of-range `day_of_month` used to come back as a 422 from `ScheduleError` and now returns 400 before any price load. Record that, and record that a non-numeric or non-finite `amount` is a 400 as well. This is the public service contract and the docstring is where it is written down.

## Do not do these

**Do not add the file to `.github/workflows/ci.yml`.** It does matter, and you may not have noticed it: the `backend-test` job runs a hand-curated file list at `.github/workflows/ci.yml:47-52`, and `test/test_sip_service_validation.py` is not on it, so as things stand it will never run in CI. (Neither does the existing `test/test_sip_api_security.py`, which is on me.) I will add your file plus the new test files from #1861 and #1875 in a single commit on `main` after this batch merges. Four fork PRs each editing the same continuation line would conflict three ways. Note the backend rule is specific: frontend tests need no registration, because the `frontend-test` job runs the whole vitest suite, so anything under `frontend/src/**/*.test.tsx` runs automatically. Backend is opt-in.

**Do not run `uv run ruff format` on `services/sip_service.py`.** That file already fails `ruff format --check` on `main` before your patch, so formatting it would add 75 lines of unrelated reformatting churn (45 added, 30 removed) to a bug fix. Run the formatter on `test/test_sip_service_validation.py` only. It rewraps the 148-character line 106 and the call spanning lines 98-100.

## Optional, no action needed

The `day_of_month` guard at `services/sip_service.py:180-185` is stricter than the HTTP layer can deliver: marshmallow coerces `"1"` and `1.5` to `1` before the service sees them, so the `isinstance` half only ever fires for direct Python callers. That is fine as written. Worth a word in the docstring from item 3 so nobody later "fixes" it by loosening it.

## Related PRs

#1861 (`market_calendar_service`), #1862 (`api_analyzer`) and #1875 (`history_service`) are the same category of change: fail-fast validation inside a service function that the public route already guards. No file overlap with this one and no rebase needed, so ordering does not matter as long as none of them touches `ci.yml`. Of the four, this is the strongest on test quality and scope.

Push the three fixes and I will merge.

siddharthg2309:
@marketcalls 
Thanks for the detailed review. I’ve pushed the three requested updates.

The focused test suite still passes (19 passed) and Ruff is clean. I left ci.yml untouched and formatted only the test file as requested.

PR #1885, opened 2026-08-24
fix(python): key multi-option Greeks batch state by leg index

### LINKED ISSUE #1819: python: Key multi-option Greeks batch state by symbol and exchange
## Summary

The multi-option Greeks batch service stores intermediate values under symbol-only keys. Identical symbol text on two exchanges can overwrite state and return the wrong quote or Greeks result.

## Evidence

- `services/option_greeks_service.py:1004-1015` builds batch lookup state.
- `services/option_greeks_service.py:1066-1168` reads and writes dictionaries using only the symbol.

## Scope

Use a stable composite key such as `(symbol, exchange)` throughout the batch calculation. Keep the public response shape unchanged and add a credential-free regression test.

## Acceptance criteria

- [ ] Both exchange entries retain their own prices and option data.
- [ ] Existing single-exchange behavior is unchanged.
- [ ] A regression test fails with symbol-only keys and passes with composite keys.

## Verification

```powershell
uv run pytest test/test_multi_option_greeks_regression.py -v
uv run ruff check services/option_greeks_service.py test/test_multi_option_greeks_regression.py
```

- Difficulty: Medium
- Estimated time: 2-4 hours
- You will learn: composite keys, batch state, and regression testing.

Pragitics:
## What this does

`get_multi_option_greeks()` stored its intermediate batch state under symbol-only keys, so two legs of one request could overwrite each other and be priced from the wrong entry. No error surfaced - the caller simply received wrong Greeks.

Per-leg state is now keyed by the leg's index in the request, which cannot collide.

## Why

Closes #1819

## Validation status

**CI has never run on this branch.** The workflows are waiting on maintainer approval, so the only evidence is local: my runs and the reviewer's independent reproduction. Everything below was run on this machine.

## Approach

The issue asked for `(symbol, exchange)`. That was the first commit, and review showed it does not close the defect class: two legs sharing a symbol *and* an exchange but differing by `underlying_symbol` / `underlying_exchange` still collapse onto one key, and that is the endpoint's own documented example at `restx_api/multi_option_greeks.py:53`. The second commit moves to the leg index.

| State | Key | Why |
| --- | --- | --- |
| `parsed_symbols` | leg index | per leg |
| `symbol_to_spot_key` | leg index | per leg - this is what differs in the documented example |
| `symbol_forward` | leg index | per leg |
| `option_prices` | `(symbol, exchange)` | a price belongs to a contract, not to a leg |

Two things fall out of the index key:

**The sort disappears.** `results` is now an indexed slot list, so every leg lands in request order by construction. `symbol_order` and the final `results.sort()` are gone. This also means the reviewer's note about the sort key being safe no longer applies to anything.

**The quote fetch dedupes.** Two legs sharing a contract need one quote, not two, so `seen_quotes` skips the repeat.

## Multiquotes exchange fallback

`option_prices` takes the exchange from the broker result, falling back to the requested exchange when the result omits it, so a future adapter degrades one leg rather than failing the whole batch.

A symbol requested on *two* exchanges maps to `None` rather than to either one. Guessing would silently reintroduce the collision this change exists to remove; leaving those legs unpriced is the safer failure. Covered by `test_multi_option_greeks_does_not_guess_ambiguous_exchange`.

To be accurate about where the guarantee lives: `services/quotes_service.py` sets `exchange` itself only on the paths it builds - the invalid-symbol entry (`:275`) and the per-symbol fallback (`:290`, `:297`). On the broker-native path it passes the adapter's list through untouched (`:311`). So the guarantee that `exchange` is present comes from the 35 adapters under `broker/*/api/data.py`, not from `quotes_service.py`. It holds in all 35 today, which is why the fallback is hardening rather than a bug fix. An earlier version of this description got that wrong.

## CI registration

`test/test_multi_option_greeks_regression.py` was not in the `backend-test` file list in `.github/workflows/ci.yml`, so the test would never have run upstream. It is now registered. The file predates this PR and was already unregistered on main.

## The response shape is unchanged

All the batch dictionaries are local to this function and never escape it. Verified by diffing the full JSON response for a single-exchange batch against the same run on main - byte-identical. The test also pins the top-level keys with `assert set(response) == {"status", "data", "summary"}`.

## How I tested

```
uv run pytest test/test_multi_option_greeks_regression.py -v
6 passed

uv run pytest <the 11 CI files> test/test_multi_option_greeks_regression.py --timeout=60
46 passed

uv run ruff check services/option_greeks_service.py test/test_multi_option_greeks_regression.py
All checks passed!

uv run pytest test/ -q
No new failures compared to the same run on main.
```

Six tests, none needing broker credentials or network access. Each one fails when its own fix is reverted - I checked them individually rather than assuming:

| Test | Guards |
| --- | --- |
| `..._keeps_per_exchange_state_for_duplicate_symbols` | the original #1819 defect; fails with `assert 63.25 == 41.5` |
| `..._keeps_per_leg_underlying_for_identical_symbols` | the documented-example case; fails with `[24500.0, 24500.0]` |
| `..._falls_back_to_requested_exchange` | a result that omits `exchange` |
| `..._does_not_guess_ambiguous_exchange` | 3 legs in, the one unambiguous leg still prices |
| `..._converts_expired_legs_to_zero_greeks` | pre-existing |
| `..._treats_naive_expiry_as_ist` | pre-existing |

The mock that contradicted production is gone. `get_underlying_exchange` is no longer patched, so it runs for real and returns NSE for both legs exactly as it would in production; the option-price assertions still pin the defect.

## Not covered

- Nothing else in this 1199-line file
- No schema or migration; no build artifacts
- Independent of #1866: different files, no overlap

## Checklist

- [x] One fix per PR
- [x] Conventional commit messages
- [x] Public response shape unchanged
- [x] Every new test fails without its corresponding fix
- [x] Tests need no broker credentials or network access
- [x] `ruff check` clean on the changed files
- [x] New test registered in CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,6 +50,7 @@ jobs:
             test/test_auth_resume.py test/test_auth_upsert_multisession.py \
             test/sandbox/test_execution_backlog.py test/test_event_bus_bounded.py \
             test/test_telegram_api_contract.py \
+            test/test_multi_option_greeks_regression.py \
             -v --timeout=60
 
   # Frontend linting
--- a/services/option_greeks_service.py
+++ b/services/option_greeks_service.py
@@ -1001,18 +1001,19 @@ def get_multi_option_greeks(
     failed_count = 0
 
     # Step 1: Parse all symbols and group by underlying for spot price fetch
-    parsed_symbols = {}  # symbol -> (base_symbol, expiry, strike, opt_type)
+    parsed_symbols = {}  # (symbol, exchange) -> (base_symbol, expiry, strike, opt_type)
     spot_keys = {}  # (spot_symbol, spot_exchange) -> spot_price
-    symbol_to_spot_key = {}  # symbol -> (spot_symbol, spot_exchange)
-    symbol_forward = {}  # symbol -> Black-76 forward (synthetic future)
+    symbol_to_spot_key = {}  # (symbol, exchange) -> (spot_symbol, spot_exchange)
+    symbol_forward = {}  # (symbol, exchange) -> Black-76 forward (synthetic future)
     synth_cache = {}  # (base, underlying_exchange, expiry_code) -> synthetic future
 
     for sym_req in symbols:
         symbol = sym_req.get("symbol")
         exchange = sym_req.get("exchange")
         try:
+            batch_key = (symbol, exchange)
             base_symbol, expiry, strike, opt_type = parse_option_symbol(symbol, exchange, expiry_time)
-            parsed_symbols[symbol] = (base_symbol, expiry, strike, opt_type)
+            parsed_symbols[batch_key] = (base_symbol, expiry, strike, opt_type)
 
             # Determine spot symbol/exchange for this option (also the fallback
             # if the synthetic future cannot be computed for an index weekly)
@@ -1021,7 +1022,7 @@ def get_multi_option_greeks(
             spot_exchange = sym_req.get("underlying_exchange") or get_underlying_exchange(base_symbol, exchange)
             spot_key = (spot_symbol, spot_exchange)
             spot_keys[spot_key] = None  # will be filled with price
-            symbol_to_spot_key[symbol] = spot_key
+            symbol_to_spot_key[batch_key] = spot_key
 
             # Black-76 forward = per-expiry synthetic future for all F&O; spot
             # fallback. Skipped if the caller passed a custom underlying
@@ -1031,7 +1032,7 @@ def get_multi_option_greeks(
                     base_symbol, exchange, spot_exchange, expiry, api_key, synth_cache
                 )
                 if forward:
-                    symbol_forward[symbol] = forward
+                    symbol_forward[batch_key] = forward
         except Exception as e:
             logger.warning(f"Failed to parse symbol {symbol}: {e}")
             failed_count += 1
@@ -1063,13 +1064,14 @@ def get_multi_option_greeks(
     option_symbols_to_fetch = []
     for sym_req in symbols:
         symbol = sym_req.get("symbol")
-        if symbol in parsed_symbols:  # only if parsing succeeded
+        exchange = sym_req.get("exchange")
+        if (symbol, exchange) in parsed_symbols:  # only if parsing succeeded
             option_symbols_to_fetch.append({
                 "symbol": symbol,
-                "exchange": sym_req.get("exchange"),
+                "exchange": exchange,
             })
 
-    option_prices = {}  # symbol -> ltp
+    option_prices = {}  # (symbol, exchange) -> ltp
     if option_symbols_to_fetch:
         logger.info(f"Batch fetching {len(option_symbols_to_fetch)} option prices via multiquotes")
         try:
@@ -1079,10 +1081,11 @@ def get_multi_option_greeks(
             if mq_success and "results" in mq_response:
                 for result in mq_response["results"]:
                     sym = result.get("symbol")
+                    exch = result.get("exchange")
                     if sym and "data" in result:
                         ltp = result["data"].get("ltp")
                         if ltp:
-                            option_prices[sym] = ltp
+                            option_prices[(sym, exch)] = ltp
         except Exception as e:
             logger.warning(f"Multiquotes fetch failed: {e}")
 
@@ -1090,15 +1093,16 @@ def get_multi_option_greeks(
     for sym_req in symbols:
         symbol = sym_req.get("symbol")
         exchange = sym_req.get("exchange")
+        batch_key = (symbol, exchange)
 
         # Skip if already failed during parsing
-        if symbol not in parsed_symbols:
+        if batch_key not in parsed_symbols:
             continue
 
         # Forward price: synthetic future for index weeklies, else spot.
         # Falls back to spot if the synthetic future could not be computed.
-        spot_key = symbol_to_spot_key.get(symbol)
-        spot_price = symbol_forward.get(symbol) or (spot_keys.get(spot_key) if spot_key else None)
+        spot_key = symbol_to_spot_key.get(batch_key)
+        spot_price = symbol_forward.get(batch_key) or (spot_keys.get(spot_key) if spot_key else None)
         if not spot_price:
             failed_count += 1
             results.append({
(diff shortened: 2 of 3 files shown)


marketcalls:
Good work on this one. The defect is real and I reproduced it independently: on a clean worktree at `a1a2dc926`, reverting only `services/option_greeks_service.py` and running your new test gives exactly `assert 63.25 == 41.5`, and with the patch all three tests pass in 0.50s with no credentials and no network. I also walked every use site of the five dictionaries and every `get_multiquotes` implementation under `broker/*/api/data.py` (all 35, including the `_process_quotes_batch` and `_process_multiquotes_batch` helpers), and they do all echo the requested OpenAlgo exchange today, so the new lookup holds. Response shape is unchanged, and the new sort key is safe because every result path emits `exchange` (`calculate_greeks` success, deep ITM, the IV fallback, `_expired_option_greeks_response`, and the `setdefault` error path).

Two things must change before this merges, and one is worth deciding on.

1. `services/option_greeks_service.py`: the multiquotes exchange needs a fallback. `option_prices[(sym, exch)] = ltp` now takes `exch` straight from the broker result. If any result item arrives without `exchange`, every later `option_prices.get(batch_key)` misses and the whole batch returns "Option LTP not available", not just the colliding legs. That is not hypothetical, the proof is already in your PR: apply your patch, then revert only `test/test_multi_option_greeks_regression.py` to its pre-PR form and run it. `test_multi_option_greeks_converts_expired_legs_to_zero_greeks` fails with `Multi Greeks completed: 0/2 successful`, response `status: "error"`, `success is False`. You hit this yourself, which is why the mock had to be edited. Fall back to the requested exchange when the result omits it, so a future broker adapter degrades one leg instead of the entire request.

2. `.github/workflows/ci.yml`: the new test never runs. `test/test_multi_option_greeks_regression.py` is not in the `backend-test` step's file list (lines 45 to 52). Add it. That file already existed unregistered on main, so the gap is not yours, but your test is exactly the profile the list is for. I ran the existing 11 files plus yours together: 43 passed in 8.38s. Worth stating plainly that CI has not run on this branch at all, the workflows are waiting on maintainer approval, so the only validation so far is your local run and mine.

3. Worth deciding: `(symbol, exchange)` does not close the defect class. Two legs that share a symbol and an exchange but differ by `underlying_symbol` / `underlying_exchange` still collapse onto one `batch_key`. I reproduced this with your patch applied:

```
[{"symbol": "NIFTY30DEC2524000CE", "exchange": "NFO"},
 {"symbol": "NIFTY30DEC2524000CE", "exchange": "NFO",
  "underlying_symbol": "NIFTY30DEC25FUT", "underlying_exchange": "NFO"}]
```

Both legs come back priced off 24500.0, the second leg's underlying. The log shows both underlyings were fetched and the first result was then thrown away, because `symbol_to_spot_key[(sym, "NFO")]` is overwritten. That request shape is the endpoint's own documented example at `restx_api/multi_option_greeks.py:53`, so users will send it. The key that removes the whole class is the leg's index in `symbols`, which also makes `symbol_order` and the final `results.sort()` unnecessary because you can build results in request order directly. Issue #1819 asked for `(symbol, exchange)` specifically, so I am not treating this as a failure to follow the brief. Either extend this PR or say so and I will open a follow-up.

Two smaller notes:

- `test/test_multi_option_greeks_regression.py`: the spot-price assertions lean on a mock that contradicts production. You mock `get_underlying_exchange` to return NSE for NFO and BSE for BFO, but the real function at `services/option_greeks_service.py:294-295` returns "NSE" for any equity base symbol regardless of the options exchange, so `RELIANCE28MAR243000CE` on NFO and on BFO would share one spot key in production. The divergent-underlying case is reachable, just through an explicit per-leg `underlying_exchange` rather than through the exchange itself. The option-price assertions (41.5 against 63.25) are the part that pins the real regression, and those are sound.

- The PR description's safety argument cites the wrong lines. On main, `services/quotes_service.py:274` is a `BrokerData` constructor branch, `:290` is a closing brace, `:297` is a `logger.debug` string. The per-symbol fallback that sets `exchange` is `:303` and `:310`, and the broker-native branch at `:316-322` does `combined_results = results + multiquotes` with nothing added, passing the adapter's list straight through. So the guarantee you rely on lives in 35 broker adapters, not in `quotes_service.py`. It does hold in all 35 today, which is why point 1 is a hardening request rather than a bug report.

No migration is needed, nothing touches the schema, and the diff is tight: two files, one commit, no build artifacts. `uv run ruff check` is clean on both files. The 101 character line at 1105 is one over the configured limit of 100, but `ruff format --check` already fails on both files on unmodified main, so fix it or leave it as you prefer. The commit scope `fix(python)` matches the issue title, so no change needed there.

Pragitics:
Thanks for digging into this properly - the underlying_symbol case reproduced exactly as you described, both legs off 24500.0. Went with extending rather than the follow-up, since that shape is the documented example. Per-leg state keys by leg index now, and symbol_order plus results.sort() are gone as you suggested.

Fallback and ci.yml are in. One place I went slightly further than you suggested: a symbol requested on two exchanges maps to None rather than either one, because guessing there would quietly bring back the collision. Those legs go unpriced and the rest of the batch is fine.

Mock dropped, line wrapped, and I fixed the description - you were right that the guarantee lives in the adapters, not quotes_service. Six tests now, each one checked to fail with its own fix reverted. Also noted in the description that CI has not run on this branch.

marketcalls:
Thanks for the thorough fix and for addressing the review feedback comprehensively.

I independently reproduced #1819: with symbol-only state, the NFO leg received the BFO LTP (63.25 instead of 41.5). The updated implementation passes that red/green regression. Keying per-leg state by request index also correctly covers identical symbol/exchange legs with different explicit underlyings, while `(symbol, exchange)` remains the right identity for contract prices. The ambiguous missing-exchange path fails closed, request order and response shape are preserved, and duplicate quote requests are deduplicated.

Security review found no high-confidence vulnerabilities. The endpoint remains API-key authenticated, rate-limited, and bounded to 50 legs, and this change improves analytics data integrity without altering order execution or shared state.

I updated the branch with current `main` and resolved the CI registration conflict by retaining both `test_multi_option_greeks_regression.py` and `test_history_format.py`. Local verification passed 53 CI-safe backend tests plus focused Ruff checks. The hosted backend, frontend, E2E, lint, security, Snyk, and reviewer checks are now green. Approved for merge.

PR #1880, opened 2026-08-23
fix: expose active navigation route

siddharthg2309:
## Summary

Expose the visually active navigation route to assistive technology with `aria-current="page"`.

## Related issue

Closes #1833

## Changes

- Reuse the existing active-route calculation for accessibility state.
- Add `aria-current="page"` to the active desktop navigation link.
- Add the same state to active links in the navigation sheet.
- Cover both the sheet’s main navigation and Quick Access link paths.
- Leave inactive links without an `aria-current` attribute.

## Coverage

- Exactly one desktop link is marked current for a nested Strategy route.
- Inactive desktop links do not receive `aria-current`.
- Exactly one active main-navigation sheet link is marked current.
- Exactly one active Quick Access sheet link is marked current.
- Existing nested-route behavior remains aligned with the visual active state.

## Testing

- [x] `npm run test:run -- src/components/layout/Navbar.test.tsx`
  - 3 passed
- [x] `npm run lint`
  - passed
- [x] `git diff --check`
  - passed

## Checklist

- [x] This PR contains one focused accessibility fix.
- [x] Active desktop and sheet links expose `aria-current="page"`.
- [x] Inactive links do not expose `aria-current`.
- [x] Nested route behavior is covered.
- [x] No routing or visual active-state behavior changed.
- [x] No UI redesign is included; screenshots are not applicable.

---
## Summary by cubic
Expose the visually active navigation route to assistive technology using aria-current="page". Previously no link declared the current page; now exactly one active link in desktop navigation, the mobile navigation sheet (main navigation and Quick Access), and the mobile bottom navigation sets aria-current="page", improving screen reader context without changing routing or visuals. Fixes #1833.

- Reuses the existing active-route logic; inactive links omit aria-current.
- Adds tests to verify exactly one current link per context (desktop, mobile sheet, mobile bottom nav) and that nested routes match the visual active state.

<sup>Written for commit ad50bee1e105bccf4db72af8732503ae19d1ad23. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1880?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #1886, opened 2026-08-24
fix(frontend): sanitize error report URLs

### LINKED ISSUE #1851: security(frontend): Strip query strings and fragments from error-report URLs
## Summary

Client error reports can include the full current URL, including query strings and fragments.

## Evidence

- `frontend/src/utils/errorReporter.ts:132` falls back to `window.location.href`.
- `frontend/src/utils/errorReporter.ts:155` uses an event filename or the full current URL.

## Why this matters

OAuth callbacks, reset links, and other routes can carry secrets or one-time values in query parameters or fragments. Those values should not be copied into diagnostics.

## Scope

Create a small URL sanitizer that retains useful origin/path information while removing search and hash components. Handle relative URLs and non-URL asset filenames safely, then use it in every report path.

## Acceptance criteria

- [ ] Current-page reports include origin and pathname but no query or fragment.
- [ ] Absolute and relative script URLs are sanitized consistently.
- [ ] Plain filenames remain useful and do not throw.
- [ ] Unit tests include obvious fake OAuth/reset secrets and assert they are absent.

## Verification

```powershell
npm run test:run -- src/utils/errorReporter.test.ts
npm run lint
```

## Security guidance

- Difficulty: Easy
- Estimated time: 1-3 hours
- You will learn: URL parsing, data minimization, and privacy-focused unit tests.

siddharthg2309:
## Summary

Sanitize client error-report URLs before they are sent to the backend.

Reports now retain only the URL origin and pathname, preventing query-string and fragment values from being copied into diagnostics.

## Related issue

Closes #1851

## Changes

- Add a shared URL sanitizer for client error reporting.
- Remove query strings and fragments from current-page and script URLs.
- Handle absolute URLs, relative URLs, and plain asset filenames safely.
- Route every client report path through the sanitized reporting boundary.
- Add regression tests using only clearly fake OAuth/reset values.

## Testing

- [x] `npm run test:run -- src/utils/errorReporter.test.ts`
  - 3 passed
- [x] `npm run lint`
  - passed
- [x] `npm run test:run`
  - 48 test files, 967 tests passed
- [x] `git diff --check`
  - passed

## Checklist

- [x] This PR contains one focused security fix.
- [x] The change is covered by unit tests.
- [x] No real secrets, credentials, or sensitive values are included.
- [x] No generated frontend build artifacts are included.

---
## Summary by cubic
Sanitizes client error-report URLs and adds server-side enforcement to prevent leaking tokens or PII. Previously we logged the full URL; now the client redacts sensitive query values and strips fragments, and the server re-sanitizes before logging.

- Redacts sensitive query parameters (e.g., token, code, apiKey, jwt) and strips fragments while keeping non-sensitive params for context; server endpoint sanitizes again before logging.
- Normalizes absolute, relative, and bare filenames; keeps `*-extension` URLs actionable without query/fragment; collapses opaque schemes like `data:` and `blob:` to scheme only; safe fallback for malformed inputs.
- Reporter uses a shared `sanitizeErrorReportUrl` and `handleWindowError`; payload shape unchanged, but `url` may retain safe query context.
- Adds frontend and backend tests for OAuth/reset flows, asset URLs, browser extensions, and malformed URLs; no migrations or UI changes.

<sup>Written for commit d8f2cdbaefb3fedb5529a4cf6ab8571a425919d5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1886?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/blueprints/admin.py
+++ b/blueprints/admin.py
@@ -2,6 +2,7 @@ import json
 import os
 from datetime import date, datetime, timedelta
 from pathlib import Path
+from urllib.parse import parse_qsl, urlencode, urlsplit, urlunsplit
 
 from flask import Blueprint, flash, jsonify, redirect, render_template, request, url_for
 
@@ -855,6 +856,27 @@ _MAX_CLIENT_URL_LEN = 2000
 _MAX_CLIENT_COMPONENT_STACK_LEN = 5000
 _MAX_CLIENT_USER_AGENT_LEN = 500
 _CLIENT_LEVEL_ALLOWLIST = frozenset({"ERROR", "WARN"})
+_REDACTED_CLIENT_URL_VALUE = "[redacted]"
+_SENSITIVE_CLIENT_URL_QUERY_PARAMETER_NAMES = frozenset(
+    {
+        "token",
+        "code",
+        "requesttoken",
+        "accesstoken",
+        "authtoken",
+        "refreshtoken",
+        "resettoken",
+        "apikey",
+        "email",
+        "state",
+        "password",
+        "otp",
+        "secret",
+        "clientsecret",
+        "idtoken",
+        "jwt",
+    }
+)
 
 # Logger dedicated to browser-reported errors. Distinct name so they're easy
 # to filter in errors.jsonl and in the grouped view.
@@ -877,6 +899,32 @@ def _scrub_control_chars(text):
     return "".join(ch for ch in text if ch == "\n" or ch == "\t" or (ch.isprintable()))
 
 
+def _sanitize_client_error_url(url):
+    """Redact sensitive query values and remove fragments before logging a URL."""
+    try:
+        parsed = urlsplit(url)
+    except ValueError:
+        return url.split("?", 1)[0].split("#", 1)[0]
+
+    if parsed.scheme in {"", "http", "https"}:
+        query = urlencode(
+            [
+                (
+                    key,
+                    _REDACTED_CLIENT_URL_VALUE
+                    if key.lower().replace("_", "").replace("-", "")
+                    in _SENSITIVE_CLIENT_URL_QUERY_PARAMETER_NAMES
+                    else value,
+                )
+                for key, value in parse_qsl(parsed.query, keep_blank_values=True)
+            ]
+        )
+        return urlunsplit((parsed.scheme, parsed.netloc, parsed.path, query, ""))
+    if parsed.scheme.endswith("-extension"):
+        return urlunsplit((parsed.scheme, parsed.netloc, parsed.path, "", ""))
+    return f"{parsed.scheme}:"
+
+
 @admin_bp.route("/api/errors/client", methods=["POST"])
 @check_session_validity
 @limiter.limit(_CLIENT_REPORT_RATE)
@@ -897,7 +945,9 @@ def api_errors_client_report():
 
         message = _scrub_control_chars(str(data.get("message") or ""))[:_MAX_CLIENT_MESSAGE_LEN]
         stack = _scrub_control_chars(str(data.get("stack") or ""))[:_MAX_CLIENT_STACK_LEN]
-        url = _scrub_control_chars(str(data.get("url") or ""))[:_MAX_CLIENT_URL_LEN]
+        url = _sanitize_client_error_url(
+            _scrub_control_chars(str(data.get("url") or ""))
+        )[:_MAX_CLIENT_URL_LEN]
         component_stack = _scrub_control_chars(str(data.get("component_stack") or ""))[
             :_MAX_CLIENT_COMPONENT_STACK_LEN
         ]
--- a/frontend/src/utils/errorReporter.test.ts
+++ b/frontend/src/utils/errorReporter.test.ts
@@ -1,0 +1,53 @@
+import { waitFor } from '@testing-library/react'
+import { afterEach, describe, expect, it, vi } from 'vitest'
+
+import { reportClientError, sanitizeErrorReportUrl } from './errorReporter'
+
+afterEach(() => {
+  vi.unstubAllGlobals()
+  window.history.replaceState(null, '', '/')
+})
+
+describe('sanitizeErrorReportUrl', () => {
+  it('removes query strings and fragments from current-page reports', async () => {
+    const fakeToken = 'test-token-not-real'
+    window.history.replaceState(null, '', `/oauth/callback?code=${fakeToken}#complete`)
+    const fetchMock = vi
+      .fn<typeof fetch>()
+      .mockResolvedValueOnce({
+        ok: true,
+        json: async () => ({ csrf_token: 'csrf-token-not-real' }),
+      } as Response)
+      .mockResolvedValueOnce({ status: 200 } as Response)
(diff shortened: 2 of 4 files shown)


marketcalls:
Thanks for this. The diagnosis is right and I verified the part that matters: the url field of the client error report was in fact the only path by which a full query string could reach log/errors.jsonl. utils/traffic_logger.py:62 and utils/logging.py:320 both record request.path, which is query-free, so nothing else on the server was already capturing it. The tests are genuine regression tests. I applied the patch and ran vitest locally, 3 pass, and with only errorReporter.ts reverted all 3 fail. frontend/vitest.config.ts includes src/**/*.test.ts and the frontend-test job in .github/workflows/ci.yml runs the whole suite with no path filter, so the file will execute in CI and no ci.yml change is needed. Biome lint is clean on both files, scope is two files, no dist artifacts, Conventional Commit title, issue linked.

Four things to change or confirm before I merge.

1. frontend/src/utils/errorReporter.ts, non-http schemes. new URL() succeeds for schemes that have no real origin, and the result is wrong rather than merely cosmetic. Running your function body against the URL parser gives:

- chrome-extension://abcdef/content.js?token=X becomes null/content.js
- moz-extension://guid/inject.js becomes null/inject.js
- data:text/html;base64,PAYLOAD becomes nulltext/html;base64,PAYLOAD (the body is not stripped)
- blob:http://host/uuid becomes http://hosthttp://host/uuid

The extension cases are reachable: an extension content script that throws in the page delivers an ErrorEvent whose filename is that URL, and the extension id is the one field that makes the report actionable. Please guard on url.protocol and return something sane for anything that is not http: or https:, for example the protocol alone for opaque schemes and protocol + host + pathname for the extension schemes.

2. frontend/src/utils/errorReporter.ts, consider redacting rather than deleting. Stripping the whole query also removes the reproduction context from the routes that legitimately carry it: frontend/src/pages/Search.tsx lines 104 to 110 (symbol, exchange, underlying, expiry, instrumenttype, strike_min, strike_max), frontend/src/pages/HistorifyCharts.tsx lines 124 to 125 (exchange, interval), frontend/src/pages/StrategyBuilder.tsx line 155 (load). After this change a crash on /search?symbol=...&expiry=... reports only /search, which I cannot reproduce from. Across the whole app, frontend/src/pages/ResetPassword.tsx lines 58 to 61 is the only page that reads a secret-bearing param, so keeping the keys and blanking the values, or redacting a fixed list (token, code, request_token, access_token, api_key, email, state), closes exactly the same hole without that cost. I will take either, but say which you intend.

3. blueprints/admin.py around line 897, apply the same strip server-side. url is currently passed through _scrub_control_chars and written verbatim at line 916. CI force-commits frontend/dist/ to main and users upgrade with a plain git pull, so a tab still holding the previous index.html keeps posting the full URL after the server is upgraded. One line there makes the guarantee independent of which bundle the browser has, and covers any future caller of the endpoint. The server is the durable boundary here, the client change is the convenience.

4. frontend/src/utils/errorReporter.test.ts, two gaps. The path that motivated the fix, the window error listener at errorReporter.ts:155 that passes event.filename, is not covered, and neither is the explicit payload.url branch at line 132. Only the window.location.href default is. Please add both, plus one case that reaches the catch fallback (sanitizeErrorReportUrl('http://[') hits it). Note that this file is not type-checked by the build: frontend/tsconfig.json references only tsconfig.app.json and tsconfig.node.json, tsconfig.app.json excludes src/**/*.test.ts, and tsconfig.test.json is not in the references array, so tsc -b never visits it. Do not rely on frontend-build to catch type mistakes in the test.

One calibration point on the framing, because it changes how urgently this needs to ship. /admin/api/errors/client is behind @check_session_validity, and utils/session.py:57 requires session['logged_in']. The reset-password page is unauthenticated, so a report from it is answered 401 and never written to the log. The reachable case is narrower: a user who already has a live OpenAlgo session and opens the reset link in the same browser. The broker OAuth callbacks are not affected at all, since blueprints/brlogin.py:37 registers /<broker>/callback as a Flask route that returns redirects and blueprints/react_app.py registers no matching rule, so the React bundle and the reporter never run on a URL carrying request_token or code. The change is still worth making, it is hardening rather than an active leak, which is exactly why I would prefer the redaction variant in point 2 that keeps the debugging value.

No migration is needed, nothing here touches the schema. Note that CI has not run on this branch yet, only Snyk and cubic reported, so the green marks do not mean lint, build or tests passed. Everything above I ran locally against your patch.

siddharthg2309:
Thanks for the review. I addressed the requested changes: I used the redaction approach to preserve safe query context, remove fragments, and redact sensitive values.

The URL handling now covers extension and opaque schemes, the same protection is applied server-side before logging, and the requested explicit-URL, window-error, and malformed-URL tests are included.

marketcalls:
Merged as `e9b29821`, and #1851 is closed.

Sanitizing on both boundaries was the right call and it is why this went in ahead of several higher-scoring PRs. The client helper alone would be bypassed by a POST straight at `api_errors_client_report`, so the server-side pass in `blueprints/admin.py` is the half that actually guards the leak. The reachable case is real: `blueprints/auth.py` redirects to `/reset-password` with a one-time token in the query string, and `blueprints/brlogin.py` handles broker OAuth callbacks carrying `code` and `request_token`, and both of those URLs were being written verbatim into `log/errors.jsonl`.

## One thing I fixed on your behalf rather than sending back

`test/test_admin_client_error_url.py` was not registered in the CI-safe list, so as merged your three server-side tests existed and executed nowhere. I confirmed it against the computed merge tree before merging: main's list came through intact at 16 entries with yours absent.

I pushed the one-line registration to main directly as `07a5f649`, so the list is now 17 and your tests run on Python 3.12, 3.13 and 3.14 from the next PR onward. Nothing for you to do.

Verified by running the amended list exactly as `backend-test` invokes it: 84 passed. The single failure in that run, `test_python_editor.py` with `PermissionError: [WinError 32]`, reproduces on main with my change stashed. It is a pre-existing Windows-only file-locking issue in a test that was already on the list, and it is green on the ubuntu runners.

## For next time

This is the one habit worth building, and it is the same note I left on #1880: **if a PR adds `test/*.py`, the `ci.yml` line is part of that PR.** The `backend-test` step enumerates files one by one, there is no `pytest test/` run, so an unlisted backend test is invisible. Frontend tests under `frontend/src/**/*.test.tsx` need nothing, which is why the `errorReporter.test.ts` half of this PR was already covered.

Second, smaller: the PR body still described removing query strings and fragments, which stopped being an accurate description of the code after your own revision to key-targeted redaction. Squash merge writes the body into the permanent commit record, so I rewrote it at merge time. Worth rereading the description against the head before saying a PR is ready.

Good fix. The two-boundary instinct is the part to keep.

PR #1887, opened 2026-08-24
test: convert history format script into pytest coverage

### LINKED ISSUE #1827: test: Convert test_history_format.py into deterministic pytest coverage
## Summary

`test/test_history_format.py` is a manual script, so pytest collects zero tests from it.

## Evidence

- `uv run pytest test/test_history_format.py --collect-only` reports zero collected tests.
- The file contains request and output checks without pytest fixtures.

## Scope

Refactor it into hermetic pytest tests with mocked history responses. Cover the intended shape, `oi=0` fallback, and malformed upstream data.

## Acceptance criteria

- [ ] Pytest collects and runs the file.
- [ ] No network or credentials are used.
- [ ] Response shape and `oi=0` are asserted.
- [ ] Malformed data has an explicit expected result.

## Verification

```powershell
uv run pytest test/test_history_format.py -v
uv run pytest test/test_history_format.py --collect-only
uv run ruff check test/test_history_format.py
```

- Difficulty: Easy
- Estimated time: 1-3 hours
- You will learn: pytest discovery, fixtures, and mocking.

K-PRAGALATHAN:
## Description

`test/test_history_format.py` was a manual reconnaissance script rather than a test. Everything ran at import time: it built an `openalgo` client against a live server on `127.0.0.1:5000` using `OPENALGO_API_KEY`, derived its date range from `datetime.now()`, and printed whatever came back. Nothing was asserted.

Because the filename matches pytest's `python_files = ["test_*.py"]` pattern, pytest imported the module, paid the cost of the network call, and then collected zero tests.

This replaces it with hermetic tests around `get_history_with_auth` (`services/history_service.py:76`). A stub broker module stands in for `broker.<name>.api.data`, and `get_token` is patched, so no server, network, or credentials are involved.

Coverage (7 cases from 4 test functions):

| Test | Asserts |
| --- | --- |
| `test_history_is_returned_as_success_records` | `(True, {"status": "success", "data": [...]}, 200)` and the record keys |
| `test_missing_oi_column_defaults_to_zero` | the `oi = 0` fallback at `history_service.py:134-136` |
| `test_broker_supplied_oi_is_preserved` | the fallback does not overwrite broker-supplied `oi` |
| `test_non_dataframe_from_broker_is_reported_as_an_error` | parametrized over `None`, dict, list and str: `(False, error, 500)` with the documented message |

No production code is changed.

`.github/workflows/ci.yml` gains one line so `backend-test` actually runs the new file: its allowlist is explicit, so without the entry the tests would never execute upstream. Added against the list as it stands today, per the review note on ordering with #1888 and #1874.

## Related Issues

Closes #1827

## Testing

```
uv run pytest test/test_history_format.py -v
7 passed in 0.76s

uv run pytest test/test_history_format.py --collect-only
7 tests collected in 0.68s      (was: 0 items, exit code 5, 8.50s)

uv run ruff check test/test_history_format.py
All checks passed!

uv run ruff format --check test/test_history_format.py
1 file already formatted
```

The full CI-safe allowlist, as `backend-test` now runs it: **47 passed in 13.09s**.

The `--collect-only` time drops from 8.50s to 0.68s because the import-time HTTP call is gone.

To confirm the assertions are not vacuous, the two behaviours under test were deliberately broken in `services/history_service.py` and the suite was re-run:

- disabling `df["oi"] = 0` at `services/history_service.py:135-136` fails 2 tests: `test_missing_oi_column_defaults_to_zero` with `KeyError: 'oi'` (test/test_history_format.py:75), and `test_history_is_returned_as_success_records` with an `AssertionError` on the whole-dict compare (line 68)
- changing the `ValueError` message fails all 4 malformed-input cases

Both mutations were reverted. The PR touches two files only: `test/test_history_format.py` and the single added line in `.github/workflows/ci.yml`. No application code is modified.

## Checklist

- [x] One fix per pull request
- [x] Follows Conventional Commits (`test:`)
- [x] Tests pass locally with `uv run pytest`
- [x] `ruff check` and `ruff format` are clean
- [x] No network access or credentials required by the tests
- [x] No production code modified
- [x] New test file registered in the `ci.yml` CI-safe allowlist
- [ ] Screenshots - not applicable, no UI change

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,6 +50,7 @@ jobs:
             test/test_auth_resume.py test/test_auth_upsert_multisession.py \
             test/sandbox/test_execution_backlog.py test/test_event_bus_bounded.py \
             test/test_telegram_api_contract.py \
+            test/test_history_format.py \
             -v --timeout=60
 
   # Frontend linting
--- a/test/test_history_format.py
+++ b/test/test_history_format.py
@@ -1,46 +1,94 @@
-import os
-import sys
+"""The history service always returns records shaped the same way, whatever the broker sends.
 
-# Add parent directory to path for imports
-sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
-
-from datetime import datetime, timedelta
+Every broker's BrokerData.get_history returns its own DataFrame, so
+get_history_with_auth is the single place that normalises the result: it rejects
+anything that is not a DataFrame, stamps in an 'oi' column when the broker omits
+one, and serialises to a list of records. These tests pin that contract down with
+a stub broker module, so no server, network or credentials are involved.
+"""
 
 import pandas as pd
-from openalgo import api
+import pytest
 
-# Test the history API to see the response format
-client = api(
-    api_key=os.getenv("OPENALGO_API_KEY"),
-    host="http://127.0.0.1:5000",
-)
+from services import history_service
 
-# Get history data
-end_date = datetime.now()
-start_date = end_date - timedelta(days=5)
+CANDLE = {
+    "timestamp": 1_700_000_000,
+    "open": 100.0,
+    "high": 101.5,
+    "low": 99.5,
+    "close": 101.0,
+    "volume": 1_000,
+}
 
-response = client.history(
-    symbol="RELIANCE",
-    exchange="NSE",
-    interval="5m",
-    start_date=start_date.strftime("%Y-%m-%d"),
-    end_date=end_date.strftime("%Y-%m-%d"),
-)
 
-print("History API Response Type:", type(response))
+def stub_broker(df):
+    """Build a stand-in for a broker's api.data module whose get_history returns df."""
 
-if isinstance(response, pd.DataFrame):
-    print("\nResponse is a DataFrame!")
-    print("Shape:", response.shape)
-    print("Columns:", list(response.columns))
-    print("\nFirst 5 rows:")
-    print(response.head())
-elif isinstance(response, dict):
-    print("\nResponse is a dictionary with keys:", response.keys())
-    if "data" in response:
-        data = response["data"]
-        print("Data type:", type(data))
-        if isinstance(data, pd.DataFrame):
-            print("Data is a DataFrame with columns:", list(data.columns))
-else:
-    print("Unexpected response type:", type(response))
+    class BrokerData:
+        def __init__(self, auth_token, feed_token=None):
+            self.auth_token = auth_token
+            self.feed_token = feed_token
+
+        def get_history(self, symbol, exchange, interval, start_date, end_date):
+            return df
+
+    return type("StubBrokerModule", (), {"BrokerData": BrokerData})
+
+
+@pytest.fixture
+def broker(monkeypatch):
+    """Install a stub broker module and a resolvable symbol; return a caller for the service."""
+
+    def call(df):
+        monkeypatch.setattr(history_service, "get_token", lambda _symbol, _exchange: "12345")
+        monkeypatch.setattr(
+            history_service, "import_broker_module", lambda _broker: stub_broker(df)
+        )
+        return history_service.get_history_with_auth(
+            auth_token="token",
+            feed_token="feed",
+            broker="stub",
+            symbol="RELIANCE",
+            exchange="NSE",
+            interval="5m",
+            start_date="2026-01-01",
+            end_date="2026-01-05",
+        )
+
+    return call
+
+
+def test_history_is_returned_as_success_records(broker):
+    success, response, status = broker(pd.DataFrame([CANDLE]))
+
+    assert success is True
+    assert status == 200
+    assert response["status"] == "success"
+    assert response["data"] == [{**CANDLE, "oi": 0}]
+
+
+def test_missing_oi_column_defaults_to_zero(broker):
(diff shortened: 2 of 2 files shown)


marketcalls:
Thanks, this is a clear improvement and I am merging it after one small change: register the file in the CI allowlist.

## What I verified

CI has not actually run on this branch. Fork PRs here wait on maintainer approval, so the only green checks on this PR are the two Snyk "no manifest changes detected" runs. No Ruff, no pytest, no frontend job has validated the branch. I ran the checks locally instead, against current `main`:

- The problem is real. With `main`'s version of the file in place, `uv run pytest test/test_history_format.py --collect-only` reports `collected 0 items ... no tests collected in 2.50s`, and the module builds an `openalgo` client against `http://127.0.0.1:5000` at import time.
- Your patch applies cleanly (`git apply --check` passes). After applying it: `uv run pytest test/test_history_format.py -v` gives 7 passed in 0.60s, `ruff check` is clean, `ruff format --check` reports the file already formatted, and the file is pure ASCII.
- The monkeypatching matches how the service resolves names. `validate_symbol_exchange` calls `get_token` (services/history_service.py:48) and `get_history_with_auth` calls `import_broker_module` (line 111) through module globals, so patching `history_service.get_token` and `history_service.import_broker_module` reaches both call sites.
- The assertions are not vacuous. I reproduced your mutation check by disabling `df["oi"] = 0` at services/history_service.py:135-136. Two tests fail, but not the way the PR body says: `test_missing_oi_column_defaults_to_zero` fails with `KeyError: 'oi'` (test/test_history_format.py:75), while `test_history_is_returned_as_success_records` fails with a plain `AssertionError` on the whole-dict compare (line 68). Worth correcting in the description; the coverage itself is sound.
- No production code is touched, which the single-file diff confirms.

## Required before merge

1. `.github/workflows/ci.yml`, the "Run CI-safe tests" list at lines 47-53: add `test/test_history_format.py`. `backend-test` runs an explicit allowlist of exactly eleven files and yours is not one of them, so as the PR stands the seven new tests still never execute upstream and cannot guard `get_history_with_auth` against a regression. That is the same hole commit 140996e1d closed for the auth session tests, and version 2.0.1.9 rewrote two allowlisted live-server tests that asserted nothing for the same reason. Nothing blocks the line: the file is hermetic, runs in under a second, and test/conftest.py pins the sqlite paths under `db/`, which exists on a fresh checkout because `db/readme.txt` is tracked.

## Optional, not blocking

2. test/test_history_format.py: two cheap branches stay uncovered even though your fixture already owns both patch points. `get_token` returning `None` reaches the 400 at services/history_service.py:107-109, and `import_broker_module` returning `None` reaches the 404 at line 113. As written, a regression that turned that 404 into a 500, or dropped symbol validation entirely, would still pass all seven tests.
3. `stub_broker` (test/test_history_format.py:25-36) always builds a `BrokerData.__init__` with three parameters (line 29), so `co_argcount` is 3 and only the `param_count > 2` path at services/history_service.py:121 is exercised. Brokers whose `BrokerData` takes `auth_token` alone go through the `else` at line 123, and the `hasattr(..., "__code__")` fallback at line 126 is never entered either. A second stub with a one-arg `__init__` covers that dispatch, which is precisely the kind of thing that breaks quietly per broker.
4. test/test_history_format.py:44-47: both `monkeypatch.setattr` calls sit inside the returned closure, so the fixture body itself does no setup. Behaviourally identical here since each test calls the closure once, so this is readability only.

## Ordering against the sibling PRs

#1874, #1887 and #1888 all cover the history and intervals data API, and all three need a line in the same seven-line list in ci.yml. This PR goes in first, so add your ci.yml line against the list as it stands today. I will take #1888 next and #1874 last so their entries stack on top of yours instead of conflicting. CONTRIBUTING.md carries no copy of that list at the moment, so there is nothing else to keep in sync; if #1889 lands one before this merges, mirror the same filename there.

K-PRAGALATHAN:
Thanks for the detailed review, and for running the checks locally when CI could not.

## Required change: done

`c7567e9b3` adds `test/test_history_format.py` to the "Run CI-safe tests" list in `.github/workflows/ci.yml`. Added against the list exactly as it stands today, as a new line rather than an edit to an existing one, so the #1888 and #1874 entries stack on top without conflicting.

Verified by running the full allowlist as `backend-test` invokes it:

```
uv run pytest test/test_log_location.py ... test/test_history_format.py -v --timeout=60
47 passed in 13.09s
```

The six warnings in that run are pre-existing `PytestReturnNotNoneWarning`s from `test/test_rate_limits_simple.py`, not from this file.

## Correction: you are right about the mutation output

My description was wrong. Disabling `df["oi"] = 0` fails `test_missing_oi_column_defaults_to_zero` with `KeyError: 'oi'` and `test_history_is_returned_as_success_records` with an `AssertionError` on the whole-dict compare, not both with `KeyError`. I read only the tail of that run and generalised from the single traceback shown. The PR description now states both failure modes with line numbers.

## Optional items

I have left 2, 3 and 4 out of this PR to keep it to the one required change, given the ordering you described.

Points 2 and 3 are worth doing and I would like to. The uncovered 400/404 branches and the single-argument `BrokerData` dispatch at `services/history_service.py:121-126` are real gaps, and the one-argument case is exactly the sort of thing that breaks per broker without anyone noticing.

Happy either way: say the word and I will add them here, or I will open a follow-up PR once this and the sibling PRs have landed so the ci.yml ordering stays clean. Your call on which is less disruptive.

Issue #1841, opened 2026-08-22
docs(contributing): Make test instructions match CI-safe commands


marketcalls:
## Summary

The contributor guide recommends commands that differ from CI and uses the interactive frontend test watcher where a one-shot verification command is expected.

## Evidence

- `CONTRIBUTING.md:379-402` recommends the full backend suite and `npm test`.
- `CONTRIBUTING.md:557` repeats the watcher command.
- `.github/workflows/ci.yml:47-53` runs the maintained backend selection, and `:106-107` uses `npm run test:run` and coverage.
- `docs/devsprint/README.md:308-342` already shows CI-safe examples.

## Scope

Document a fast local check, a one-shot frontend check, and the fuller CI-equivalent checks. Explain that `npm test` is watch mode rather than using it in a completion checklist.

## Acceptance criteria

- [ ] Completion checklists use `npm run test:run`.
- [ ] Backend commands are valid for the current test layout.
- [ ] Fast versus full checks are clearly distinguished.
- [ ] Commands agree with package scripts and CI.

## Verification

```powershell
rg -n "npm test|npm run test:run|uv run pytest" CONTRIBUTING.md .github/workflows/ci.yml frontend/package.json
git diff --check
```

## First-time contributor notes

- Difficulty: Easy
- Estimated time: 1-2 hours
- Broker credentials required: No
- You will learn: mapping contributor commands to CI and package scripts.
- Please comment on this issue before starting so contributors can avoid duplicate work.

nightcityblade:
Hi, I'd like to work on this. I'll submit a PR shortly.

nightcityblade:
I’m stepping back from this issue after reviewing the project’s sequential-PR policy, so I don’t hold it while another contribution is under review. No PR was opened for this issue.

suhaslord:
Hi, I can take this. I’ve already compared the contributor commands against the current CI workflow and I’ll keep the change focused to the test instructions/checklist in CONTRIBUTING.md.

PR #1883, opened 2026-08-23
docs(contributing): complete table of contents

nightcityblade:
## Description

Completes the `CONTRIBUTING.md` table of contents so every level-two heading is linked exactly once and appears in document order. The new entries use GitHub-compatible anchors, including punctuation-sensitive headings.

## Related issue

Fixes #1847

## Testing

- [x] `rg -n '^## ' CONTRIBUTING.md`
- [x] `git diff --check`
- [x] One-to-one heading/TOC order comparison
- [x] `markdownlint-cli2` MD051 link-fragment validation (0 issues)

## Checklist

- [x] This PR contains one focused documentation fix.
- [x] No screenshots are needed for this Markdown-only change.

---
## Summary by cubic
Completes the CONTRIBUTING.md table of contents so every level-two heading appears once in document order with GitHub-compatible anchors, including punctuation-sensitive headings. Previously some H2s were missing; now all are linked. Fixes #1847.

- Verify links for punctuation-heavy headings resolve (e.g., "Let's democratize algorithmic trading, together!" → #lets-democratize-algorithmic-trading-together; "Recognition & Community" → #recognition--community).
- No content changes; only the table of contents list was updated.

<sup>Written for commit 3090ee4ca9782023f2e6ca2a5f1d252ff64f5ea7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1883?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #1863, opened 2026-08-22
docs(contributing): correct frontend build-artifact and Node guidance

thaildhe172591:
CONTRIBUTING said `frontend/dist` is gitignored and not tracked, and listed Node.js as a general prerequisite. Both contradict the repository: dist is tracked on `main` and force-added by the `commit-dist` CI job, and a fresh clone of `main` runs the UI with no Node.js installed.

- Node.js is called out as frontend-only, with the exact engine floors from frontend/package.json (>=20.20.0 || >=22.22.0 || >=24.13.0)
- The frontend build is no longer part of the required first-run setup
- The build-assets section now explains that the .gitignore entry and the tracked copy on main are not in conflict, and who builds and commits dist
- Frontend items in the PR checklist are scoped to frontend changes
- Stale "(gitignored)" note in the project structure tree corrected

Closes #1842

---
## Summary by cubic
Corrects CONTRIBUTING to match the repo: `frontend/dist` is tracked on `main` and built/committed by CI, and Node.js is required only for frontend work. This removes misleading setup steps that made backend/doc contributors install Node and build the UI.

- States exact Node engines from `frontend/package.json` (`>=20.20.0 || >=22.22.0 || >=24.13.0`).
- Removes frontend build from first-run setup; explains why a fresh `main` works without Node.
- Clarifies `.gitignore` vs tracked `frontend/dist` and documents the `commit-dist` job in `.github/workflows/ci.yml`.
- Scopes PR checklist items to frontend changes and fixes the stale "(gitignored)" note in the project tree.

**Contributor actions**
- Do not commit `frontend/dist/`; CI owns it.
- Install Node only if you change frontend code and meet the engines above.
- For a working UI without Node, pull `main`; build locally only for frontend development.

<sup>Written for commit 0bbd7ed1132441b8ca06b42725829d785208ae00. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1863?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #1889, opened 2026-08-24
docs: align contributor test commands with CI

suhaslord:
Aligns the contributor test instructions with the maintained CI-safe backend selection and one-shot frontend test command. Keeps the change docs-only.\n\nRefs #1841.

---
## Summary by cubic
Aligns CONTRIBUTING test commands with CI to reduce local/CI mismatches and flakiness. Replaces the full backend suite and `npm test` with the CI-maintained backend selection (`--timeout=60` via `uv`/`pytest`) and a one‑shot `npm run test:run`; preserves guide formatting and addresses #1841.

- Backend: direct contributors to the CI-safe fast check; keep `uv run pytest test/ -v` as optional and note it’s not CI-safe and may be red on a clean checkout.
- Frontend: use `npm run test:run` and update the checklist to reference the CI-safe backend selection and one-shot frontend run.
- Docs-only; review is limited to verifying docs accuracy and CI parity.

<sup>Written for commit 96fc97265c0e1b2af1f3f6b3e4b873fbf2589e0d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1889?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -384,7 +384,7 @@ uv run pytest test/test_log_location.py test/test_navigation_update.py \
   test/sandbox/test_execution_backlog.py test/test_event_bus_bounded.py \
   test/test_telegram_api_contract.py -v --timeout=60
 
-# Full backend suite (slower)
+# Full backend suite (not CI-safe; not green on a clean checkout)
 uv run pytest test/ -v
 
 # Run React tests once (CI/completion check)
@@ -406,7 +406,7 @@ npm run e2e
 - [ ] Application starts without errors (`uv run app.py`)
 - [ ] All existing features still work
 - [ ] New feature works as expected
-- [ ] Python tests pass (`uv run pytest test/ -v`)
+- [ ] CI-safe backend tests pass (the pytest selection above)
 - [ ] Frontend tests pass (`cd frontend && npm run test:run`)
 - [ ] No TypeScript errors (`cd frontend && npm run build`)
 - [ ] No linting errors (Ruff for Python, Biome for frontend)
@@ -427,7 +427,6 @@ git push origin feature/your-feature-name
 ```
 
 ### 7. Create a Pull Request
-
 1. Go to your fork on GitHub
 2. Click **"Compare & pull request"**
 3. Fill out the PR template:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -427,6 +427,7 @@ git push origin feature/your-feature-name
 ```
 
 ### 7. Create a Pull Request
+
 1. Go to your fork on GitHub
 2. Click **"Compare & pull request"**
 3. Fill out the PR template:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -376,12 +376,20 @@ git commit -m "refactor: optimize order processing pipeline"
 ### 5. Test Your Changes
 
 ```bash
-# Run Python tests
+# Fast backend check (same maintained selection used by CI)
+uv run pytest test/test_log_location.py test/test_navigation_update.py \
+  test/test_python_editor.py test/test_rate_limits_simple.py \
+  test/test_logout_csrf.py test/test_auth_logout.py \
+  test/test_auth_resume.py test/test_auth_upsert_multisession.py \
+  test/sandbox/test_execution_backlog.py test/test_event_bus_bounded.py \
+  test/test_telegram_api_contract.py -v --timeout=60
+
+# Full backend suite (slower)
 uv run pytest test/ -v
 
-# Run React tests
+# Run React tests once (CI/completion check)
 cd frontend
-npm test
+npm run test:run
 
 # Run end-to-end tests
 npm run e2e
@@ -399,7 +407,7 @@ npm run e2e
 - [ ] All existing features still work
 - [ ] New feature works as expected
 - [ ] Python tests pass (`uv run pytest test/ -v`)
-- [ ] Frontend tests pass (`cd frontend && npm test`)
+- [ ] Frontend tests pass (`cd frontend && npm run test:run`)
 - [ ] No TypeScript errors (`cd frontend && npm run build`)
 - [ ] No linting errors (Ruff for Python, Biome for frontend)
 - [ ] API endpoints return correct responses


marketcalls:
Merging after two one-line fixes. The direction is right and I want this in, but the "Full backend suite" comment and the Python checklist item directly below it will both send contributors down a dead end as written.

## What I verified

CI has not run on this branch (fork PRs wait on maintainer approval here), so none of this was machine-checked upstream. I ran it locally against current `main`:

- The 11-file selection matches the `backend-test` step in `.github/workflows/ci.yml`: paths on lines 47-52, flags on line 53. All 11 files exist, including `test/sandbox/test_execution_backlog.py`.
- I ran your fast command verbatim: `40 passed in 5.67s`. It works.
- `frontend/package.json` confirms the swap: `"test": "vitest"` is watch mode, `"test:run": "vitest run"` is one-shot, and `.github/workflows/ci.yml:106` runs `npm run test:run`. Note the frontend distinction that makes this worth doing: the `frontend-test` job runs the whole vitest suite, so every `frontend/src/**/*.test.tsx` is gated by CI, whereas a backend test only runs in CI if it is named explicitly in the `backend-test` step.
- Patch applies cleanly to `CONTRIBUTING.md` at `c4d9371`, every added line is pure ASCII, and `docs:` is a valid Conventional Commit prefix. Docs-only, so no migration and no runtime invariant is in play.

## Required before merge

1. **`CONTRIBUTING.md:387`, `# Full backend suite (slower)`.** "Slower" is the wrong qualifier, and I checked why. On a clean `uv sync` checkout of `main`, `uv run pytest test/ -v` does not run a single test: collection is interrupted by two import errors and pytest exits non-zero. `test/test_bot_web.py:9` imports `get_telegram_bot` from `services.telegram_bot_service`, a name that does not exist in that module. `test/test_telegram_startup.py:16` imports `eventlet`, which lives only in `requirements-nginx.txt:149` and is in neither `pyproject.toml` nor `uv.lock`, so `uv sync` never installs it. Beyond that, a large share of what does collect needs a running instance, a connected broker and open markets (`test/test_greeks_live.py:5-14` spells out exactly those prerequisites). Please state the real distinction, for example:

   ```
   # Full backend suite: not CI-safe, and not green on a clean checkout.
   # Many of these tests need a running OpenAlgo instance, a connected broker
   # and open markets, and two modules currently fail to import at collection.
   uv run pytest test/ -v
   ```

   `docs/devsprint/README.md:324` already sets the house tone for this ("Do not run the whole `test/` directory and panic"). Match it.

2. **`CONTRIBUTING.md:409`, `- [ ] Python tests pass (uv run pytest test/ -v)`.** The PR demotes that command in the block above but leaves the checklist pointing at it, so the gate contributors actually tick is a command that cannot pass, and the fast selection you just introduced never appears in the checklist at all. Please do not paste the file list a second time; a reference is enough:

   ```
   - [ ] CI-safe backend tests pass (the pytest selection above)
   ```

## Optional, not blocking

3. Single source of truth for the selection. The list is now in three places, not two: `.github/workflows/ci.yml:47-52`, your `CONTRIBUTING.md:380-385`, and `docs/devsprint/README.md:317-322`. The third copy is already stale, listing 10 files and omitting `test/test_telegram_api_contract.py`, which is the drift this PR set out to stop. There is no `markers` key under `[tool.pytest.ini_options]` (`pyproject.toml:201-212`) and no wrapper script, so nothing enforces sync. A registered `ci_safe` marker, or a small script both `ci.yml` and the docs invoke, would fix it properly. Separate PR is fine.

4. `-v --timeout=60` on line 385 is redundant, since `pyproject.toml:208` sets `addopts = "-v --timeout=60"` for every invocation. Mirroring CI verbatim is a defensible reason to keep it. Leave it as is.

Also worth knowing: this block now pins the current 11-file list into the docs, so any PR that adds a file to the `backend-test` step has to update `CONTRIBUTING.md` in the same pass or the doc is stale on merge.

## Merge order

This collides with **#1863**. Both rewrite the same line, pre-patch `CONTRIBUTING.md:402`, `- [ ] Frontend tests pass (cd frontend && npm test)`. You change it to `npm run test:run`; #1863 changes it to `- [ ] Frontend tests pass, if you changed frontend code (cd frontend && npm test)`, keeping watch mode. I applied #1867, #1883, #1889, #1863 in that order to a scratch copy: the first three apply clean, #1863 fails with `patch failed: CONTRIBUTING.md:399`.

So: **#1889 merges before #1863**, and #1863 rebases afterwards and resolves that line by hand to the union, `- [ ] Frontend tests pass, if you changed frontend code (cd frontend && npm run test:run)`. Nothing to do on your side beyond the two fixes above. #1867 and #1883 do not touch your lines (#1867's nearest hunk starts at line 388, yours ends at 387), so no rebase is needed against either.

Push the two edits and I will merge.

suhaslord:
Pushed both requested fixes and verified the PR diff is clean. Thanks
for the detailed review!

marketcalls:
Merge after one two-line fix. Both round-1 items are done, and nothing regressed.

## Verified at head 96fc9726

- `CONTRIBUTING.md:387` now reads `# Full backend suite (not CI-safe; not green on a clean checkout)`. The misleading "(slower)" is gone, and the new wording is still accurate against current `main`: `test/test_bot_web.py:9` imports `get_telegram_bot`, a name that does not exist in `services/telegram_bot_service.py`, and `test/test_telegram_startup.py:16` imports `eventlet`, which is in neither `pyproject.toml` nor `uv.lock`.
- `CONTRIBUTING.md:409` now reads `- [ ] CI-safe backend tests pass (the pytest selection above)`. Exactly right, and no second copy of the file list.
- No regression. Three single-parent commits, all yours, `CONTRIBUTING.md` only. Zero CR bytes, so LF is intact and there is no line-ending flip. Trailing newline present. `git merge-file` against current `main` returns 0 conflicts and GitHub reports `MERGEABLE / CLEAN`, so #1883's expanded table of contents and #1867's `/dashboard` route text survive the merge untouched. Your branch is 6 behind, but it reverts none of it.
- Good self-catch on the third commit: `e67b0875` accidentally dropped the blank line after `### 7. Create a Pull Request` and stripped the newline at end of file, and `96fc9726` restored both. Base-to-head is now exactly your two intended hunks and nothing else.

## Required before merge

`CONTRIBUTING.md:379` says "same maintained selection used by CI". That is no longer true, through no fault of yours. `main` moved after your merge base `a1a2dc92`, and the `backend-test` step now runs 13 files while your block lists 11. Missing:

```
test/test_multi_option_greeks_regression.py   (added by #1885, 0f99ab6c)
test/test_history_format.py                   (added by #1887, 87568c55)
```

I ran both against `main`: your 11-file command gives `40 passed`, the current CI list gives `49 passed in 5.57s`. A contributor who ticks your new checklist can still go red in CI on those 9 tests, which is the exact failure this PR exists to prevent. Change the tail of the block to:

```
  test/test_telegram_api_contract.py \
  test/test_multi_option_greeks_regression.py \
  test/test_history_format.py \
  -v --timeout=60
```

That is the whole fix.

## Not blocking

1. Consider naming the real hazard on line 387 rather than only "not CI-safe": `test/test_all_order_types_execution.py:22` hardcodes an API key and its `test_all_order_types()` posts orders to `http://127.0.0.1:5000`, which is precisely the state this guide tells contributors to be in. One clause such as "some tests place real orders against a running instance" would close it.
2. The single-source-of-truth point is now proven rather than hypothetical: the list drifted within a day of you writing it. A registered `ci_safe` marker in `pyproject.toml`, or a script both `ci.yml` and the docs call, ends it for good. Separate PR, as agreed. `docs/devsprint/README.md:317` is a third copy and is still stale.
3. Merge order against #1863 is unchanged: this lands first, then #1863 rebases and takes the union on the frontend checklist line.

Push that one edit and I will merge.

PR #1892, opened 2026-08-24
fix(options): validate strike interval and option type at the resolver boundary

### LINKED ISSUE #1829: python: Validate legacy option-symbol resolver inputs
## Summary

The legacy option-symbol resolver assumes a safe strike divisor and maps every unrecognized option type to the put path.

## Evidence

- `services/option_symbol_service.py:200` divides by strike without guarding the divisor.
- `services/option_symbol_service.py:249-276` and `:543-565` choose the put path for unexpected types.

## Scope

Validate strike and option type at the resolver boundary while preserving valid formatting.

## Acceptance criteria

- [ ] Zero, invalid, and non-finite strikes return a controlled error.
- [ ] Unsupported option types are rejected.
- [ ] Valid calls and puts remain unchanged.
- [ ] Both legacy paths have tests.

## Verification

```powershell
uv run pytest test/test_option_symbol_service_validation.py -v
uv run ruff check services/option_symbol_service.py test/test_option_symbol_service_validation.py
```

- Difficulty: Medium
- Estimated time: 2-4 hours
- You will learn: legacy contracts and defensive numeric validation.

Narasimha722:
## Summary

Closes #1829.

The legacy option-symbol resolver trusted two inputs it should not have.

**1. The strike interval was used as a divisor with no guard.**
`get_atm_strike()` computes `round(ltp / strike_int) * strike_int`. A `strike_int` of `0` raised `ZeroDivisionError`, and a negative, NaN or infinite interval produced a strike that formats into a symbol no exchange lists. Both reached the client as an HTTP 500 with a traceback instead of a client error.

**2. An unrecognised option type was silently priced as a put.**
Every offset calculation branches on `CE` and takes the `else` branch for everything else, in both legacy paths (`calculate_offset_strike()` and `calculate_offset_strike_from_actual()`). An option type of `"CALL"`, `"C"` or `None` therefore resolved to the *put* strike. The symbol built around it carries the same bogus suffix, so `find_option_in_database()` returns 404 and the request fails closed - no wrong leg is placed. The harm is the diagnosis: the caller is told the symbol was "not found in NFO" when the real fault was the option type, and the resolver spends a broker quote call before saying so.

The same shape of failure existed for a non-finite LTP in `find_atm_strike_from_actual()`: `NaN` compares false against every strike, so `min()` returned the *lowest* strike in the chain and reported it as ATM.

Behaviour on `main`, before this change:

```
CE ITM2          -> 23500
PE ITM2          -> 23700
CALL ITM2        -> 23700      <- the put strike, no error
actual CALL ITM2 -> 23700.0    <- same in the actual-strikes path
strike_int=0     -> ZeroDivisionError: division by zero
NaN LTP ATM      -> 23400.0    <- the lowest strike in the chain, reported as ATM
```

## Why the fix belongs in the service, not the schema

`OptionSymbolSchema` validates `strike_int` and `option_type` for `POST /api/v1/optionsymbol`, and the options-order endpoints validate their own legs (`OptionsOrderSchema`, `OptionsMultiOrderLegSchema`, both `OneOf(["CE","PE","ce","pe"])` with `strike_int` on `Range(min=1)`).

The in-process bypass is `services/flow_openalgo_client.py:397`, which calls `get_option_symbol()` directly. Its caller, `services/flow_executor_service.py:2086`, reads `optionType` through `get_str()` with no enum check - unlike the options-order nodes in the same file, which check against `frozenset({"CE", "PE"})`. A Flow graph can therefore reach the resolver with any string at all.

The other direct callers were checked and are **not** bypasses: `mcp/mcpserver.py` goes over HTTP through the `openalgo` SDK client, and `services/synthetic_future_service.py` hardcodes `strike_int=None` and `option_type="CE"/"PE"`.

Validating at the resolver boundary covers the Flow path and every future caller in one place, instead of adding a fourth schema.

Range validation alone is also not enough at the boundary. `strike_int` is `fields.Int` with `Range(min=1)`, which accepts a 401-digit integer; `math.isfinite()` raises `OverflowError` on that, and `ValueError` on `Decimal("sNaN")`, so both have to be refused explicitly rather than assumed away.

## Changes

`services/option_symbol_service.py`

- Add `VALID_OPTION_TYPES`, `validate_option_type()` and `validate_strike_interval()`.
- Apply both at the top of `get_option_symbol()`, **before** the quote request and the database lookup, so a rejected request never reaches the broker. The raised `ValueError` is answered by the existing handler as HTTP 400 with the reason.
- Apply them defensively inside `get_atm_strike()`, `calculate_offset_strike()` and `calculate_offset_strike_from_actual()` too, so the helpers are safe for any caller.
- `validate_strike_interval()` returns `float(strike_int)`, not the value unchanged, so the next line is guaranteed to be able to divide by it: `Decimal("50")` is finite and positive but `float / Decimal` raises `TypeError`. The same conversion makes the numpy-scalar acceptance uniform and covers `Fraction` for free.
- `_is_finite_number()` uses `math.isfinite()` rather than an `isinstance()` check so a numpy scalar off a pandas or DuckDB path still passes (`np.int64` is not an `int`), while strings, `None` and sequences are refused. It catches `(TypeError, OverflowError, ValueError)`, so an int too large to convert to a float (previously a 500) and a signalling NaN (previously a 400 reading `cannot convert signaling NaN to float`) are both refused with the interval's own error message. `bool` is excluded explicitly, since it is an `int` subclass and `True` would otherwise pass as an interval of 1.
- Refuse a non-finite LTP with a controlled error instead of building a symbol around `NaN` or infinity.

Contract change worth calling out: `find_atm_strike_from_actual()` now returns `None` for a non-numeric LTP where it previously raised `TypeError`. Every caller (`custom_straddle_service.py:138`, `iv_chart_service.py:242` and `:548`, `option_chain_service.py:491`, `straddle_chart_service.py:219`, `vol_surface_service.py:141`) already branches on `None`, so a bad underlying LTP now degrades to a clean "no data" answer instead of a 500.

## What does not change

Valid input is untouched: calls and puts, lowercase `ce`/`pe`, fractional strike intervals such as `2.5`, the `ATM`/`ITM`/`OTM` maths, out-of-range offsets still returning `None`, and the existing error contracts. The only new rejections are values that previously crashed or resolved to the wrong strike.

## Testing

`test/test_option_symbol_service_validation.py` (new, 145 cases) covers both legacy paths:

- Valid calls, puts, lowercase types and fractional intervals resolve exactly as before.
- Zero, negative, NaN, infinite, `None`, string and `bool` intervals raise `ValueError`, as do a 401-digit integer and `Decimal("sNaN")`.
- An accepted interval comes back as something the caller can divide by, `Decimal("50")` included, and resolves the same strike as the equivalent `int`.
- Unsupported option types are rejected in both offset calculations instead of being priced as puts.
- At the boundary, `get_option_symbol()` returns `(False, {...}, 400)` and the quote stub records **no** call, proving the broker is never contacted for a rejected request.
- A non-finite LTP is reported rather than resolved into a symbol.

No broker credentials, no network, no database rows: the quote fetch, strike lookup and symbol lookup are stubbed with `monkeypatch`, and test values are obvious dummies (`test-api-key-not-real`).

```
uv run pytest test/test_option_symbol_service_validation.py -v
145 passed

uv run ruff check services/option_symbol_service.py test/test_option_symbol_service_validation.py
All checks passed!
```

Run against the unfixed source, 12 of the 13 new cases fail; the single case that passes on both is the positive-contract test.

Regression check on the neighbouring suites, plus the full CI-safe list, all green:

```
uv run pytest test/test_mcx_underlying_resolution.py test/test_data_schemas_validation.py \
  test/test_flow_qa_regressions.py test/test_flow_workflow_validator.py \
  test/test_multi_option_greeks_regression.py
568 passed

# the CI-safe subset from ci.yml, including the new file
217 passed
```

The new file is added to the CI-safe list in `.github/workflows/ci.yml`, following the same pattern as `test_event_bus_bounded.py` and `test_telegram_api_contract.py`. The branch is rebased onto current `main`, which resolves the `ci.yml` conflict with the new Python 3.12-3.14 test matrix.

## Checklist

- [x] One fix, self-contained, no unrelated changes
- [x] Conventional Commit message
- [x] Python tests added and passing, with no broker credentials or network access
- [x] `ruff check` clean on both touched Python files (the file's pre-existing `ruff format` differences are left alone to keep the diff reviewable)
- [x] No new dependencies
- [x] No API contract change for valid requests

---
## Summary by cubic
Validate strike interval and option type at the option-symbol resolver to prevent wrong-leg placements and 500s. Previously, invalid intervals crashed or produced invalid symbols and unknown types defaulted to puts; now we reject these with 400 and report non-finite LTPs without symbol resolution (500).

- Apply input guards at the start of get_option_symbol() and inside get_atm_strike(), calculate_offset_strike(), and calculate_offset_strike_from_actual(); add VALID_OPTION_TYPES, exclude bools from numeric checks, and return a float from strike interval validation so callers can safely divide.
- find_atm_strike_from_actual() returns None for non-finite/non-numeric LTPs instead of selecting the lowest strike.
- Non-finite LTPs in get_option_symbol() return a controlled 500 without constructing a symbol; rejected requests never hit the broker.
- Valid CE/PE (any case), fractional intervals, offsets, and existing error contracts are unchanged.
- Tests: add test/test_option_symbol_service_validation.py; include it in CI.

<sup>Written for commit da5c6c3fc1f1178b9626bd0d475cf03c927e2c25. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1892?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,6 +58,7 @@ jobs:
             test/test_multi_option_greeks_regression.py \
             test/test_history_format.py \
             test/test_precompress_assets.py \
+            test/test_option_symbol_service_validation.py \
             -v --timeout=60
 
   # Frontend linting
--- a/services/option_symbol_service.py
+++ b/services/option_symbol_service.py
@@ -34,6 +34,7 @@ Example Usage (OLD METHOD - Legacy):
 """
 
 import importlib
+import math
 import re
 from datetime import datetime
 from typing import Any, Optional
@@ -169,6 +170,85 @@ def resolve_underlying_quote(base_symbol: str, exchange: str) -> tuple[str, str]
     return (fut["symbol"], fut["exchange"])
 
 
+# ============================================================================
+# INPUT VALIDATION - Guards for the resolver boundary
+# ============================================================================
+#: Option types this resolver understands. Every strike calculation below
+#: branches on CE and treats anything else as a put, so an unrecognised type
+#: such as "CALL", "C" or None used to resolve to the put strike instead of
+#: being refused. The symbol built around it then failed the master-contract
+#: lookup, so the caller saw a confusing "not found" where the real fault was
+#: the option type.
+VALID_OPTION_TYPES = ("CE", "PE")
+
+
+def _is_finite_number(value: Any) -> bool:
+    """Whether a value is a real number that is neither NaN nor infinite.
+
+    math.isfinite() is used rather than an isinstance() check so that a numpy
+    scalar off a pandas or DuckDB path is still accepted - np.int64 is not an
+    int - while strings, None and sequences raise TypeError and are refused.
+    bool is excluded explicitly because it is an int subclass, so True would
+    otherwise pass as a strike interval of 1.
+
+    OverflowError and ValueError are refused alongside TypeError: an int too
+    large to convert to a float raises OverflowError, and Decimal("sNaN")
+    raises ValueError. Both are reachable from the public endpoint, because
+    OptionSymbolSchema validates a range, not a magnitude or a numeric type.
+    """
+    if isinstance(value, bool):
+        return False
+    try:
+        return math.isfinite(value)
+    except (TypeError, OverflowError, ValueError):
+        return False
+
+
+def validate_option_type(option_type: Any) -> str:
+    """Normalize an option type to CE or PE.
+
+    Args:
+        option_type: Option type supplied by the caller, in any case.
+
+    Returns:
+        "CE" or "PE".
+
+    Raises:
+        ValueError: If the option type is missing or unsupported.
+    """
+    if isinstance(option_type, str):
+        normalized = option_type.strip().upper()
+        if normalized in VALID_OPTION_TYPES:
+            return normalized
+
+    raise ValueError(f"Invalid option_type: {option_type!r}. Supported option types are CE and PE.")
+
+
+def validate_strike_interval(strike_int: Any) -> float:
+    """Validate a strike interval before it is used as a divisor.
+
+    Args:
+        strike_int: Strike interval supplied by the caller.
+
+    Returns:
+        The strike interval as a float, so every caller is guaranteed a value
+        it can divide by. Returning it unchanged would not be: Decimal("50")
+        is finite and positive, but float / Decimal raises TypeError. The
+        conversion also covers Fraction and numpy scalars uniformly.
+
+    Raises:
+        ValueError: If the interval is not a positive finite number. Zero
+            raises ZeroDivisionError inside the ATM calculation, and a
+            negative, NaN or infinite interval produces a strike that no
+            exchange lists.
+    """
+    if not _is_finite_number(strike_int) or strike_int <= 0:
+        raise ValueError(
+            f"Invalid strike_int: {strike_int!r}. Strike interval must be a positive number."
+        )
+    return float(strike_int)
+
+
 def get_atm_strike(ltp: float, strike_int: int) -> float:
     """
     Calculate ATM strike price based on LTP and strike interval.
@@ -180,10 +260,14 @@ def get_atm_strike(ltp: float, strike_int: int) -> float:
     Returns:
         ATM strike price rounded to nearest strike interval
 
+    Raises:
+        ValueError: If strike_int is not a positive finite number.
+
     Example:
         LTP = 23587.50, strike_int = 50
         ATM = round(23587.50 / 50) * 50 = 23600
     """
+    strike_int = validate_strike_interval(strike_int)
     atm_strike = round(ltp / strike_int) * strike_int
     logger.info(f"Calculated ATM: LTP={ltp}, strike_int={strike_int}, ATM={atm_strike}")
     return atm_strike
@@ -213,12 +297,17 @@ def calculate_offset_strike(
             - ITM: ATM + (N * strike_int)  [Higher strike]
             - OTM: ATM - (N * strike_int)  [Lower strike]
 
+    Raises:
+        ValueError: If option_type is not CE or PE, if strike_int is not a
+            positive finite number, or if the offset is unrecognised.
+
     Example:
         ATM = 23600, strike_int = 50, option_type = "CE", offset = "ITM2"
         Target = 23600 - (2 * 50) = 23500
     """
+    option_type = validate_option_type(option_type)
+    strike_int = validate_strike_interval(strike_int)
     offset = offset.upper()
-    option_type = option_type.upper()
 
     if offset == "ATM":
         target_strike = atm_strike
@@ -468,6 +557,12 @@ def find_atm_strike_from_actual(ltp: float, available_strikes: list) -> float | 
         logger.warning("No available strikes to find ATM")
         return None
 
+    if not _is_finite_number(ltp):
+        # min() with a NaN key compares false against every strike and returns
+        # the first one, which would then be reported as the ATM strike.
+        logger.error(f"Cannot find ATM strike from a non-numeric LTP: {ltp!r}")
+        return None
+
     # Find the strike closest to LTP
     atm_strike = min(available_strikes, key=lambda x: abs(x - ltp))
 
@@ -490,6 +585,9 @@ def calculate_offset_strike_from_actual(
     Returns:
         Target strike price or None if offset is out of range
 
+    Raises:
+        ValueError: If option_type is not CE or PE.
+
     Logic:
         For CE (Call):
             - ITM: Lower strikes (traverse down the list from ATM)
@@ -507,12 +605,13 @@ def calculate_offset_strike_from_actual(
         For CE ITM2: Move 2 positions DOWN from ATM
         Result: 23400 (actual strike from database)
     """
+    option_type = validate_option_type(option_type)
+
     if not available_strikes or atm_strike not in available_strikes:
         logger.error(f"ATM strike {atm_strike} not found in available strikes")
         return None
 
     offset = offset.upper()
-    option_type = option_type.upper()
 
     # Find the index of ATM in the sorted strikes list
     atm_index = available_strikes.index(atm_strike)
@@ -623,6 +722,15 @@ def get_option_symbol(
         Tuple of (success, response_data, status_code)
     """
     try:
+        # Step 0: Refuse unusable inputs before spending a quote request or a
+        # database lookup on them. Flow, multi-leg orders and MCP call this
+        # resolver directly, without the REST schema that checks these two
+        # fields, so the resolver validates them for itself. The ValueErrors
+        # raised here are answered as HTTP 400 at the bottom of this function.
+        option_type = validate_option_type(option_type)
+        if strike_int is not None:
+            strike_int = validate_strike_interval(strike_int)
+
         # Step 1: Parse underlying to extract base symbol and expiry
         base_symbol, embedded_expiry = parse_underlying_symbol(underlying)
 
@@ -722,6 +830,19 @@ def get_option_symbol(
                 )
 
             logger.info(f"Got LTP: {ltp} for {quote_symbol}")
+
+        # A non-finite LTP resolves to a strike no exchange lists, so stop
+        # here rather than constructing a symbol around NaN or infinity.
+        if not _is_finite_number(ltp):
+            logger.error(f"Unusable LTP {ltp!r} for {quote_symbol}")
+            return (
+                False,
+                {
+                    "status": "error",
+                    "message": f"Could not determine a usable LTP for {quote_symbol}.",
+                },
+                500,
+            )
 
         # Step 4: Map to options exchange
         options_exchange = get_option_exchange(quote_exchange)
--- a/test/test_option_symbol_service_validation.py
+++ b/test/test_option_symbol_service_validation.py
@@ -1,0 +1,347 @@
+"""The option-symbol resolver refuses unusable strikes and option types.
+
+Two defects lived in the strike calculations. The legacy strike_int path
+divided the underlying LTP by the caller's interval with no guard, so a zero
+interval raised ZeroDivisionError and a NaN or infinite one produced a strike
+that formats into a symbol no exchange lists. Separately, every offset
+calculation branches on CE and treats anything else as a put, so an option
+type of "CALL", "C" or None resolved to the put strike. The symbol built
+around it then failed the master-contract lookup, so the caller saw a
+confusing "not found in NFO" where the real fault was the option type.
+
+The REST schema validates both fields, but services/flow_openalgo_client.py
+calls get_option_symbol() in process and bypasses it, so the fix and these
+tests sit at the resolver boundary rather than in the schema.
+
+The tests are fully local: no broker credentials, no network, and no database
+rows - the quote fetch and the symbol lookup are stubbed.
+"""
+
+import math
+import os
+import sys
+from decimal import Decimal
+
+import numpy
+import pytest
+
+sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+
+from services import option_symbol_service as oss  # noqa: E402
+from services.option_symbol_service import (  # noqa: E402
+    VALID_OPTION_TYPES,
+    calculate_offset_strike,
+    calculate_offset_strike_from_actual,
+    find_atm_strike_from_actual,
+    get_atm_strike,
+    get_option_symbol,
+    validate_option_type,
+    validate_strike_interval,
+)
+
+#: A realistic NIFTY chain: 50-point strikes either side of a 23600 ATM.
+STRIKES = [23400.0, 23450.0, 23500.0, 23550.0, 23600.0, 23650.0, 23700.0, 23750.0]
+
+#: Values that are not a usable divisor. Zero and the negatives are the
+#: reported defect; True is included because bool is an int subclass and would
+#: otherwise slip through as an interval of 1.
+UNUSABLE_INTERVALS = [
+    0,
+    0.0,
+    -50,
+    -0.5,
+    float("nan"),
+    float("inf"),
+    float("-inf"),
+    None,
+    "50",
+    "",
+    True,
+    [50],
+    #: math.isfinite() raises rather than returning False for these two: an int
+    #: too large to convert to a float raises OverflowError, and a signalling
+    #: NaN raises ValueError. Both are reachable from the public endpoint.
+    10**400,
+    Decimal("sNaN"),
+]
+
+#: Everything the resolver must refuse rather than silently price as a put.
+UNSUPPORTED_OPTION_TYPES = ["CALL", "PUT", "C", "P", "XX", "", "  ", None, 0, 1, ["CE"]]
+
+
+class TestValidateOptionType:
+    @pytest.mark.parametrize(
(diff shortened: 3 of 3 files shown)


Kalaiviswa:
Merging this after two one-line fixes to `_is_finite_number` and a correction to the description. The bug is real, the validation is placed where your own rationale says it has to be, and the tests genuinely lock it in. What has to change is that the central helper contradicts its own docstring in two directions, and the description credits four bypasses that are not bypasses.

**What I verified** (base `447786fab` and head `65e3105` each checked out clean, tests run against both; line numbers post-patch unless I say otherwise):

- The two defects reproduce on main. `calculate_offset_strike(23600, "ITM2", 50, "CALL")` returns **23700**, which is the PE ITM2 strike (CE is 23500). Same for `"C"`, `"XX"` and `""`, and `calculate_offset_strike_from_actual` is identical. `get_atm_strike(23587.5, 0)` raises `ZeroDivisionError`. `find_atm_strike_from_actual(nan, strikes)` returns **23400**, the lowest strike, because every NaN comparison is false so `min()` yields element 0.
- The issue's line numbers are stale and should not be trusted by the next reader: the divisor is at `:187`, not `:200`, and the option-type `else` branches are at `:231`/`:240` and `:531`/`:540`, not `:249-276`/`:543-565` (those ranges are a docstring and the invalid-offset `else`, which is correct code).
- The `ValueError` to HTTP 400 claim holds. It relies on the pre-existing `except ValueError` at `:944` (base `:833`), which this PR does not add, and `restx_api/option_symbol.py:91` returns the status verbatim. Confirmed end to end with the quote and DB stubbed.
- Valid behaviour is untouched, and provably so. I ran a base-vs-head differential over CE/PE/ce/pe by ATM/ITM1-5/OTM1-5 by `strike_int` {50, 2.5} by several LTPs. The output diff is empty apart from the five intended non-finite-LTP lines.
- `find_atm_strike_from_actual` now returning `None` is safe at every call site: `custom_straddle_service.py:138`, `iv_chart_service.py:242` and `:548`, `option_chain_service.py:491`, `straddle_chart_service.py:219`, `vol_surface_service.py:141` all already branch on `is None`.
- The tests are not decorative. 132 pass on head; against unfixed main **103 of 132 fail**, and 67 of those are genuine behavioural detections rather than "the validator does not exist yet" (`DID NOT RAISE ValueError`, `ZeroDivisionError at :187`, `assert 23400.0 is None`, `assert (True, 200) == (False, 400)`). The 29 that pass on both are exactly the positive-contract tests, which is the correct shape.
- No regressions: 16 pre-existing option, flow, order and MCP test files give `981 passed, 3 skipped` identically on base and head.
- `ruff check` is clean on both changed files. `ruff format --check` still flags `option_symbol_service.py`, but that is pre-existing on base at its line 681 and the CI step is `continue-on-error: true`.

One CI note: all 11 checks are green, but the cubic reviewer did not actually review this PR. Its output is "You've reached your plan's monthly review limit. cubic has reviewed 81,254 of the 80,000 allowed lines of code this month. Reviews resume on 1 September 2026." So nothing automated has looked at this.

## Required before merge

**1. `services/option_symbol_service.py:194-197`: `_is_finite_number` catches only `TypeError`.**

`math.isfinite()` also raises `OverflowError` for a Python int too large to convert to a float, and `ValueError` for `Decimal("sNaN")`. `OptionSymbolSchema.strike_int` is `fields.Int` with `validate.Range(min=1)`, which happily accepts a 401-digit integer, so:

```
POST /api/v1/optionsymbol  {"strike_int": 10**400, ...}
-> HTTP 500  "An error occurred ...: int too large to convert to float"
```

That is the exact case this validator claims to close, and it is reachable from the public endpoint. Base returns 500 here too, so it is not a regression, but the control is incomplete. Fix:

```python
    except (TypeError, OverflowError, ValueError):
        return False
```

**2. `services/option_symbol_service.py:235-239`: `validate_strike_interval` accepts a value the next line cannot divide by.**

`math.isfinite(Decimal("50"))` is `True`, so the validator returns `Decimal('50')` "unchanged", and then `round(ltp / strike_int)` raises `TypeError: unsupported operand type(s) for /: 'float' and 'decimal.Decimal'`, giving a 500. The docstring says the function validates the interval "before it is used as a divisor", and as written it does not guarantee that. Return a float instead of passing through:

```python
    return float(strike_int)
```

That also makes the `np.int64` acceptance in your rationale uniform rather than accidental, and covers `Fraction` and 0-d arrays for free.

**3. The description credits four bypasses that are not bypasses.**

This matters because the description becomes the merge commit. Verified one by one:

- `mcp/mcpserver.py` is **not** a bypass. Line 12 and line 50 construct the openalgo HTTP SDK client, so `client.optionsymbol(...)` goes over HTTP through `OptionSymbolSchema` like any other REST caller.
- `services/synthetic_future_service.py` is **not** a bypass. Lines 50-57 and 93-101 hardcode `strike_int=None` and `option_type="CE"`/`"PE"`. No user value reaches the resolver.
- `place_options_order_service.py` and `options_multiorder_service.py` bypass `OptionSymbolSchema`, but their own endpoints validate first: `OptionsOrderSchema` (`restx_api/schemas.py:215-221`) and `OptionsMultiOrderLegSchema` (`:254-255`), both `OneOf(["CE","PE","ce","pe"])` with `strike_int` on `Range(min=1)`.

The one genuine in-process bypass is `services/flow_openalgo_client.py:397`, and its caller `services/flow_executor_service.py:2086` reads `optionType` through `get_str` with no enum check, unlike the options-order nodes at `:788` and `:919` which use `frozenset({"CE","PE"})`. That single caller justifies the whole change. Keep it and drop the other four.

While you are in there: the "wrong leg placed" framing does not hold end to end. A bogus `option_type` builds a symbol carrying the same bogus suffix, so `find_option_in_database` returns 404 and the path fails closed. I confirmed this with the quote and DB stubbed. The real harm today is a confusing "not found in NFO" where the user should have seen "invalid option type", which is still worth fixing, so say that instead.

## Ordering

Base is current `main` (`447786fab`), mergeable state clean, no drift on the touched file.

No collision with the MCX options work on my side either. `git merge-tree pr-1892 main` writes a clean tree, and on that merged tree `test_option_symbol_service_validation.py` plus `test_flow_mcx_options.py` give **168 passed**. The two changes touch `get_option_symbol` without overlapping, and the ordering is the right way round: your Step 0 guard lands at `:720-722` and the MCX near-month branch at `:785`, so a bad `option_type` on a commodity leg is refused with a 400 before the futures lookup runs, instead of surfacing as a 404 further down.

## Optional, not blocking

- **`offset` is still unvalidated at the same boundary.** `offset=None` gives 500 `'NoneType' object has no attribute 'upper'`, and it arrives from the same `leg_data.get()` on the same call as the two fields you did fix. `OptionsOrderSchema.offset` (`schemas.py:218`) is a bare `fields.Str`. It fails closed, so this is error-class only, but a third `validate_offset()` would finish the boundary.
- **`get_atm_strike` has no LTP guard**, which is asymmetric with the two places that got one. `get_atm_strike(inf, 50)` raises `OverflowError`, which the `except ValueError` handler does not catch, so a direct caller gets a 500. Its new `Raises:` docstring is also silent about `ltp`.
- **`np.bool_(True)` passes `validate_strike_interval`.** The docstring excludes `bool` explicitly and accepts numpy scalars explicitly, and `np.bool_` is neither, so it lands in the gap between the two stated intents.
- **"Positive and finite" is not "plausible".** `strike_int=1e-9` passes and builds `NIFTY28OCT2523587.499999998CE`. In production that fails closed at the master-contract lookup, so it is cosmetic.
- **`VALID_OPTION_TYPES` as a tuple** is asserted verbatim by `test_only_ce_and_pe_are_supported`, which locks the type in for no reason. A `frozenset` matches this file's own idiom (`NO_SPOT_EXCHANGES` at `:84`) and `flow_executor_service.py:788`.
- **The `find_atm_strike_from_actual` contract change is not in the description.** It goes from raising `TypeError` to returning `None` for a non-numeric LTP, and five external callers see it. `vol_surface_service.py:141` turns that into `continue`, so a bad underlying LTP now silently skips every expiry and returns 404 "No valid expiry data found" instead of a 500. That is arguably the better behaviour, but it is a contract change to a shared helper and should be stated.

Narasimha722:
Thanks for the depth here - particularly for running the base-vs-head differential and for checking the bypass list caller by caller. Both required code changes are in, the description is corrected, and the branch is rebased onto current `main`.

**1. `_is_finite_number` now catches `(TypeError, OverflowError, ValueError)`**

You are right that this was the exact case the validator claimed to close and did not. Confirmed both directions locally with the quote and DB stubbed:

```
strike_int = 10**400
  before: 500  An error occurred ...: int too large to convert to float
  after:  400  Invalid strike_int: 1000...000. Strike interval must be a positive number.

strike_int = Decimal("sNaN")
  before: 400  cannot convert signaling NaN to float
  after:  400  Invalid strike_int: Decimal('sNaN'). Strike interval must be a positive number.
```

The `sNaN` case is one detail different from your write-up: it already returned 400, because the `float()` call raises `ValueError` and the existing handler catches it - but with a message that tells the caller nothing about which field was wrong. The `10**400` case is a 500 exactly as you said, since `OverflowError` walks past that handler.

Both values are now in the `UNUSABLE_INTERVALS` list the parametrised tests run over, so they are covered in `validate_strike_interval`, `get_atm_strike`, `calculate_offset_strike` and at the `get_option_symbol` boundary.

**2. `validate_strike_interval` returns `float(strike_int)`**

Also correct, and the docstring contradiction was the giveaway:

```
strike_int = Decimal("50")
  before: 500  An error occurred ...: unsupported operand type(s) for /: 'float' and 'decimal.Decimal'
  after:  200  NIFTY28OCT2523600CE
```

It now returns a float, so the guarantee the docstring makes is the guarantee the caller gets. Two tests lock it in: one asserts an accepted interval is a usable divisor for `int`, `float`, `np.int64` and `Decimal`, and one asserts `get_atm_strike(23587.50, Decimal("50")) == get_atm_strike(23587.50, 50)`.

Checked that returning a float does not leak into symbol text: `construct_option_symbol` already normalises with `if strike == int(strike)`, so `23600.0` still formats as `23600`.

**3. Description corrected**

- The bypass list now names only `services/flow_openalgo_client.py:397` and its unchecked `optionType` at `services/flow_executor_service.py:2086`, and explicitly states that `mcp/mcpserver.py` (HTTP via the SDK client) and `services/synthetic_future_service.py` (hardcoded) are not bypasses, and that the two options-order services validate through their own schemas first.
- The "wrong leg placed" framing is gone. It now says what actually happens: the bogus suffix flows into the symbol, `find_option_in_database` returns 404, the path fails closed, and the real cost is a misleading "not found in NFO" plus a wasted quote call.
- Added the point about `Range(min=1)` not being enough on its own, which is what makes the `10**400` case reachable.
- Added the `find_atm_strike_from_actual` contract change (`TypeError` -> `None`) with the caller list, per your optional note. Worth stating explicitly, as you say.

The same two corrections are in the commit message body, since that is what lands in history.

**4. Conflict resolved**

Rebased onto `5fd89238b`. The only conflict was the CI-safe test list in `ci.yml`, which collided with the new 3.12-3.14 matrix from #1894. Resolution keeps your list intact and appends one line. `ci.yml` is now `+1` against `main`.

**Test counts after the change**

```
uv run pytest test/test_option_symbol_service_validation.py -q
145 passed          (132 before; 13 new cases, 12 of which fail against unfixed source)

# the full CI-safe list from ci.yml, including this file
217 passed

uv run pytest test/test_mcx_underlying_resolution.py test/test_data_schemas_validation.py \
  test/test_flow_qa_regressions.py test/test_flow_workflow_validator.py \
  test/test_multi_option_greeks_regression.py
568 passed

uv run ruff check services/option_symbol_service.py test/test_option_symbol_service_validation.py
All checks passed!
```

`ruff format --check` is clean on the new test file; the pre-existing drift in `option_symbol_service.py` is left alone for the reason in the checklist.

**On the optional items**

Not taken here, to keep this one fix and avoid a second review pass:

- `validate_offset()` and an LTP guard in `get_atm_strike` are the same class of defect at the same boundary, and I agree the boundary is not finished without them. Happy to open a follow-up issue and PR for both together if you would like them.
- `np.bool_` in the gap between the two stated intents, and `strike_int=1e-9` being positive but not plausible, are both real. Both would ride along with the follow-up.
- `VALID_OPTION_TYPES` as a `frozenset` matching the file's own `NO_SPOT_EXCHANGES` idiom is a fair point; leaving the tuple only because the change would touch a test assertion for no behavioural gain.

Also noting for the record that the cubic check did not review this PR - its output is a plan-limit message, so nothing automated has looked at it.

marketcalls:
Merged as 0c41679d, with you as the commit author.

The only thing standing in the way was a conflict in `.github/workflows/ci.yml`, and it was mine, not yours: #1914 and #1916 both landed earlier today and each appended to the same CI-safe test list you were appending to. Rather than send you round for a third rebase over one line, I resolved it here by keeping both entries.

Verified before merging, independently rather than from the description:

- All five defects reproduce on current `main`. `calculate_offset_strike(23600, "ITM2", 50, "CALL")` returns 23700, the put strike, and logs it as if correct. `strike_int=0` raises ZeroDivisionError, `Decimal("50")` raises TypeError, `10**400` raises OverflowError, and `find_atm_strike_from_actual(nan, ...)` returns the lowest strike in the chain.
- All five are fixed on the branch, with valid input untouched: CE, PE, lowercase `ce`, fractional `2.5`, plus `sNaN` and `bool` now refused.
- The float return does not leak into symbol text. `construct_option_symbol` renders `23500.0` as `NIFTY28OCT2523500CE`, and `292.5` still renders `292.5`, so the documented `VEDL25APR24292.5CE` shape is intact. That was the one thing in the change that could have been a quiet regression.
- The tests are not decorative. Swapping in main's unfixed service only errors at collection, since the module imports `VALID_OPTION_TYPES`, so that measurement is not available. Neutering the two validators' behaviour while leaving their names in place fails **91 of 145**, and the 54 that still pass are exactly the positive-contract tests. That is the right shape.
- No regressions. The full CI-safe list, 21 files including yours, gives 243 passed. The five neighbouring suites give 568 passed. `ruff check` clean on both changed files.
- The guards sit at Step 0, ahead of the parse and the quote exchange, so a rejected request never reaches the broker.

One small thing for next time, not worth holding this for: both errors echo the input through `!r`, so `strike_int=10**400` puts a 401-character value into the response and the log. It is JSON-escaped, so not injection, but truncating would be tidier on an endpoint that accepts that input.

Please do open the follow-up you offered, for `validate_offset()` and the LTP guard in `get_atm_strike()`. Your own reasoning is that the boundary is not finished without them, and I agree. `np.bool_` and the `frozenset` change can ride along.

This was unusually good work: the description argued why the fix belongs at the resolver rather than in a fourth schema, named the single genuine bypass after checking the others one by one, and stated a contract change to a shared helper along with its call sites. You also corrected the review where it was wrong about `sNaN` already returning 400. Thank you.

mevin1491 (commented):
good

PR #1914, opened 2026-08-26
fix: redact API keys from core service logs

### LINKED ISSUE #1855: security: Remove API-key material from core service logs

solstxce:
## Summary

- Remove API keys from WebSocket close-failure logs.
- Replace invalid API-key prefixes in both option Greeks endpoints with short SHA-256 fingerprints.
- Add regression tests covering log capture, redaction, exact fingerprints, and preserved 401 responses.
- Run the credential-log regression in the explicit backend CI-safe test list.

Closes #1855

## Validation

- Local focused test: uv run pytest test/test_core_credential_logging.py -v, 3 passed.
- Ruff check and format check passed for test/test_core_credential_logging.py.
- git diff --check passed.
- GitHub Actions is running the updated CI-safe suite on Python 3.12, 3.13, and 3.14.

The full targeted Ruff command reports four pre-existing modernization findings in services/websocket_client.py that are outside the changed lines.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,6 +58,7 @@ jobs:
             test/test_telegram_api_contract.py test/test_cors_config.py \
             test/test_market_calendar_validation.py \
             test/test_admin_client_error_url.py \
+            test/test_core_credential_logging.py \
             test/test_multi_option_greeks_regression.py \
             test/test_history_format.py \
             test/test_precompress_assets.py \
--- a/restx_api/multi_option_greeks.py
+++ b/restx_api/multi_option_greeks.py
@@ -1,3 +1,4 @@
+import hashlib
 import os
 
 from flask import jsonify, make_response, request
@@ -103,7 +104,7 @@ class MultiOptionGreeks(Resource):
 
             # Verify API key
             if not verify_api_key(api_key):
-                logger.warning(f"Invalid API key used for multi option greeks: {api_key[:10]}...")
+                logger.warning(f"Invalid API key used for multi option greeks: sha256:{hashlib.sha256(api_key.encode()).hexdigest()[:12]}")
                 return make_response(
                     jsonify({"status": "error", "message": "Invalid openalgo apikey"}), 401
                 )
--- a/restx_api/option_greeks.py
+++ b/restx_api/option_greeks.py
@@ -1,3 +1,4 @@
+import hashlib
 import os
 
 from flask import jsonify, make_response, request
@@ -113,7 +114,7 @@ class OptionGreeks(Resource):
 
             # Verify API key
             if not verify_api_key(api_key):
-                logger.warning(f"Invalid API key used for option greeks: {api_key[:10]}...")
+                logger.warning(f"Invalid API key used for option greeks: sha256:{hashlib.sha256(api_key.encode()).hexdigest()[:12]}")
                 return make_response(
                     jsonify({"status": "error", "message": "Invalid openalgo apikey"}), 401
                 )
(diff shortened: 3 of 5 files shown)


marketcalls (approved):
Merging. Marking it ready first, since it was still a draft.

I verified the premise rather than taking it from the issue. Both original lines really do survive `SensitiveDataFilter`, so this was a live leak and not a theoretical one:

```
Invalid API key used for option greeks: SEKRETKEY1...   -> LEAKS
Error closing client for API key SEKRETFULLKEY: boom    -> LEAKS
```

The reason is worth knowing, because it decides whether any given log line is safe: the filter matches key-value shapes such as `api_key=` or `token:`. "API key" with a space matches nothing, and in the first line a word sits between the keyword and the colon. Wording, not intent, is what determined the outcome.

I also verified the tests are not vacuous. Reverting both fixes to their leaking originals fails exactly the two corresponding tests, and the third keeps passing because it covers the file I did not revert. They assert on captured log output rather than re-implementing the check, so they will keep working after a refactor.

`close_all_clients` is clean too. `logger.exception(f"Error closing client: {e}")` still interpolates the exception, which is the shape that leaked in `definedge/api/auth_api.py` where `httpx.HTTPStatusError` embedded a credential-bearing URL. Traced it here: `ws_url` is `ws://{host}:{port}` with no credential, and `disconnect()` only sets flags and schedules a coroutine, so the exception cannot carry the key. Safe as written.

Merges clean onto current `main`, and clean against #1916, which also edits the CI-safe list but inserts at a different point. 15 checks pass. Adding the test to that list is the right move: a redaction regression that only runs locally is one nobody notices.

## Two notes, neither blocking

**The fingerprint has no domain separator.** The issue's Scope asked for "a short one-way diagnostic fingerprint with a documented domain separator" and this is a plain unsalted `sha256(api_key)[:12]`. I would keep it as is: it matches what the repo already does in `.claude/skills/security-audit/audit.py:81` (`sha256:{...[:12]}`) and `database/auth_db.py:960`. Consistency with the existing convention is worth more here than the issue's letter, and keys are generated from `secrets`, so a preimage is not the concern. A separator would only prevent correlating the same key across contexts.

**The expression is inlined twice.** `hashlib.sha256(api_key.encode()).hexdigest()[:12]` appears in both endpoints, and `audit.py` has a third copy of the same idea in a `fingerprint()` helper. Worth extracting to a shared util so the format has one definition, otherwise the three drift.

Good use of `caplog` and of the mocked logger to assert on the real call arguments. That is the form of test that keeps paying off.

PR #1916, opened 2026-08-27
fix: remove credentials from broker logs

### LINKED ISSUE #1854: security: Remove credentials from non-XTS broker authentication logs
## Summary

Multiple non-XTS integrations log complete tokens, token prefixes, OTP request bodies, headers, or credential-bearing responses during authentication and funds requests.

## Evidence

- `broker/fivepaisa/api/auth_api.py:63-90` logs token requests and responses.
- `broker/mstock/api/auth_api.py:116` and `:258-272` log tokens, headers, OTP/request data, and raw responses; `broker/mstock/api/funds.py:24` logs a token prefix.
- `broker/groww/api/funds.py:16` logs the auth token.
- `broker/flattrade/api/auth_api.py:24-30`, `broker/dhan/api/auth_api.py:41,103,118`, `broker/paytm/api/auth_api.py:43`, and `broker/kotak/database/master_contract_db.py:317` expose full or partial credential material.

## Why this matters

Logs are often retained, shared for support, and included in diagnostics. Redacting only part of a bearer token is not a safe substitute for omitting it.

## Scope

Replace credential-bearing logs in the listed modules with safe status metadata. Keep broker response errors useful by selecting non-sensitive fields rather than dumping whole payloads.

## Acceptance criteria

- [ ] No listed log statement emits token content, API-key fragments, OTP values, auth headers, or raw credential responses.
- [ ] Safe status, length, or presence metadata remains where operationally useful.
- [ ] A static regression test covers the listed patterns/files.
- [ ] No authentication request or response handling behavior changes.

## Verification

```powershell
uv run pytest test/test_broker_credential_logging.py -v
uv run ruff check broker/fivepaisa/api/auth_api.py broker/mstock/api/auth_api.py broker/mstock/api/funds.py broker/groww/api/funds.py broker/flattrade/api/auth_api.py broker/dhan/api/auth_api.py broker/paytm/api/auth_api.py broker/kotak/database/master_contract_db.py test/test_broker_credential_logging.py
```

## Security guidance

- Difficulty: Medium
- Estimated time: 4-8 hours
- You will learn: credential redaction, safe operational logging, and multi-file regression tests.

solstxce:
## Summary

Remove credential material from authentication and funds-related logs in the broker modules identified by #1854.

## Changes

- Removed full authentication response logging from 5paisa and Paytm.
- Removed mStock client-code, token-prefix, response-body, response-header, and raw error logging.
- Removed mStock and Groww funds payload/token logging.
- Removed Flattrade request-payload, response-body, and exception-value logging.
- Removed Dhan client ID, API-secret prefix, consent ID, and additional authentication-data logging.
- Removed Kotak access-token prefix and raw scripmaster response logging.
- Replaced sensitive values with safe status, presence, length, field-name, content-type, and exception-type metadata where useful.
- Preserved authentication request/response handling and caller-visible error values.
- Added a static regression test covering all eight files listed in the issue.
- Added the regression test to the CI-safe backend test selection.

## Security Coverage

The updated log statements no longer emit:

- Access tokens, refresh tokens, or token prefixes
- API keys, API-secret fragments, or client-code values
- OTP-related credential material
- Authentication request payloads or response payloads
- Response headers that may contain session credentials
- Raw credential-bearing HTTP error responses

## Issue

Closes #1854

## Verification

- `uv run pytest test/test_broker_credential_logging.py -v --timeout=60`: 1 passed
- Maintained CI-safe backend selection including the new test: 95 passed
- `uv run ruff check` on all changed files: passed
- `uv run ruff format --check` on all changed files: passed
- `git diff --check`: passed
- FD audit completed; no new resource-management concerns introduced

No real broker credentials, tokens, OTPs, or network calls were used.

---
## Summary by cubic
Removes credential material from broker authentication and funds-related logs to close #1854.

**Bug Fixes**
- Authentication responses, tokens, secrets, OTP material, and response headers are no longer logged across the affected broker modules.
- Sensitive values are replaced with safe metadata such as presence flags, field names, response status, and exception type.
- Caller-visible error values and request/response handling are unchanged.
- Adds a static regression test that blocks reintroduction of the removed log patterns and runs in the CI-safe backend test selection.

<sup>Written for commit 7b73ca15cfb8d3aa4b95326b77114cfaee95ed5f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1916?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,6 +61,7 @@ jobs:
             test/test_multi_option_greeks_regression.py \
             test/test_history_format.py \
             test/test_precompress_assets.py \
+            test/test_broker_credential_logging.py \
             -v --timeout=60
 
   # Frontend linting
--- a/broker/dhan/api/auth_api.py
+++ b/broker/dhan/api/auth_api.py
@@ -37,11 +37,9 @@ def generate_consent(dhan_client_id):
         # Build URL with client_id parameter - REQUIRED by Dhan API
         url = f"{AUTH_BASE_URL}/app/generate-consent"
 
-        logger.info(f"Generating consent for Dhan Client ID: {dhan_client_id}")
+        logger.info("Generating consent for configured Dhan client")
         logger.info("Using configured broker API key")
-        logger.info(
-            f"Using API Secret: {BROKER_API_SECRET[:8] if BROKER_API_SECRET else 'None'}..."
-        )
+        logger.info("Using configured Dhan API secret")
 
         # Make the POST request with the client_id as a query parameter
         # The client_id parameter is REQUIRED for generate-consent
@@ -55,19 +53,25 @@ def generate_consent(dhan_client_id):
             data = response.json()
             if data.get("status") == "success":
                 consent_app_id = data.get("consentAppId")
-                logger.info(f"Consent generated successfully: {consent_app_id}")
+                logger.info(
+                    "Consent generated successfully; consent_id_present=%s",
+                    bool(consent_app_id),
+                )
                 return consent_app_id, None
             else:
                 error_msg = f"Failed to generate consent: {data}"
-                logger.error(error_msg)
+                logger.error(
+                    "Failed to generate consent; status=%s",
+                    data.get("status", "unknown"),
+                )
                 return None, error_msg
         else:
             error_msg = f"Failed to generate consent: HTTP {response.status_code} - {response.text}"
-            logger.error(error_msg)
+            logger.error("Failed to generate consent; HTTP status=%s", response.status_code)
             return None, error_msg
 
     except Exception as e:
-        logger.error(f"Exception in generate_consent: {str(e)}")
+        logger.error("Exception in generate_consent; type=%s", type(e).__name__)
         return None, f"An exception occurred: {str(e)}"
 
 
@@ -116,7 +120,10 @@ def consume_consent(token_id):
                     "token_expiry": data.get("expiryTime"),
                 }
                 logger.debug("Access token obtained")
-                logger.debug(f"Additional Data: {additional_data}")
+                logger.debug(
+                    "Additional authentication data fields: %s",
+                    list(additional_data.keys()),
+                )
                 return access_token, additional_data
             else:
                 return None, "Access token not found in response"
@@ -124,7 +131,7 @@ def consume_consent(token_id):
             return None, f"Failed to consume consent: {response.status_code}"
 
     except Exception as e:
-        logger.error(f"Exception in consume_consent: {str(e)}")
+        logger.error("Exception in consume_consent; type=%s", type(e).__name__)
         return None, f"An exception occurred: {str(e)}"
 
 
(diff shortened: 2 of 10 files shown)


marketcalls (approved):
Merging. The source changes are right and this lands cleanly on top of the three credential-logging commits from earlier today (`438af07f`, `c8dad476`, `fe0875f3`) rather than colliding with them, which I checked before anything else.

Verified rather than assumed:

- Merges clean onto current `main`, and clean against #1914 as well. You both edit the CI-safe list in `ci.yml` but insert at different points.
- No credential interpolation remains in any of the eight files. I grepped them for `access_token`, `auth_token`, `api_key`, `api_secret`, `refresh_token`, `clientcode`, `password` and `otp` inside log statements, excluding the safe `bool()` / `list()` / `len()` forms you introduced. Clean.
- `broker/kotak/database/master_contract_db.py` still logs `base_url_attempt`. I traced it: `base_url` is the third field of `auth_token.split(":::")`, so it is a URL and not a credential. Safe to log.
- 14 CI checks pass; the one pending is Snyk, and this PR changes no manifest.

The best thing in here is `str(e)` becoming `type(e).__name__`. That is not cosmetic. An exception message can carry a credential that never appears in the format string: `httpx.HTTPStatusError` embeds the full request URL, so a broker that puts its key in the URL path leaks on every 4xx, which is exactly the wrong-credential case you most want logged. `definedge/api/auth_api.py` had precisely that bug. Applying the pattern preemptively across these modules is the right instinct.

## One thing to improve, not blocking

The regression test is a static string blocklist. It reads each source file and asserts about thirty exact literals are absent. It never executes a log statement, and that limits it in three ways worth knowing:

1. **It cannot catch a new leak.** Only reintroduction of those exact strings. None of the roughly 35 leak sites fixed across `broker/` earlier today would have tripped it, because they were different strings.
2. **It is brittle to refactoring.** The literals are pinned including quoting and spacing. Rename a variable or reflow a line and the pattern stops matching, so the protection lapses silently while the test stays green.
3. **It fires on harmless text.** I appended one of the forbidden strings as a `#` comment and the test failed. So it blocks a comment while missing a real leak worded differently.

Green here means "these thirty strings are absent", not "no credentials are logged", and the second is what the issue actually asks for.

Compare #1914, which captures real log records with `caplog` and asserts the key is not in them, or `test/test_logging_redaction.py`, which runs the production `SENSITIVE_PATTERNS` against rendered messages. Both test behaviour, so both keep working after a refactor and can catch a leak nobody has seen yet.

Worth a follow-up: point this at behaviour instead. Render each log call with a marker value, run it through `SENSITIVE_PATTERNS`, and assert the marker does not survive. That version would have caught today's 35 sites, and it would not care how the line is written.

Thanks for rebasing onto current `main` before submitting, and for scoping it to the files the issue names.

PR #1931, opened 2026-08-28
fix(positions): stop overwriting broker-supplied P&L with the sandbox model

punksterlabs:
`useLivePrice` recomputes P&L on every tick as `(ltp - avg) * qty * lot_size`. That is the **sandbox's** model, and it is applied to live broker positions too.

A live broker's P&L comes from the exchange and can use entirely different accounting. Delta Exchange reports option unrealized P&L as the current *buyback cost* — it tracks `-(mark * contract_value)` and ignores entry price completely, so no `(ltp - avg)` formula reproduces it in magnitude **or in sign**.

## Measured on a live account

`contract_value = 0.001`:

| position | avg | mark | exchange reports | `-(mark × cv)` | `(ltp − avg) × qty × cv` |
| --- | --- | --- | --- | --- | --- |
| BTC call | 1104.00 | 1211.79 | **−1.21** | **−1.2118** ✓ | **+0.11** ✗ |
| BTC put | 1072.00 | 517.22 | −0.51 | −0.5172 ✓ | −0.55 |

The call is the clear case: **a losing position is rendered as a gain.** On the page this showed as −0.72 against the exchange's −6.86.

The API is already correct — `broker/deltaexchange/mapping/order_data.py` reads `realized_pnl` and `unrealized_pnl` verbatim. This is purely the display layer overwriting a good number.

## The fix

Make the recomputation opt-in via `recalculatePnl`, defaulting to `false`.

**Positions** passes analyzer mode only. Every broker adapter returns a real positions P&L (`flattrade`/`tradesmart` `round(total_pnl, 2)`, `motilal` `marktomarket`, `shoonya` `pnl`), so recomputing there overwrites the exchange's figure.

**Holdings is deliberately different**, and this is the part worth reviewing closely. Four adapters hard-code holdings `pnl`/`pnlpercent` to `0` and rely on the browser to fill them in — their own comments say so:

```
flattrade   0.0  # LTP is missing, so P&L is effectively 0 for now
motilal     0.0  # Would need current price to calculate P&L
shoonya     0.0  # Will be calculated when market price is available
tradesmart  0.0
```

Disabling recalculation for all live holdings would show `0.00` on every holding for those brokers, and refetching cannot fix it because the zero originates in the adapter. So Holdings also opts in for those, via a named set kept next to a comment pointing back at the mappings.

The same broker therefore recalculates on Holdings and not on Positions. That looks inconsistent and is not — the two books carry different data quality from the same adapter, and a later "consistency" cleanup would reintroduce the bug.

## Also

Stop synthesizing `pnlpercent = 0` when a source omits the field. `Positions.calculatePnlPercent` is designed to derive a percentage when it is absent, but a synthesized zero makes the placeholder look API-provided, so every nonzero Delta position rendered `+0.00%`. The field is now left undefined and only emitted when present.

## Verification

- `tsc --noEmit` clean.
- Frontend suite unchanged by this PR: **19 failing / 1044 passing on `main`, and the same 19 / 1044 with the change applied.** Those failures are pre-existing (`IndicatorPickerDialog` and three others) and unrelated.
- Behaviour verified against a live Delta account, numbers above.

## Note on history

This supersedes #1712, which I closed myself while reworking it — it was never reviewed. This version differs in two substantive ways that came out of downstream review: the placeholder-adapter handling for Holdings (without which four brokers regress), and a periodic broker-book refresh so a preserved broker P&L does not sit frozen while the page is open.

---
## Summary by cubic
Stops the display layer from overwriting broker-supplied P&L with the sandbox's `(ltp – avg) * qty * lot_size` formula, which can show a losing position as a gain (measured: a Delta BTC call rendered +0.11 against the exchange's –1.21). Recalculation is now opt-in via `recalculatePnl`, defaulting to `false`.

**Migration / Behavior**
- Positions recalculate only in analyzer mode; every live broker adapter returns a real P&L, so the exchange's figure is now preserved.
- Holdings still recalculate for `flattrade`, `motilal`, `shoonya`, and `tradesmart`, whose holdings adapters hard-code pnl to 0 and rely on the browser to derive it.
- `pnlpercent` is no longer synthesized as `0` when the source omits it, so missing values are not shown as authoritative `+0.00%`.

<sup>Written for commit 27f13fc2b5d452403fdc60ac22bbe57b57b2f366. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1931?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/src/hooks/useLivePrice.ts
+++ b/frontend/src/hooks/useLivePrice.ts
@@ -34,6 +34,14 @@ export interface UseLivePriceOptions {
   multiQuotesRefreshInterval?: number
   /** Pause WebSocket and polling when tab is hidden (default: true) */
   pauseWhenHidden?: boolean
+  /**
+   * Recompute P&L from the live tick instead of using the source's value.
+   * Only correct when the caller's P&L model matches this hook's, i.e.
+   * unrealized = (ltp - avg) * qty * lot_size. A live broker's P&L comes
+   * from the exchange and may use entirely different accounting, so this
+   * defaults to false. See the note at the recomputation below.
+   */
+  recalculatePnl?: boolean
   /** Time in ms to wait before pausing when hidden (default: 5000) */
   pauseDelay?: number
 }
@@ -86,6 +94,7 @@ export function useLivePrice<T extends PriceableItem>(
     useMultiQuotesFallback = true,
     multiQuotesRefreshInterval = 30000,
     pauseWhenHidden = true,
+    recalculatePnl = false,
   } = options
 
   const { apiKey } = useAuthStore()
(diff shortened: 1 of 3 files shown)


marketcalls (commented):
Review score: 76/100 at `27f13fc2b5d452403fdc60ac22bbe57b57b2f366`. Changes are needed before approval and merge.

- `frontend/src/pages/Positions.tsx:153`: Preserved live P&L stops updating while the page stays open. An open position continues receiving live LTP ticks while its P&L, totals and grouped P&L remain indefinitely at the last book fetch until an order, visibility transition or manual refresh. The same issue affects preserved holdings P&L. Add bounded visibility-aware position/holdings book refresh or a broker P&L stream when preserving broker values, and verify P&L changes without order events.

- `frontend/src/pages/Positions.tsx:153`: Some live adapters do not supply the complete P&L this change now requires. Tradejini open positions become zero P&L and Firstock positions lose their unrealized component. Refreshing those existing responses cannot fix the missing data. Normalize these broker adapters to complete total P&L before disabling recomputation, or retain explicit broker-aware fallback computation with the correct accounting model.

- `frontend/src/hooks/useLivePrice.ts:270`: Preserved broker string P&L reaches numeric reducers unchanged. Paytm/Firstock numeric strings concatenate in total/group sums (0+"12.34"+"5.00" becomes "012.345.00") and produce invalid or inflated totals. Coerce and validate preserved P&L to a finite number before returning enhanced items, and cover string-valued broker payloads in totals/grouping tests.

Issue #1735, opened 2026-08-03
[Bug] Smart orders starve regular orders in the strategy processor (time.sleep blocking)


vishalrateria:
Hey guys! I was profiling the strategy execution engine and noticed a pretty severe queue starvation issue in `blueprints/strategy.py`.

In the `process_orders()` thread, when a smart order is processed, it hits a hardcoded `time_module.sleep(1)` and then calls `continue`. Because of the `continue`, it skips the regular order block entirely until the smart order queue is 100% empty. If a user fires several smart orders rapidly, their regular orders get completely starved and delayed by several seconds.

**Proposed Fix:** Instead of blocking with `sleep(1)` and `continue`, we can just use a non-blocking timestamp check (e.g., `now - last_smart_order_ts >= 1.0`). If a second hasn't passed, we just skip the smart order block for that loop iteration and move straight to the regular orders.

Let me know if you agree and I'm happy to open a quick PR for this!

garudaccs:
Hi, I'd like to work on this.

Confirmed on current `main`: in `process_orders()` (`blueprints/strategy.py`), after handling one smart order the loop always does `time_module.sleep(1); continue`. The `continue` sends control straight back to the top of the loop, which checks `smart_order_queue` again before ever reaching the regular-order block below it. So if smart orders keep arriving faster than the 1/sec drain, regular orders don't just get delayed — they're starved for as long as the smart-order queue stays non-empty.

Planning to replace the unconditional sleep+continue with a timestamp-based rate check for smart orders, mirroring the `last_regular_orders` deque pattern already used a few lines below for regular orders — so every loop iteration gets a chance to reach the regular-order block regardless of smart-order volume. Will add a test that pushes both order types into the queues and asserts the regular order is serviced within a bound, failing on current `main` and passing after the fix.

Will open a PR shortly.

PR #1937, opened 2026-08-28
fix(chartink): stop smart orders from starving regular orders in the queue processor

garudaccs:
## What breaks for the user

A ChartInk strategy that fires orders in a burst (a scanner that pushes several symbols in the same tick) silently delays every *regular* order: the queue worker handles one smart order, sleeps a full second, and restarts the loop — so with a busy smart-order queue the regular-order block is never reached. Your regular orders sit behind an unbounded stream of smart orders. This is the same starvation shape as #1735, which was fixed in the strategy processor; `blueprints/chartink.py` is a separate implementation of the same loop and kept the bug.

## The cause

`process_orders()` in `blueprints/chartink.py` did:

```python
smart_order = smart_order_queue.get_nowait()
...place it...
time_module.sleep(1)
continue  # restarts the loop before the regular-order block
```

## The fix

Replace the blocking sleep with a timestamp gate (`last_smart_order_time`, ≥1s between smart orders) that falls through to the regular-order block **in the same iteration** — regular orders dispatch even while the smart-order queue is busy, and smart orders stay capped near one per second.

## Scope note (vs the original version of this PR)

The original push also patched the legacy strategy processor. That module has since been retired upstream (#58dc8adb5), so this PR now carries only the ChartInk fix, rebased onto current `main`. The shutdown-drain contract the review raised for the strategy module is moot here: nothing sends a shutdown sentinel into the ChartInk queues — the worker runs for the process lifetime.

## Verification

Focused suite (new regression tests — the first fails on `main`, both pass with the fix):

```
$ .venv/bin/python -m pytest test/test_chartink_order_processor_fairness.py -v
test_chartink_order_processor_fairness.py::test_regular_order_is_not_starved_by_a_smart_order_backlog PASSED
test_chartink_order_processor_fairness.py::test_smart_orders_still_capped_near_one_per_second PASSED
2 passed in 0.97s
```

`ruff check blueprints/chartink.py` reports the same 3 pre-existing findings as `main` (no new lint delta). Rebased head: 76ad11cf4.

### DIFF
--- a/blueprints/chartink.py
+++ b/blueprints/chartink.py
@@ -75,44 +75,55 @@ order_processor_lock = threading.Lock()
 # Rate limiting state for regular orders
 last_regular_orders = deque(maxlen=10)  # Track last 10 regular order timestamps
 
+# Rate limiting state for smart orders (1/sec, timestamp-gated rather than a
+# blocking sleep so a busy smart-order queue can never starve regular orders)
+last_smart_order_time = 0.0
+
 
 def process_orders():
     """Background task to process orders from both queues with rate limiting"""
-    global order_processor_running
+    global order_processor_running, last_smart_order_time
 
     while True:
         try:
-            # Process smart orders first (1 per second)
-            try:
-                smart_order = smart_order_queue.get_nowait()
-                if smart_order is None:  # Poison pill
-                    break
+            now = time()
 
+            # Process at most one smart order per second. This gate must not
+            # skip the regular-order block below when it fires (or doesn't) -
+            # a busy smart-order queue must never starve regular orders.
+            if now - last_smart_order_time >= 1:
                 try:
-                    response = requests.post(
-                        f"{BASE_URL}/api/v1/placesmartorder",
-                        json=smart_order["payload"],
-                        timeout=30,
-                    )
-                    if response.ok:
-                        logger.info(
-                            f"Smart order placed for {smart_order['payload']['symbol']} in strategy {smart_order['payload']['strategy']}"
-                        )
-                    else:
-                        logger.error(
-                            f"Error placing smart order for {smart_order['payload']['symbol']}: {response.text}"
-                        )
-                except Exception as e:
-                    logger.exception(f"Error placing smart order: {str(e)}")
+                    smart_order = smart_order_queue.get_nowait()
+                    if smart_order is None:  # Poison pill
+                        break
 
-                # Always wait 1 second after smart order
-                time_module.sleep(1)
-                continue  # Start next iteration
+                    try:
+                        response = requests.post(
+                            f"{BASE_URL}/api/v1/placesmartorder",
+                            json=smart_order["payload"],
+                            timeout=30,
+                        )
+                        if response.ok:
+                            logger.info(
+                                f"Smart order placed for {smart_order['payload']['symbol']} in strategy {smart_order['payload']['strategy']}"
+                            )
+                        else:
+                            logger.error(
+                                f"Error placing smart order for {smart_order['payload']['symbol']}: {response.text}"
+                            )
+                    except Exception as e:
+                        logger.exception(f"Error placing smart order: {str(e)}")
 
-            except queue.Empty:
-                pass  # No smart orders, continue to regular orders
+                    last_smart_order_time = now
+                except queue.Empty:
+                    pass  # No smart orders
 
             # Process regular orders (up to 10 per second)
+
+            # Re-take the clock here: the smart-order request above can block
+            # for seconds, and reusing that stale `now` would expire queued
+            # regular-order timestamps prematurely — a burst could then
+            # exceed the regular API's 10/sec limit.
             now = time()
 
             # Clean up old timestamps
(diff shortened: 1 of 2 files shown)


marketcalls (commented):
Review score: 80/100 at `efb655d4a3a3849a3170494c06a1a6e5844de7df`. Changes are needed before approval and merge.

- `blueprints/strategy.py:150`: The fairness loop consumes the regular-queue shutdown sentinel before draining queued smart orders. _shutdown_order_processor still puts None only on regular_order_queue; with five queued smart orders the parent sends all five, while this head sends one and leaves four unsent. Preserve a defined drain-on-shutdown contract across both queues while retaining fairness, add this shutdown regression case, then rebase the current merge conflict and rerun both processor suites.

Validation: Parent/head shutdown regression reproduced with synthetic queues and a mocked HTTP client; no network calls. Both processor diffs and their new tests reviewed. Current head conflicts with main.

garudaccs:
Thanks for the detailed review and the 80/100 pass — the shutdown finding was fair, and the resolution ended up coming from your side of the fence:

1. **Strategy module retired.** While preparing the rebase we found #58dc8adb5 ("refactor: retire the legacy /strategy module") removed `blueprints/strategy.py` entirely, so the half of this PR (and the drain-on-shutdown patch against it) no longer has a target. We've dropped that half rather than resurrect retired code.

2. **Shutdown contract.** For the remaining surface: graceful shutdown for the strategy processor had already landed in 2fb34f30a (atexit drain before the daemon thread is killed). `blueprints/chartink.py` — the file this PR now fixes — has no shutdown sentinel sender at all; its worker runs for the process lifetime, so the sentinel-before-drain hazard doesn't arise there. If you'd rather have an explicit drain there too, happy to follow up in a separate PR per your one-change-per-PR preference.

3. **Rebased + rescoped.** The PR now carries only the ChartInk starvation fix, rebased onto current `main` (head `76ad11cf4`): the blocking `sleep(1)`-then-`continue` is replaced with a timestamp gate that falls through to the regular-order block in the same iteration. New regression tests in `test/test_chartink_order_processor_fairness.py` — the starvation case fails on `main` and passes with the fix; smart orders stay capped near 1/sec.

```
$ .venv/bin/python -m pytest test/test_chartink_order_processor_fairness.py -v
...::test_regular_order_is_not_starved_by_a_smart_order_backlog PASSED
...::test_smart_orders_still_capped_near_one_per_second PASSED
2 passed in 0.97s
```

`ruff check` on the touched files shows the same 3 pre-existing `main` findings — no new lint delta. Ready for re-review.

marketcalls:
Re-reviewed at `76ad11cf41a908408c4b15fef84ffbc913b86e75`.

The Chartink starvation fix works: both new tests pass, and the starvation test fails against the parent. I also verified the scope change: the legacy `blueprints/strategy.py` was removed upstream, and Chartink has no production shutdown-sentinel sender. The previous shutdown objection therefore no longer applies to this revision. The rebase also resolves the merge conflict.

One new blocker remains before approval:

**[P1] Refresh the regular-order timestamp after the smart-order request.** In [`blueprints/chartink.py:124–128`](https://github.com/marketcalls/openalgo/blob/76ad11cf41a908408c4b15fef84ffbc913b86e75/blueprints/chartink.py#L124-L128), the regular-order block now reuses `now` captured before the blocking smart-order HTTP request. A slow smart response makes the first regular order's timestamp stale, so it expires from `last_regular_orders` too early and the worker can exceed the regular API's 10/sec limit. A rejected order is logged and removed from the queue without retry.

Reproduced with one smart request taking 1.2 seconds, followed by 11 regular orders, using real time and an in-process Flask-Limiter configured for a moving window of 10/sec:

| Version | Result |
| --- | --- |
| Parent | All 11 accepted, with the 11th correctly delayed |
| This head | 11 regular requests dispatched within about 60 ms; the 11th returns HTTP 429 and is not retried |
| This head with `now = time()` restored before the regular-order block | All 11 accepted, with the 11th correctly delayed |

Please restore the fresh timestamp immediately before regular-order rate-limit cleanup and add a regression covering slow smart responses plus a regular-order burst. The existing tests mock immediate responses and do not catch this case.

Also, the new fairness test file is absent from the explicit CI-safe pytest list in `.github/workflows/ci.yml`; please include it so CI runs these regressions.

Validation used isolated source copies, mocked order dispatch, and an in-process limiter. No broker orders or external HTTP calls were made.

garudaccs:
Rebased head is now `f406e59db` — the P1 is fixed.

**[P1] Stale clock before the regular-order rate-limit cleanup.** `now = time()` is re-taken immediately before the regular-order cleanup loop, so timestamps are never aged by a blocking smart-order request. `last_regular_orders.append(now)` in that block now records the actual dispatch time for the same reason.

New regression test `test_regular_burst_after_slow_smart_response_is_rate_limited` — one smart request mocked to take 1.2s, then an 11-order regular burst:

- fails on the parent head (all 11 dispatched in a ~60ms window),
- passes with the fix (the 11th is held for the next 1/sec window, none dropped).
Verified fail-before → pass-after by reverting the one-line fix locally; the two pre-existing fairness tests are untouched and green.

**CI list:** tried to add `test/test_chartink_order_processor_fairness.py` to the CI-safe pytest list in `.github/workflows/ci.yml`; our push token can't modify workflow files, so applying that hunk on your side:

```suggestion
          uv run pytest test/test_chartink_order_processor_fairness.py \
            test/test_log_location.py test/test_navigation_update.py \
```
(first line of the list in the `Run CI-safe tests` step)

PR #1949, opened 2026-08-31
Add daily iron condor example strategy with matching bhavcopy backtest

ayshvm:
## What this adds

Two self-contained scripts in `strategies/examples/`, following the existing
`simple_ema_strategy.py` pattern — no changes to core platform code.

### `iron_condor_daily.py`

An intraday iron condor on NIFTY / BANKNIFTY, uploadable via the Python Strategy Host (`/python`).

- Sells OTM4 CE/PE, hedges with OTM6 CE/PE (offsets configurable per instrument)
- **Sizes itself from a live `client.margin()` check** against a configurable `TOTAL_CAPITAL` / `CAPITAL_ALLOCATION_FRACTION`, rather than a hardcoded lot count — so the same file works on any account size, with `MAX_LOTS_PER_INSTRUMENT` as a hard ceiling
- Entry window 9:20–10:30; exits on profit target (50% of credit), stop loss (100% of credit), or forced square-off at 14:45
- One entry per instrument per day, with local state in `.iron_condor_daily_state.json` so a restart can't double-enter
- Uses `place_smart_order` / `quotes` / `optionchain` through the OpenAlgo SDK

**Safety gate:** defaults to `SANDBOX_MODE = True`. Going live requires *two* independent switches — flipping that flag **and** setting `LIVE_TRADING_CONFIRMED=YES` in the environment. The script refuses to start if only one is set.

No credentials are hardcoded. `OPENALGO_API_KEY`, `HOST_SERVER` and `WEBSOCKET_URL` are read from the environment, matching the convention in `simple_ema_strategy.py` and `strategies/README.md`.

### `backtest_iron_condor_daily.py`

An EOD backtest of the *same* configuration against real NSE F&O bhavcopy settlement data, downloaded from `nsearchives.nseindia.com`. Standard library only — no broker connection, no API key, no new dependencies. Results are written to `backtest_results.json`.

It deliberately models only what daily data can support: **enter at open, exit at close**. It does not synthesise intraday stop/target triggers from four independent daily H/L ranges. Days with unreliable prices (credit ≤ 0, credit > width, missing strike) are skipped and counted, never traded.

### `strategies/.gitignore`

Ignores the local run artifacts these two scripts produce (`.bhavcopy_cache/`, `backtest_results.json`, `.iron_condor_daily_state.json`) so they can't be committed as source.

## Results and caveats

Over roughly 60 trading days of bhavcopy data:

- **NIFTY** backtests reasonably — about an 81% win rate on gross open-to-close P&L
- **BANKNIFTY does not.** Since BANKNIFTY moved to monthly-only expiry, this "daily" setup usually holds options 2–4 weeks from expiry rather than near-dated ones. That produced both a net loss and a high rate (~40%) of unreliable / illiquid-strike days.

That caveat is written into the script header. This is offered as a *worked, documented* strategy to learn the Python Strategy Host and the option chain API from — **not** as a validated edge, and explicitly not validated for BANKNIFTY.

The backtest also reports **gross** P&L only; it does not model STT, brokerage, GST or slippage. For a high-win-rate options seller those are a material fraction of the gross number, so the win rate above should be read as an upper bound.

## Testing done

- Backtest run end-to-end against downloaded NSE bhavcopy for both underlyings (numbers above)
- Live script exercised in **sandbox mode only** — entry, profit-target, stop-loss and forced-exit paths all reached. Never run against real money.
- Verified the two-switch live gate blocks startup when either switch is unset
- Both files byte-compile clean; no core platform files touched

## Scope

Per the one-feature-per-PR policy in CONTRIBUTING.md, this PR is limited to the two example scripts and the `.gitignore` entries they need. Happy to adjust naming, defaults, or the docstrings to match house style.

---
## Summary by cubic
Adds two self-contained example scripts — an intraday iron condor for NIFTY/BANKNIFTY and an EOD backtest against NSE bhavcopy settlement data — with no core platform changes and run artifacts excluded via `strategies/.gitignore`.

**Strategy script**
- Sells OTM4 CE/PE and hedges with OTM6 CE/PE, entering once per instrument between 9:20–10:30 IST.
- Sizes each position from a live margin check against configurable capital rather than a fixed lot count.
- Exits on a 50%-of-credit profit target, 100%-of-credit stop-loss, or forced square-off at 14:45.
- Defaults to sandbox mode; live trading requires both `SANDBOX_MODE=False` and `LIVE_TRADING_CONFIRMED=YES`, otherwise the script refuses to start.

**Backtest and results**
- Models enter-at-open/exit-at-close only and skips days with unreliable prices rather than trading them.
- Shows about an 81% win rate on NIFTY but a net loss on BANKNIFTY, whose monthly-only expiry leaves far-dated and often illiquid strikes.
- Reports gross P&L only — no STT, brokerage, or slippage — so win rates are an upper bound.
- The live script was tested in sandbox only and never run against real money.

<sup>Written for commit 91047a5065be633dec9d0116cca5b6d41a56a63e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1949?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/strategies/.gitignore
+++ b/strategies/.gitignore
@@ -27,3 +27,8 @@ __pycache__/
 *.pyc
 *.pyo
 *.pyd
+
+# Local run artifacts from the example strategies (not source)
+examples/.bhavcopy_cache/
+examples/backtest_results.json
+examples/.iron_condor_daily_state.json
--- a/strategies/examples/backtest_iron_condor_daily.py
+++ b/strategies/examples/backtest_iron_condor_daily.py
@@ -1,0 +1,270 @@
+#!/usr/bin/env python
+"""
+EOD backtest for iron_condor_daily.py against real NSE settlement data.
+
+Pulls NSE's public daily F&O bhavcopy (one file per trading day, contains
+every strike/expiry's Open/High/Low/Close/Settlement for that day - this is
+NSE's own official record, not a broker feed) for NIFTY and BANKNIFTY index
+options over a date range, resolves the same OTM6-hedge / OTM4-short legs
+the live strategy uses, and sums up what selling that spread would have
+made or lost each day.
+
+HONESTY ABOUT WHAT THIS DOES AND DOESN'T PROVE:
+NSE's daily bhavcopy has one Open/High/Low/Close per contract per day - not
+a full intraday tick history. That means this backtest can faithfully
+answer "what if you entered at the day's open and held to the day's close"
+(no intraday exit at all), but it CANNOT faithfully simulate the live
+script's actual intraday profit-target / stop-loss logic - those trigger at
+a specific mark-to-market moment during the day, and reconstructing that
+from four independent daily High/Low ranges would silently assume all four
+legs hit their extreme at the same instant, which is not something the data
+supports. So this script reports two numbers, clearly separated:
+  1. Open-to-close P&L (the trustworthy one - directly backed by settlement
+     data)
+  2. A rough intraday sensitivity band using each day's High/Low (a loose
+     "how much did it move" indicator, NOT a stop-loss simulation)
+Treat (1) as the honest backtest and (2) as a hint, not a promise.
+
+Requires only the Python standard library (urllib, zipfile, csv) - no
+broker connection needed, this is pure historical NSE data.
+"""
+import csv
+import io
+import json
+import os
+import statistics
+import time
+import urllib.error
+import urllib.request
+import zipfile
+from datetime import date, timedelta
+
+# ---------------------------------------------------------------------------
+# Config - keep in sync with strategies/examples/iron_condor_daily.py
+# ---------------------------------------------------------------------------
+UNDERLYINGS = ["NIFTY", "BANKNIFTY"]
+HEDGE_OFFSET = 6   # strikes away from ATM for the long (insurance) legs
+SHORT_OFFSET = 4   # strikes away from ATM for the short (premium) legs
+PROFIT_TARGET_CREDIT_FRACTION = 0.5
+STOP_LOSS_CREDIT_MULTIPLE = 1.0
+
+START_DATE = date.today() - timedelta(days=90)
+END_DATE = date.today() - timedelta(days=1)
+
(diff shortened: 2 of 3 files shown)


marketcalls (commented):
Review score: 57/100 at `91047a5065be633dec9d0116cca5b6d41a56a63e`. Changes are needed before approval and merge.

- `strategies/examples/iron_condor_daily.py:307`: SANDBOX_MODE=True is only a label and bypasses the local live-confirmation check; it does not enforce sandbox routing. With server Analyzer mode off, the default script enters its order loop and can place live orders although the comments promise both modes must agree and the log says fake money. Fail closed unless the actual server mode matches the requested mode before entry, make live confirmation effective for every live placement, and keep exit routing bound to the position entry mode.

- `strategies/examples/iron_condor_daily.py:294`: Every exit branch marks the record closed without validating the basket exit outcome. A rejected exit leaves real exposure open while subsequent loops stop managing it. Inspect per-leg exit results, reconcile fills/remaining exposure, retain management on rejected or partial exits, and persist closed only after confirmed closure; cover rejection, partial fill and lost response cases.

- `strategies/examples/iron_condor_daily.py:222`: Entry trusts aggregate optionsmultiorder success without reconciling individual legs. That service can return overall success with failed leg results, so the script can call a partially entered spread open without confirming its hedges and fills. Validate every leg acceptance and fill, record actual quantities/product, and reconcile or unwind partial entry before treating the spread as fully established. Cover one rejected hedge and a partial short fill.

Validation: Inspected both new examples. Synthetic probes confirmed the ineffective sandbox gate and rejected-exit closure; no broker calls or orders. Independent main/run/entry probe confirmed an ordinary four-leg request with sandbox flag true and no live confirmation; REST/service routing was traced through the global Analyzer branch.

PR #1962, opened 2026-09-01
feat(trading): add TradingView-style column sorting to the watchlist

aakash-code:
## Summary
- Every watchlist column header (Symbol, Last, Chg, Chg%, Vol, High, Low, Open) is now clickable and cycles ascending -> descending -> the list's own drag order, matching TradingView's watchlist sort behaviour.
- A row with no quote yet, or no value for the sorted column, sorts to the bottom regardless of direction rather than being treated as zero (which would otherwise park it among the biggest losers).
- Drag-to-reorder is disabled while a column sort is active, since the on-screen order no longer matches the list's underlying drag order; clicking the active header a third time clears the sort and restores manual dragging.
- The sort choice persists per browser (`oa-trading-watchlist-sort` in localStorage) alongside the existing column/display preferences.

## Test plan
- [x] `npx tsc --noEmit` passes
- [x] `npm run build` succeeds
- [x] `npx biome check` shows no new errors (2 pre-existing errors/warnings on this file are unchanged from main)
- [ ] Manually verified in a running instance: click each header, confirm asc/desc/clear cycle and that drag reordering re-enables once cleared

---
## Summary by cubic
Adds TradingView-style column sorting to the watchlist so every column header cycles ascending → descending → back to the user's own drag order. Previously headers were static and the list could only be reordered by dragging rows.

- Rows with no quote yet, or no value for the sorted column, sort to the bottom regardless of direction instead of being treated as zero.
- Drag reordering is disabled while a sort is active, since the on-screen order no longer matches the list's underlying drag order; clicking the active header a third time clears the sort and restores dragging.
- The sort choice persists per browser in `oa-trading-watchlist-sort` in localStorage; a stale or malformed saved sort is ignored, and hiding the sorted column clears the sort.

<sup>Written for commit b6f5fefaafdd4907b897aa4b8a9090dcce27f186. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1962?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/src/components/trading/WatchlistPanel.tsx
+++ b/frontend/src/components/trading/WatchlistPanel.tsx
@@ -13,6 +13,8 @@
  */
 
 import {
+  ArrowDown,
+  ArrowUp,
   ChevronDown,
   GripVertical,
   MoreHorizontal,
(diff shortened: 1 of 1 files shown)


marketcalls (commented):
Review score: 87/100. Reviewed head: b6f5fefaafdd4907b897aa4b8a9090dcce27f186.

Please replace the newly added ArrowUp/ArrowDown direction icons at `frontend/src/components/trading/WatchlistPanel.tsx:926` and the numeric column header with plain-text sort state, as required by current `CLAUDE.md`. Expose Ascending/Descending in the accessible label or state as well; the current decorative arrows leave screen-reader users without the current sort direction. Preserve the ascending/descending/manual cycle, unknown values last and disabled dragging during sorting. Add focused verification of the displayed/accessibility state and restoring manual order. The comparator checks passed; this is a standards/accessibility correction, not a security vulnerability.

PR #1976, opened 2026-09-02
fix(strategy): cash and equity leg parity, and three defects that stranded runs

marketcalls:
Cash and equity legs were the weak segment of `/strategy`. Derivatives resolve against the master contract, count in lots and carry a venue-correct product; cash reached several of those paths unchecked, and the controls that would have caught it lived only in the browser. Chasing that turned up three unrelated defects in the same area, each of which could strand a run.

Everything here was verified against a running instance in sandbox on both NSE and BSE, not only against tests.

## Cash and equity parity

**Server-side validation.** `universe_tab` was free text up to thirty characters, and every rule hanging off it lived in the frontend, so a cash leg on an index tab validated here and failed at run start with `No cash contract found for NIFTY on NSE`. It is now one of four values, and a leg's segment is checked against the tab that offers it. A caller that omits the tab has it derived from its own legs rather than defaulted, so an API caller is never refused about a field it never set.

A signal leg's segment and its exchange had to agree with nothing: a leg marked cash on NFO validated, the segment was then ignored downstream, and the leg traded whatever the symbol happened to be.

**Cash cannot be carried short.** The product is read as intent, so anything that is not `MIS` reaches a cash venue as `CNC`, which makes a short a naked short delivery. A batch leg with `position: "S"` is refused at save; a signal leg is refused at signal time, where the side being opened is known rather than the sides it accepts. Neither product had any control for this.

**A signal cash symbol is now checked against the master contract.** The check was guarded on a derivative exchange, which left a misspelled equity as the one instrument nothing verified: a cash leg is not resolved from an underlying either, so `RELAINCE` on NSE reached the broker verbatim while batch mode refused the identical typo.

**`strategy_kind` left `UPDATABLE_FIELDS`.** The two kinds do not share a leg shape, so a PATCH that flipped it left every stored leg describing the other kind's contract. It stays settable at create, and a request to change it is now refused by name instead of silently dropped.

## Three defects found while testing

**A startup prune held the SQLite write lock for the life of the process.** `_prune_old_tags` committed only when it had deleted something. A DELETE takes the write transaction whether or not it matches a row, so the ordinary case, a prune with nothing old enough to remove, left that transaction open. It runs from `init_strategy_book_db()` at startup, which has no Flask app context, so the teardown that would have released the session never fired. Measured on a real instance: the lock was held 15.5 seconds and counting from `DELETE FROM strategy_order_tags WHERE created_at < ?` against a table holding one row. Every other writer on `openalgo.db` failed with `database is locked`, strategy recovery among them, which is why a finished run stayed open across every restart with its strategy stuck reading `running`. Both prunes now commit unconditionally and log conditionally, which is the shape `strategy_module_db._prune_webhook_events` and `traffic_db` already use.

**Pressing Start on a signal strategy stranded it.** `start_run` is the batch lifecycle and had no kind guard, so it ran over signal legs and raised `ValueError: Leg 1 has an unusable position: ''`. The failure handler then could not clean up: it looks for live state to decide whether anything is held, and `init_run_state` was the call that raised. The run was left open and the strategy claimed. `start.md` already documented the endpoint as batch-only; it is now enforced, and a start that dispatched nothing finalises its run whether or not run state exists, since `dispatch_attempted` is written immediately before each dispatch and an empty set is therefore proof of flatness.

**Recovery read through an exit row that was not there.** `exit_filled` has a clause for a leg with no order rows at all, where the checkpoint is the only witness, but two reads went through the missing row regardless and raised `AttributeError`. Contained per run, so nothing was corrupted, but a run that could have been rebuilt was finalised as `recovery_failed` instead.

## Operator surfaces

- A start whose entries were all rejected now names the run and the venue's own reason. It previously answered `Every entry order was rejected` with no run id and no cause, while the reason sat on the order rows the whole time.
- `risk_unit` was never sent. `legToPayload` copied `sl_pts`, `target_pts` and `trail` and dropped the unit, so a leg configured as a percentage of entry saved as points, and editing it converted it back.
- The wizard wrote `webhook_ip_allowlist: null` into every create and every patch, destroying an allowlist configured out of band on the first save.
- A batch cash leg is labelled and capped in shares. The derivative cap of 50 made fifty shares the largest cash order a batch strategy could place, which the server had already fixed and the form had not.
- The stocks tab trades any listed equity on NSE or BSE. The underlying search ran against the derivative venue, so it queried NFO and only ever suggested stocks with an F&O contract: searching `ANSAL` returns nothing there, four rows on NSE and six on BSE. The exchange was also pinned to NSE though the server has always accepted BSE.
- A cash leg is counted in shares rather than units, and switching the toggle converts: one lot of RELIANCE becomes 500, not 1. In units mode on a derivative the spinner steps by the lot size, so 500 goes to 1000 rather than 501.
- The Start button is hidden for a signal strategy.

## Migration

`universe_tab` is normalised on existing rows so the new enum and segment rules cannot lock an operator out of editing a strategy that saved cleanly under the old ones. Nothing about what a strategy trades is changed: a row whose tab is already valid for its own legs is left alone, and a row that has to move has its tab derived from its own legs rather than defaulted. It is idempotent, supports `--status`, and was tested against a copy of a real database seeded with the bad rows it exists to fix.

The code does not require it to have run. The validator derives the tab from the legs when one is absent, so an installation that pulls this and never runs `migrate_all.py` is not broken.

## Documentation

Twenty errors corrected across `docs/api/strategy-services/`, six introduced by this branch and the rest pre-existing drift found while checking: `risk_unit` documented as a strategy field when it is a leg field, the scheduler payload attributed to `run_started` which carries none, the webhook said to refuse `"LIVE"` when it lower-cases and accepts it, and `rejected_engine_error` mapped to 500 for signal mode when an engine refusal there is 400. All fourteen documented vocabularies were then checked against their code tuples programmatically, and every request field on all nine API-key schemas against its endpoint table.

The strategy reference claimed batch cash counts the configured number as units. It multiplies by the master-contract lot size on every segment and only equals a share count because a cash row's lot size is 1. Corrected.

## Testing

- 3929 passing. The 19 failures and 3 errors are pre-existing and fail identically on `main` (mstock, tradesmart rate limiter, flow custom legs, pending-order GTT, telegram, python editor, rate-limit mocks).
- 186 frontend tests, `tsc -b` clean, biome clean, production build clean.
- New tests: 18 for the cash rules, 3 pinning the risk-unit round trip, 4 for the prune lock, 3 for the signal start, 2 for the recovery guard, 7 for the quantity conversion. The lock, start and recovery tests each fail with their fix reverted, so none can pass vacuously.
- `test_strategy_module_signals` now seeds the cash contracts it trades. It passed only while the shared symbol table was empty, and `test/sandbox/conftest.py` seeds NSE rows at session scope, so the master-contract change would otherwise have introduced an ordering-dependent flake.
- Verified live in sandbox through the public webhook on both venues: entry accepted, filled 6 @ 1313.10 on NSE and 6 @ 1313.15 on BSE, exit accepted, filled SELL 6 both sides. Plus 35 equity scenarios covering the cash rules, the quantity cap, segment and venue agreement, percent risk, and product translation sending CNC to NSE and NRML to NFO in one basket.

## Deliberately not done

`close_all` is an alias for stop, differing only in its audit event, so a second button beside "Stop and Close Positions" would tell the operator the run continues when it does not.

Two direction-filter improvements are left open for a decision: failing closed on an unrecognised direction, where the reference implementation is stricter, and giving the direction refusal its own 403 label instead of sharing `rejected_invalid_action` with two other causes.

https://claude.ai/code/session_0116uGaZ7nQmoMVGwGWM9eUy

---
## Summary by cubic
Brings cash and equity legs in `/strategy` to parity with derivatives: server-side validation, no short cash positions, and three defects fixed that could strand a run. A migration normalizes existing `universe_tab` values so the new enum does not lock operators out of editing old strategies; it is not required to run, since the validator derives a missing tab from the legs.

**Cash and equity parity**
- `universe_tab` is now one of four values; a leg's segment is checked against the tab that offers it, and a missing tab is derived from the validated legs.
- A signal leg's segment and exchange must agree, and cash symbols are checked against the master contract like derivatives.
- Cash cannot be carried short: a batch leg with `position: "S"` is refused at save, a signal leg at signal time, before anything is squared, so a refused short never liquidates a position.
- `strategy_kind` can no longer be changed via PATCH; a change is refused by name and a no-op dropped.

**Defects fixed**
- The startup prune in `strategy_book_db` committed only when it deleted something, leaving SQLite's write lock held for the process lifetime; it now commits unconditionally.
- `start_run` refuses signal strategies by name instead of stranding a run; a start that dispatched nothing finalises its run whether or not run state exists.
- Recovery no longer reads through a missing exit order row; it falls back to the checkpoint, and the tests now fail with the fix reverted.
- Frontend: `risk_unit` is sent, the wizard no longer writes `webhook_ip_allowlist: null`, cash quantity is capped in shares, percent trails are converted and labelled by unit, and the stocks tab trades any listed NSE or BSE equity with an explicit BSE choice sticking.

<sup>Written for commit 4c5bbdd7c9b5bf2e798c398b65aff1c0cc20316b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1976?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/blueprints/strategy_module.py
+++ b/blueprints/strategy_module.py
@@ -867,11 +867,17 @@ def _validate_strategy_config(payload: Any) -> dict:
     raw = _mapping(payload, "The request body")
     _reject_unknown(raw, CONFIG_FIELDS, "The request")
 
+    # Legs first, because the universe tab is derived from them when the caller
+    # names none, and that derivation has to read the segment a leg will
+    # actually carry. Reading the raw payload saw no segment on a signal leg
+    # that relied on the default, derived a tab for a derivative universe, and
+    # then refused the very cash leg the default had just produced.
+    kind = _choice(raw.get("strategy_kind") or "batch", store.STRATEGY_KINDS, "strategy_kind")
+    legs = _validate_legs(_required(raw, "legs"), kind)
+
     config: dict[str, Any] = {
         "name": _text(_required(raw, "name"), "name", max_length=MAX_NAME_LENGTH),
-        "strategy_kind": _choice(
-            raw.get("strategy_kind") or "batch", store.STRATEGY_KINDS, "strategy_kind"
-        ),
+        "strategy_kind": kind,
         # Only read for signal strategies; harmless and inert on a batch one.
         "direction": _choice(raw.get("direction") or "both", store.DIRECTIONS, "direction"),
         # Derived from the legs when the caller does not say, because the tab
@@ -880,7 +886,7 @@ def _validate_strategy_config(payload: Any) -> dict:
         # cash leg underneath it would be a refusal about a field the caller
         # never set. An explicit tab is still checked, and still governs.
         "universe_tab": _choice(
-            raw.get("universe_tab") or _tab_for_legs(raw), UNIVERSE_TABS, "universe_tab"
+            raw.get("universe_tab") or _tab_for_legs(raw, legs), UNIVERSE_TABS, "universe_tab"
         ),
         "underlying": _text(
             _required(raw, "underlying"), "underlying", max_length=MAX_UNDERLYING_LENGTH
@@ -893,10 +899,7 @@ def _validate_strategy_config(payload: Any) -> dict:
         ),
         "product": _choice(raw.get("product") or "NRML", PRODUCTS, "product"),
         "pricetype": _choice(raw.get("pricetype") or "MARKET", PRICETYPES, "pricetype"),
-        "legs": _validate_legs(
-            _required(raw, "legs"),
-            _choice(raw.get("strategy_kind") or "batch", store.STRATEGY_KINDS, "strategy_kind"),
-        ),
+        "legs": legs,
         "overall_sl_mtm": _loss_amount(raw.get("overall_sl_mtm"), "overall_sl_mtm"),
         "overall_target_mtm": _gain_amount(raw.get("overall_target_mtm"), "overall_target_mtm"),
         "lock_profit": _validate_lock_profit(raw.get("lock_profit")),
@@ -936,17 +939,17 @@ _COMMODITY_EXCHANGES = frozenset({"MCX", "NCDEX", "NCO"})
 _WEEKLY_RANKS = frozenset({"weekly", "next_week"})
 
 
-def _tab_for_legs(raw: Any) -> str:
+def _tab_for_legs(raw: Any, legs: list[dict[str, Any]]) -> str:
     """The tab a configuration belongs to, read off the configuration itself.
 
-    Used only when the caller did not name one. Kept in step with the same
-    derivation in upgrade/migrate_strategy_universe_tab.py, which normalizes
-    rows written before the tab was validated.
+    Used only when the caller did not name one. Takes the *validated* legs
+    rather than the raw ones: a signal leg's segment defaults to cash, and
+    deriving from the raw payload missed that default, picked a derivative
+    universe, and then refused the cash leg validation had just produced. Kept
+    in step with the same derivation in
+    upgrade/migrate_strategy_universe_tab.py, which normalizes rows written
+    before the tab was validated.
     """
-    legs = raw.get("legs")
-    if not isinstance(legs, list):
-        return "weekly_monthly"
-
     segments = {str(leg.get("segment") or "").lower() for leg in legs if isinstance(leg, dict)}
     if "cash" in segments:
         return "stocks_fno"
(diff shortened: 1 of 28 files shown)


marketcalls:
All seven cubic findings and both Codex findings were valid and are fixed in `4c5bbdd7c`. The two Codex items duplicate cubic's on the exchange precedence and the percent trail.

**P1, and the one that mattered most.** The uncarryable-short check ran *after* the flip, so a `short_entry` on a cash leg held long under a carry product squared that long and only then refused the short. A signal that was never going to open anything closed a position instead. The check now runs before anything is squared. There is a test for it that fails against the previous ordering.

**`strategy_kind` on PATCH.** It stays in `CONFIG_FIELDS` because the merge seeds itself from that set, and a signal strategy whose kind was dropped would have its legs re-validated as batch legs and fail on its own stored configuration. The route now refuses a change by name before the merge and drops a no-op, so a changed kind no longer surfaces as a leg-shape error naming a field the caller never touched.

**Tab derived before the leg defaults.** Correct, and reachable exactly as described: a signal payload naming neither the tab nor a leg segment derived a derivative universe from the raw legs, then refused the cash leg that validation had just produced. The derivation now takes the validated legs. Verified on all three shapes: signal with no segment gives `stocks_fno`, signal futures gives `monthly_only`, batch cash gives `stocks_fno`.

**Seeded BSE never stuck.** Right, and most stocks are in the seed list, so BSE could be selected and never submitted. The explicit choice now wins on an open universe; a closed one still follows the index that was picked, which is the case with nothing to choose.

**Percent trail.** Both numbers are converted against the leg's entry now, and neither is shown before a fill supplies one, which is the rule `risk_adapter` already applies. The arming price was also adding a percentage to a price, which predates this branch but is fixed here since the label made it visible.

**The vacuous recovery test.** Entirely right: the entry row made `identity` non-None, `exit_filled` stayed false, and the exit-price read was never reached, so it passed against the unguarded code it existed to pin. It now drops the entry so the checkpoint is the only witness, and asserts the price and the kind come back from it. Reverted against the pre-fix code the two tests fail at different lines, 995 and 1018, which are the two reads.

**The app-context comment.** Correct, and worth having flagged. `_init_databases_and_schedulers` wraps the whole of startup in `app.app_context()`. The effect was real but my reason was wrong: that single context spans every startup step, so teardown does not fire until they have all finished, which is why the lock outlived the steps that still had writing to do. The comment and the test docstring now say that.

3930 Python tests passing, with the same 19 pre-existing failures that fail identically on `main`. 186 frontend tests, typecheck, lint and build clean.

PR #1979, opened 2026-09-02
fix(python-strategy): stop waiting for a dying process under the global lock

### LINKED ISSUE #1737: [Bug] Stopping a strategy freezes the entire web server for 5+ seconds (Global Lock contention)
Hey! I found a massive bottleneck when stopping strategies in `blueprints/python_strategy.py`.

The `stop_strategy_process()` function acquires the global `PROCESS_LOCK`, and while holding it, calls `process.wait(timeout=5)`. Because this lock is held by the main application thread during that 5-7 second wait, the entire web server and API freeze for all other requests until the subprocess finally dies.

**Proposed Fix:** We should send the `process.terminate()` signal inside the lock, but move the blocking `process.wait()` loops *outside* the `with PROCESS_LOCK:` block so they don't halt the rest of the application.

Let me know if you want me to drop a PR for this!

marketcalls:
Stopping a strategy froze the whole app. `stop_strategy_process` held `PROCESS_LOCK` across `process.wait(timeout=5)`, and that one line carried two separate faults.

## The half the report did not catch

`Popen.wait(timeout=...)` blocks inside C, `waitpid` on Linux and `WaitForSingleObject` on Windows. Neither is a yield point, so under gunicorn-eventlet, where one OS thread runs every greenlet, the call stops the entire worker for the length of the timeout rather than just the caller.

That matters for the proposed fix. Moving the wait outside the lock, as #1737 suggested, removes the contention but leaves the freeze, because the hub is stopped by the C call itself. It would look correct on `uv run app.py`, which never monkey-patches, and stay broken in production.

The pattern was already solved a hundred lines below: the `psutil.Process` branch of this same function uses `wait_for_psutil_process_exit`, which polls, and carries a comment explaining that psutil's pidfd path needs `select.poll()` that eventlet removes. The `subprocess.Popen` branch beside it never got the same treatment.

## The half it did catch

`PROCESS_LOCK` is process-wide and every start, stop, restart and status read takes it, so holding it for the death of a subprocess serialised all of them behind a wait that lasts as long as the strategy takes to notice its signal.

## Fix

- `wait_for_popen_exit` polls `poll()` against a deadline, mirroring `wait_for_psutil_process_exit`. `poll()` is non-blocking and reaps the child, so the hub keeps running between checks.
- `terminate_popen_safely` escalates exactly as the inline code did: graceful signal first, forced kill only if the process outlives it. Windows goes `terminate()` then `taskkill /F /T`; Unix signals the process group then `SIGKILL`, with the same direct-signal fallback when the process is not in its own group.
- The lock is now held to claim the strategy and, afterwards, to write the bookkeeping back. The wait sits between the two, outside it.

Two consequences worth calling out, both of which the previous shape got wrong:

- **Claiming pops the entry**, so a second stop arriving mid-wait is told the strategy is not running instead of signalling the same PID again.
- **A failed termination puts the entry back.** The process is still alive, so dropping it would leave nothing tracking it and no way to stop it again.

## Testing

`test/test_python_strategy_stop_lock.py`, 7 tests.

Two assert the defects themselves and **fail on the previous code**, so the file cannot pass vacuously:

```
test_wait_never_calls_the_blocking_popen_wait
  AssertionError: Popen.wait was called with [5]; it blocks the eventlet hub
  and must be replaced by a cooperative poll

test_lock_is_free_while_the_process_is_dying
  AssertionError: PROCESS_LOCK was held for the whole termination; every other
  start, stop and status read blocks behind it
```

The fake `Popen` reproduces the real blocking `wait()` rather than stubbing it out, which is what lets the second test demonstrate contention instead of merely showing that nothing waited.

The rest cover the immortal-process path, concurrent stops, both poll outcomes, and one end-to-end test against a real spawned subprocess.

```
uv run pytest test/test_python_strategy_stop_lock.py -v          # 7 passed
uv run pytest test/ -k "strategy or python_strat or eventlet or lock"
# 1345 passed, 3 skipped, 1 xfailed
# 1 failure, test_telegram_startup.py::test_bot_starts_and_stops_cleanly_in_eventlet_env,
# which fails identically on clean HEAD and is unrelated
uv run ruff check blueprints/python_strategy.py test/test_python_strategy_stop_lock.py
# All checks passed
```

`ruff format` reports the file would be reformatted, but that drift pre-exists on `main` and is elsewhere in the file. Verified that no formatting change is wanted on any line this PR touches.

Developed and tested on Windows, so the Windows branch is exercised directly, including the real-subprocess test. The Unix branch keeps the original signal sequence unchanged and is covered through the fake.

Closes #1737

https://claude.ai/code/session_0116uGaZ7nQmoMVGwGWM9eUy

---
## Summary by cubic
Stopping a Python strategy no longer freezes the app: the old blocking `Popen.wait()` held `PROCESS_LOCK` until termination, while the new path polls cooperatively outside the lock. Failed stops leave the process tracked instead of reporting success or clearing its configuration.

- Prevents concurrent stops from signaling the same PID and starts from launching replacements during termination.
- Uses cooperative polling for Windows `taskkill` escalation.
- Refuses deletion with HTTP 409 when the strategy cannot be stopped.
- Releases `PROCESS_LOCK` before stopping during deletion and shutdown cleanup.
- Broadcasts the stopped status before releasing the stop claim.
- Adds regression coverage for lock contention, termination races, failure paths, and real subprocess cleanup.
- Fixes #1737.

<sup>Written for commit f8cd7019e454289a90ff4ff27cfe8a32474cd81f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1979?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/blueprints/python_strategy.py
+++ b/blueprints/python_strategy.py
@@ -603,113 +603,107 @@ def start_strategy_process(strategy_id):
 
 
 def stop_strategy_process(strategy_id):
-    """Stop a running strategy process - cross-platform implementation"""
-    with PROCESS_LOCK:  # Thread-safe operation
-        if strategy_id not in RUNNING_STRATEGIES:
-            # Check if process is still running by PID
-            if strategy_id in STRATEGY_CONFIGS:
-                pid = STRATEGY_CONFIGS[strategy_id].get("pid")
-                if pid and check_process_status(pid):
-                    try:
-                        terminate_process_cross_platform(pid)
-                        STRATEGY_CONFIGS[strategy_id]["is_running"] = False
-                        STRATEGY_CONFIGS[strategy_id]["pid"] = None
-                        STRATEGY_CONFIGS[strategy_id]["last_stopped"] = get_ist_time().isoformat()
-                        save_configs()
-                        return True, "Strategy stopped"
-                    except Exception:
-                        pass
-            return False, "Strategy not running"
+    """Stop a running strategy process - cross-platform implementation.
 
+    Waiting for the process to die happens **outside** PROCESS_LOCK. The lock
+    is held only long enough to claim the strategy and, afterwards, to write
+    the bookkeeping back. Terminating a strategy takes as long as the strategy
+    takes to notice its signal, seconds if it is mid-request; holding a
+    process-wide lock across that stalls every other caller of it, which is
+    every start, stop, restart and status read on the page.
+
+    Claiming means popping the entry, so a second stop arriving during the wait
+    is told the strategy is not running rather than sending a second signal to
+    the same PID. If termination fails the entry goes back, because the process
+    is still alive and something has to still be tracking it.
+    """
+    # --- Claim, under the lock -------------------------------------------------
+    with PROCESS_LOCK:
+        strategy_info = RUNNING_STRATEGIES.pop(strategy_id, None)
+        orphan_pid = None
+
+        if strategy_info is None:
+            # Not tracked: it may still be alive from a previous app run.
+            config = STRATEGY_CONFIGS.get(strategy_id)
+            if config:
+                candidate = config.get("pid")
+                if candidate and check_process_status(candidate):
+                    orphan_pid = candidate
+            if orphan_pid is None:
+                return False, "Strategy not running"
+
+    # --- Terminate and wait, outside the lock ---------------------------------
+    if orphan_pid is not None:
         try:
-            strategy_info = RUNNING_STRATEGIES[strategy_id]
-            process = strategy_info["process"]
-            pid = strategy_info["pid"]
-
-            # Handle different process types
-            if isinstance(process, subprocess.Popen):
-                # For subprocess.Popen objects
-                # Platform-specific termination
-                if IS_WINDOWS:
-                    # Windows: Use terminate() then kill() if needed
-                    try:
-                        process.terminate()
-                        process.wait(timeout=5)
-                    except subprocess.TimeoutExpired:
-                        # Force kill using taskkill
-                        subprocess.run(
-                            ["taskkill", "/F", "/T", "/PID", str(pid)],
-                            capture_output=True,
-                            check=False,
-                        )
-                        process.wait(timeout=2)
-                else:
-                    # Unix-like systems (Linux, macOS)
-                    try:
-                        # Try to kill process group if it exists
-                        try:
-                            # Try SIGTERM first (graceful shutdown)
-                            os.killpg(os.getpgid(pid), signal.SIGTERM)
-                            process.wait(timeout=5)
-                        except OSError:
-                            # Process might not be in a process group, kill it directly
-                            process.terminate()
-                            process.wait(timeout=5)
-                    except (subprocess.TimeoutExpired, ProcessLookupError):
-                        try:
-                            # Force kill with SIGKILL
-                            try:
-                                os.killpg(os.getpgid(pid), signal.SIGKILL)
-                            except OSError:
-                                # Process might not be in a process group, kill it directly
-                                process.kill()
-                            process.wait(timeout=2)
-                        except ProcessLookupError:
-                            pass  # Process already dead
-            elif hasattr(process, "terminate"):
-                # Restored strategies are tracked as psutil.Process objects.
-                # Do not call psutil.Process.wait(timeout): under
-                # gunicorn-eventlet on Linux, psutil's pidfd wait path needs
-                # select.poll(), which eventlet removes from the patched
-                # select module.
-                if not terminate_psutil_process_safely(process, terminate_timeout=5, kill_timeout=2):
-                    return False, f"Failed to stop strategy PID {pid}"
-            else:
-                # Fallback: use PID directly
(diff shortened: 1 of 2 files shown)


marketcalls:
All six cubic findings and the Codex finding were valid, and fixed in `1d904324a`. Two of them were defects this PR introduced, which is the important part.

**The claim opened a start/stop race (Codex P1, cubic P2).** Popping the entry left the strategy in neither registry for the length of the wait. A start arriving in that window saw the id as absent and launched a replacement, and the finishing stop then wrote `is_running=False` and `pid=None` over the new process's config, leaving a live trading process the normal stop route would no longer even attempt to terminate. A second stop had the mirror problem: the config still carried a live PID, so it fell through to the orphan branch and signalled the same process again.

`STOPPING_STRATEGIES` now marks that window. Start is refused while it is set, a second stop is told the strategy is already stopping, and a `finally` releases it on every path including both early returns, or the strategy could never be started or stopped again for the life of the process.

**Two callers held the lock across the call (cubic P1).** `delete_strategy` and `cleanup_on_exit`. Reentrant, so no deadlock, which is exactly why it was easy to miss: it silently reinstated the stall this PR exists to remove. Both now decide under the lock and stop outside it. Good catch, this one defeated the fix for two of the six call sites.

**Orphan stop reported success unconditionally (cubic P1).** `terminate_process_cross_platform` swallows its own errors, so a clean return is not evidence the process died. The PID is checked before the config is touched. The PID fallback in the tracked path made the same assumption and is checked too.

**Windows taskkill blocked (cubic P2).** `subprocess.run` waits synchronously, the same problem one level down. Spawned and polled cooperatively now, and reaped either way.

**The real-subprocess test could have killed the test run (cubic P1).** Correct, and it would only ever have fired on Linux, which is why it was invisible here. Without `start_new_session`, the child shares the pytest runner's process group and `killpg` sends SIGTERM then SIGKILL to the whole run. Fixed, gated on non-Windows since the flag is meaningless there.

**The duplicate-stop test was weak (cubic P2).** Also right: it carried a `stop_finished` guard it never set, and did not reach the orphan path it was supposed to cover. Rewritten to make the config's PID look alive so the orphan branch is genuinely reachable, and it now asserts both that the second caller is refused and that the orphan path was never entered.

Three of the new tests fail against `7cc3f2a49`, so they pin the defects rather than merely describing them:

```
test_a_start_during_the_wait_is_refused
test_a_second_stop_during_the_wait_does_not_signal_twice
test_an_orphan_that_survives_termination_is_not_reported_stopped
```

A fourth asserts the claim is released when termination fails.

```
uv run pytest test/test_python_strategy_stop_lock.py -v     # 10 passed
uv run pytest test/ -k "strategy or python_strat or lock"   # 1351 passed, 3 skipped, 1 xfailed
uv run ruff check blueprints/python_strategy.py test/test_python_strategy_stop_lock.py
```

The `ruff format` drift in `python_strategy.py` still pre-exists on `main` and none of it is on lines this PR touches.

marketcalls:
All three valid, fixed in `f8cd7019e`.

**`delete_strategy` ignored the stop's return value (P1).** Correct, and the previous commit made it reachable in a new way: a stop can now fail because another stop already owns the claim, not only because the process outlived both signals. Either way the process is still running, and deleting took its config and its file with it, leaving a live strategy with nothing tracking it and no route to stop it. The route now refuses with a 409 that names the reason.

**Broadcast ordering (P2).** Also right. The broadcast sat after the `finally` that releases the claim, so a start could begin, broadcast `running`, and then be overwritten by the finishing stop's `stopped`, leaving every watching page showing a running strategy as stopped. The broadcast, the log line and the log cleanup are now inside the claim.

**The `spawned` assertion was vacuous (P2).** Entirely right, nothing ever appended to that list. Worth noting what fixing it properly required, because the obvious approach does not work:

- `subprocess.Popen` cannot be patched here. `stop_strategy_process` branches on `isinstance(process, subprocess.Popen)`, so a non-class stand-in raises `isinstance() arg 2 must be a type`.
- The spy is now on `create_subprocess_args`, which only `start_strategy_process` calls and which runs immediately before the spawn.
- `check_master_contract_ready` had to be stubbed as well. Without it the start refuses at that gate before ever reaching the spawn, so the spy could never fire and the assertion would have stayed vacuous in a different way.

Verified by removing the stopping guard: the test then fails on the spy with `start reached the spawn path while a stop was in flight, so a second process would have been created for one strategy`. That assertion is ordered first now, so the failure names the defect rather than the downstream message mismatch.

```
uv run pytest test/test_python_strategy_stop_lock.py -v     # 10 passed
uv run pytest test/ -k "strategy or python_strat or lock"   # 1351 passed, 3 skipped, 1 xfailed
uv run ruff check blueprints/python_strategy.py test/test_python_strategy_stop_lock.py
```

PR #1991, opened 2026-09-04
feat(orders): idempotent order placement via client_order_id

spsp21366-commits:
feat(orders): idempotent order placement via client_order_id

Adds an optional client_order_id to the placeorder API. A retry of a
timed-out request with the same id replays the original broker response
instead of double-placing the order.

- database/idempotency_db.py: (sha256(apikey), client_order_id) ->
  {orderid, tag, status} store with in-flight reservations, release on
  failure, and 24h TTL; own SQLite DB, NullPool per project policy
- place_order_with_auth: reserve before the broker call, replay 200 +
  duplicate:true on placed keys, 409 while in flight, release on any
  failure so corrected retries are not blocked; live path only (sandbox
  accepts but does not record the field)
- orderbook_service: echo recorded client_order_id / tag onto
  broker-proxied order rows, keyed by orderid
- OrderSchema: optional client_order_id and tag (1-128 chars), dropped
  when absent so existing payloads are unchanged; tag keeps its native
  broker-side flow, client_order_id is never forwarded to brokers

Closes the double-placement half of the client_order_id feature request
(Discord). Idempotency is scoped per apikey; the plaintext key is never
stored.

---

Happy to adjust the response shape (`duplicate` flag, 409 semantics) or scope to maintainer preference.

---
## Summary by cubic
Adds an optional `client_order_id` to the placeorder API so retrying a timed-out request with the same id replays the original broker response instead of double-placing the order. Requests without it keep the existing behavior.

- A retry while the original placement is in flight returns 409; a confirmed broker rejection releases the key so a corrected retry can proceed.
- Ambiguous outcomes (exceptions, 5xx, ACK without an order id) block the key; a retry reconciles against the broker order book — one live match with matching parameters (normalized to schema defaults on both sides) replays it, none releases it, and ambiguity or a changed-order retry (different quantity, price, product, or pricetype) keeps it blocked.
- Store failures refuse the placement with 503 instead of placing without duplicate protection; a write failure after a successful broker placement still returns the broker's 200.
- Keys are scoped per apikey (SHA-256 hashed, plaintext never stored), expire after 24 hours, and are created on existing installs by `upgrade/migrate_idempotency_store.py`.
- `client_order_id` is never forwarded to brokers; `tag` is only persisted alongside it and keeps its existing broker-side flow.
- Live path only — sandbox accepts the field but does not record it; orderbook rows echo the recorded `client_order_id` and `tag`.

<sup>Written for commit 0646838ee6696b1866b99415d58b0c57f243374f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1991?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/database/idempotency_db.py
+++ b/database/idempotency_db.py
@@ -1,0 +1,239 @@
+# database/idempotency_db.py
+"""
+Application-level order idempotency store.
+
+Maps (api_key_hash, client_order_id) -> broker orderid so a client that
+retries a timed-out /api/v1/placeorder POST gets the existing order echoed
+back instead of a duplicate position. Also records the caller-supplied
+`tag` so the orderbook can echo it.
+
+Design notes:
+
+* **Keyed by SHA-256 of the OpenAlgo apikey**, never the plaintext. The
+  auth module already treats the apikey as a secret (Argon2 verify, hashed
+  cache keys); this store must not introduce a plaintext copy.
+* **One row per placement attempt resolution.** A successful placement
+  writes (client_order_id -> orderid). A duplicate POST short-circuits at
+  the service layer and replays the recorded response — the broker is never
+  called twice for the same client_order_id.
+* **In-flight reservations.** A row is written before the broker call
+  (status="in_flight") so a retry racing a slow placement cannot double-fire.
+  If the placement fails, the reservation is released so a corrected retry
+  with the same id is not blocked forever.
+* **TTL.** Rows expire after ORDER_IDEMPOTENCY_TTL_HOURS (default 24h) so
+  the table cannot grow unbounded; a client_order_id reused after expiry is
+  treated as new. This matches broker-side order-book retention horizons
+  closely enough for the timeout-retry use case.
+* **Own SQLite file** (idempotency.db) with the project-wide NullPool
+  engine policy — writes are one per order placement, reads are one per
+  placement, so contention is negligible.
+"""
+
+import hashlib
+import os
+import threading
+from datetime import datetime, timedelta
+
+from sqlalchemy import Column, DateTime, Integer, String, UniqueConstraint, create_engine
+from sqlalchemy.exc import IntegrityError
+from sqlalchemy.ext.declarative import declarative_base
+from sqlalchemy.orm import scoped_session, sessionmaker
+from sqlalchemy.pool import NullPool
+
+from utils.logging import get_logger
+
+logger = get_logger(__name__)
+
+IDEMPOTENCY_DATABASE_URL = os.getenv("IDEMPOTENCY_DATABASE_URL", "sqlite:///db/idempotency.db")
+
+if IDEMPOTENCY_DATABASE_URL and "sqlite" in IDEMPOTENCY_DATABASE_URL:
+    # SQLite: NullPool to prevent connection pool exhaustion (project-wide policy)
+    idempotency_engine = create_engine(
+        IDEMPOTENCY_DATABASE_URL, poolclass=NullPool, connect_args={"check_same_thread": False}
+    )
+else:
+    # For other databases like PostgreSQL, use connection pooling
+    idempotency_engine = create_engine(IDEMPOTENCY_DATABASE_URL, pool_size=10, max_overflow=20)
+
+IDEMPOTENCY_TTL_HOURS = int(os.getenv("ORDER_IDEMPOTENCY_TTL_HOURS", "24"))
+
+idempotency_session = scoped_session(
+    sessionmaker(autocommit=False, autoflush=False, bind=idempotency_engine)
+)
+IdempotencyBase = declarative_base()
+IdempotencyBase.query = idempotency_session.query_property()
+
+
+class ClientOrderId(IdempotencyBase):
+    """A client-supplied idempotency key and the order it resolved to."""
+
+    __tablename__ = "client_order_ids"
+
+    id = Column(Integer, primary_key=True)
+    api_key_hash = Column(
+        String(64), nullable=False, index=True
+    )  # sha256 hex of the OpenAlgo apikey
+    client_order_id = Column(String(128), nullable=False)
+    orderid = Column(String(64), nullable=True)  # broker/OpenAlgo orderid once known
+    tag = Column(
+        String(128), nullable=True
+    )  # caller's original `tag` passthrough, echoed in the orderbook
+    # in_flight: broker call not yet resolved. placed: orderid recorded.
+    status = Column(String(16), nullable=False, default="in_flight")
+    created_at = Column(DateTime, nullable=False, default=datetime.now)
+    updated_at = Column(DateTime, nullable=False, default=datetime.now, onupdate=datetime.now)
+
+    __table_args__ = (
+        # The idempotency key is (user, client id) — one resolution per pair.
+        UniqueConstraint("api_key_hash", "client_order_id", name="uq_client_order_ids_key"),
+    )
+
+
+_init_lock = threading.Lock()
+_initialized = False
+
+
+def _hash_api_key(api_key: str) -> str:
+    return hashlib.sha256(api_key.encode("utf-8")).hexdigest()
+
+
+def init_idempotency_db() -> None:
+    """Create tables if absent. Idempotent; safe to call per-request."""
+    global _initialized
+    if _initialized:
+        return
+    with _init_lock:
+        if _initialized:
+            return
+        IdempotencyBase.metadata.create_all(bind=idempotency_engine)
+        _initialized = True
+
+
+def _prune_expired(session) -> None:
+    """Delete rows older than the TTL. Called opportunistically on writes."""
+    cutoff = datetime.now() - timedelta(hours=IDEMPOTENCY_TTL_HOURS)
+    session.query(ClientOrderId).filter(ClientOrderId.created_at < cutoff).delete()
+    session.commit()
+
+
+def reserve_client_order_id(
+    api_key: str, client_order_id: str, tag: str | None = None
+) -> tuple[str, str | None]:
+    """Claim (api_key, client_order_id) before the broker call.
+
+    INSERT-first: the unique constraint on (api_key_hash, client_order_id)
+    makes the claim atomic, so two concurrent retries cannot both proceed.
+
+    Returns:
+        ("reserved", None) — this call created the reservation and must
+        proceed with the placement, then record_success() or release it.
+        ("existing", status) — another call already claimed the key; status
+        is "placed" (replay the recorded orderid) or "in_flight" (a placement
+        is racing right now; caller should report 409).
+    """
+    init_idempotency_db()
+    key_hash = _hash_api_key(api_key)
+    with _init_lock:
+        try:
+            row = ClientOrderId(
+                api_key_hash=key_hash,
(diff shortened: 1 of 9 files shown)


spsp21366-commits:
Follow-up hardening pushed to this branch (3 commits, TDD throughout — the suite now covers 27 cases):

1. **83fa18042** — a broker 200 *without* an order id was reported as plain success while the reservation silently stayed in-flight. It is now reported as an unresolved placement (explicit 500 + `OrderFailedEvent`), and the reservation is **kept**, so retries of that id keep getting 409 instead of double-placing a possibly-live order. Callers without `client_order_id` are unaffected.

2. **3f4ff0bad** — perf: `client_order_ids` is now indexed on `created_at` (the TTL prune ran a full-table DELETE on *every* reserve) and on `(api_key_hash, orderid)` (orderbook label echo per poll). Existing installs are upgraded in-place by `init_idempotency_db()` via `CREATE INDEX IF NOT EXISTS`.

3. **e3defe8ec** — robustness: an idempotency-store failure (e.g. SQLite locked) used to escape as a raw 500 *after* the caller had opted into duplicate protection. All store interactions are now contained fail-closed: retryable 503 + event, never a silent fall-through to an unprotected placement.

Happy to split any of these into a separate PR if you'd rather keep this one minimal — say the word and I'll rebase them out.

marketcalls (commented):
Review score: 65/100 at `e3defe8ec2fb25962a8c5327ad9aeb46579fc614`. Changes are needed before approval and merge.

- `services/place_order_service.py:297`: Any broker exception releases the idempotency reservation, including an ambiguous timeout after the broker accepted the order. Retrying the same client_order_id can place a second real order, defeating the new retry-safety guarantee. Release only on a confirmed non-placement/rejection. Keep ambiguous outcomes reserved, including transport failures converted into adapter 500 responses; expose an unresolved state and reconcile with the broker before another placement. Test lost acknowledgements, invalid replies and normalized 500 responses followed by same-ID retries.

- `database/idempotency_db.py:61`: The new scoped session is not registered for application/background cleanup, and read helpers leave its transaction open. get_resolution/get_labels_for_orderids and the reservation conflict re-read can retain SQLite connections in the long-lived worker. Direct create_engine also bypasses the mandated cooperative SQLite engine factory. Use create_db_engine, register idempotency_session in the central cleanup registry, and ensure every request/background/error path removes it. Verify repeated successful reads, retries and failures release connections; wire initialization/migration into the documented upgrade flow.

Validation: Inspected the database, schemas, placement/orderbook integration and tests. Lost-ack duplicate placement reproduced with inspected functions and synthetic broker/state stubs; no live calls. Independent exact Zerodha adapter plus real isolated SQLite reservation store reproduced two acceptances after httpx.ReadTimeout and retry. A real NullPool connection probe remained checked out after project cleanup and returned only after explicit idempotency_session.remove().

spsp21366-commits:
Thanks for the review - the double-placement is fixed and the store now follows the house SQLite/session contract. Branch rebased onto current main (both registry files gained upstream entries meanwhile). Three commits:

**2423545ff - house contract**
- engine built via `database/engine_factory.create_db_engine` (the hand-rolled NullPool branch is gone)
- `idempotency_session` registered in `utils/db_sessions.SCOPED_SESSION_MODULES`; read helpers (`get_resolution`, `get_labels_for_orderids`, the reserve conflict re-read) end their transactions so a NullPool connection is never pinned between operations; the strategy-dispatch background path releases via `remove_all_scoped_sessions()` in a `finally` (tick-feed green threads never hit `teardown_appcontext`)
- `upgrade/migrate_idempotency_store.py` registered in `migrate_all.py`, so existing installs get the store through the documented upgrade flow
- regression tests: reads do not hold transactions open; engine pool policy; registry entry

**147019deb - release only on a confirmed rejection**
- released only when the response provably refused the order (`CONFIRMED_REJECTION_STATUSES`: 400/401/403/404/405/409/422/429)
- ambiguous outcomes - escaping exceptions, 5xx (including transport failures converted into adapter 500s), and a 200 without an order id - move the reservation to a new **unresolved** state (`mark_unresolved`, never overwriting `placed`); same-id retries return 409
- tests: adapter 500 after acceptance -> unresolved + retry 409 with exactly one placement; escaping exception -> same; 429 -> released, retry places; 200-without-orderid -> unresolved

**65001c3b9 - reconcile before another placement**
- the reservation now records the original order parameters (symbol/exchange/action/quantity/price); existing stores are upgraded in place at init and by the migration
- a retry of an unresolved key reconciles against the live broker order book (mapped uniform shape via `services.orderbook_service.get_orderbook_with_auth`):
  - exactly one live (non-rejected/cancelled) match with equal parameters -> the key resolves to that orderid and the retry replays it (`duplicate: true, reconciled: true`); the broker is never asked to place again
  - zero live matches with a reachable book -> the attempt provably never placed; the key is released and the retry places fresh
  - book unavailable or multiple matches -> 409, key stays blocked
- a corrected retry (parameters differ from the reservation) is refused with 409: the ambiguous attempt may still be live with the original parameters, so neither releasing nor replaying would be safe
- reconciliation is fail-safe: any error answers 409-unresolved, never releases the key and never surfaces as a store failure (503)

The matching is deliberately conservative - replaying a matching order is safe because that order already exists, while any ambiguity blocks the key instead of risking a second placement.

39 cases in `test/test_place_order_idempotency.py`; backend CI green on 3.12/3.13/3.14.

spsp21366-commits:
Follow-up hardening pushed (0accbd694 + 0d9ff3320):

- **record_success is now failure-tolerant**: if the idempotency DB write fails after the broker accepted the order, the caller still receives the broker's 200 and `OrderPlacedEvent` (with a CRITICAL log), instead of a misleading 500 with the row stuck in_flight - the order itself can no longer be lost from the caller's point of view
- the double-vacated reserve race is guarded: if no reservation could be obtained after the retry, the placement is refused with 503 instead of proceeding unprotected
- `get_labels_for_orderids` is skipped when the request carries no apikey (internal orderbook callers) - no AttributeError log spam per poll
- `tag` is dropped when `client_order_id` is absent, so the schema never forwards a dangling label; test isolation now uses an unconditional temp-file store (an exported `IDEMPOTENCY_DATABASE_URL` can never point the cleanup fixture at a real database)
- tests: 39 cases, all green

Vamana41:
Thank you for the review — the release-policy finding was correct and is fixed. Addressed across `147019d`, `65001c3`, `0accbd6` and `0d9ff33` (pushed 2026-09-06):

- **`place_order_service.py` release policy:** the reservation is now released **only** on a confirmed rejection (adapter refusals and 429 rate-limit — tests pin both). Broker exceptions, a 200 without an orderid, and inconclusive adapter 500s move the key to the **unresolved** state instead. A retry with an unresolved id gets 409 and triggers `_reconcile_unresolved_key` (`65001c3`): a matching order in the broker book is replayed as `duplicate: true`, a provably never-placed order releases the key for a fresh placement, and any reconciliation error keeps the key blocked (fail-closed). Tests in `test/test_place_order_idempotency.py`: lost-ack (broker exception) retry, ambiguous-500 retry, 200-without-orderid retry stays blocked, 429 / non-200 release paths, same-id replay without re-placement.
- **Session/lifecycle hygiene:** `idempotency_session` registered in `SCOPED_SESSION_MODULES` (`utils/db_sessions.py`) so the central cleanup removes it on every request/background path; engine built via the mandated `create_db_engine` factory; migration shipped as `upgrade/migrate_idempotency_store.py` alongside the existing upgrade scripts; test store made file-backed so tests exercise real SQLite cleanup (`0d9ff33`).
- **Store unavailability:** reserve failures now refuse the placement (503) rather than placing without duplicate protection (`5063c9e`).

CI (backend-lint, backend-test 3.12/3.13/3.14, frontend-build) green at head `0d9ff33`. Ready for re-review.

Vamana41:
Correction to my earlier "ready for re-review" note: going back through the morning review messages properly, most of the findings were still open at `0d9ff33` — my re-verification was too optimistic. All of them are addressed in `bff7484`:

1. **Reservation matched on a subset of params (P1)** — the reservation now also records `product`, `pricetype` and `trigger_price`; a corrected retry that changes any of them is refused with 409 instead of silently reusing the unresolved key.
2. **`quantity` INTEGER collapses fractional crypto lots (P1)** — the column is now FLOAT (fresh store; the migration script rebuilds the table for existing reviewer DBs, since SQLite can't ALTER a column type).
3. **Price truthiness dropped 0.0 (P2)** — recording is now `is not None`, so a LIMIT price of 0.0 is stored as 0.0. Market-type orders skip the price comparison entirely: the broker book reports the average fill there, not a caller-fixed limit price.
4. **Raw request reached the reserve path (found while testing #3)** — `validate_order_data` discards the schema-loaded dict, so absent shaping fields are now defaulted at the store (MIS / MARKET / 0.0), matching what the broker will actually execute.
5. **`resolve_live_auth` ran outside try/finally (P2)** — moved inside it, so the scoped auth session is released on the early-return path too.
6. **Migration missed `sqlite+pysqlite:///` (P2)** — handled.
7. **Reconciliation skipped `CANCELLED` but not `CANCELED` (P2)** — both spellings are skipped now.
8. **Datetime convention comment (P3)** — corrected; it cites `strategy_book_db` / `master_contract_status_db` as the naive-`datetime.now()` references.

`test/test_place_order_idempotency.py` covers each finding (param-mismatch 409s including fractional quantity and limit-price changes, zero-price recorded, raw-request defaults, CANCELED release, fill-price reconcile). 50/50 in the file locally; the full-tree run matches the pre-change baseline, and CI is green on `bff7484`.

Vamana41:
All four verified and fixed in `970a24a` (CI green):

1. **P1 matcher refused the identical raw retry** — confirmed: the stored side carries the schema defaults while the retry arrives raw, so `"" != "MARKET"` 409'd a legitimate repeat. Both sides of `_params_match_reservation` now normalize `product`/`pricetype` with the same `DEFAULT_*` constants; broker book rows always carry them explicitly, so normalization is a no-op there. Regression test: `test_identical_raw_retry_reconciles_not_409`.
2. **P2 legacy rows with NULL shaping fields** — handled in the matcher: rows created before `product`/`pricetype`/`trigger_price` existed (the ALTER migration leaves them NULL) keep the original five-field semantics and stay reconcilable. Test: `test_legacy_null_shaping_rows_keep_five_field_semantics`.
3. **P3 dead `MARKETISH_PRICETYPES`** — the matcher now imports and uses it; the inline duplicate is gone.

Vamana41:
Fixed in `0646838` — valid catch: `_make_unresolved` merged `BASE_ORDER` (explicit `pricetype`/`product`) into the payload, so only the retry side was raw. The initial placement now goes through the same raw payload the retry uses, with assertions that the store applied the schema defaults (`MARKET`/`MIS`) to the raw original — the raw-original vs raw-retry path the docstring described is now genuinely exercised. 52/52 in-file, CI green.

PR #1998, opened 2026-09-04
fix(zerodha): report MCX quantity in units, like every other broker 

Kalaiviswa:
fix(zerodha): report MCX quantity in units, like every other broker

---
## Summary by cubic
Fixes Zerodha MCX order sizing so quantity is in units, like every other broker, instead of Kite's contract count. One lot of CRUDEOIL is now quantity 100, the same as on Angel. Also fixes scalping orders on MCX, BFO and CDS, where an unconfigured freeze quantity refused every order. Zerodha MCX orders must now use units, not contracts, and the master contract must be re-downloaded for the new lot sizes to take effect.

**MCX units**
- Adds an MCX contract-size table and converts quantity at every Kite boundary: place/modify orders, GTT, margin, order/trade/position books, funds, and the websocket order stream.
- The master-contract row is authoritative, so a revised contract size resolves per expiry (MCXBULLDEX is 15 from Nov 2026); a revised underlying without a row is refused outbound and passed through inbound, never guessed.
- Refuses non-whole-contract quantities instead of rounding, and names the offending field, so a bad `disclosed_quantity` is not reported as "Quantity".
- Rejected requests return a 400 naming the reason instead of the service layer's generic 500; a margin basket refuses outright rather than pricing the remaining legs; close-all reports exits it could not place.
- Funds P&L and tradebook value now correct for the quotation unit, so the gold family and base metals are no longer misvalued.

**Freeze quantity**
- `get_freeze_qty` returns 0 for an unconfigured exchange instead of 1; the old default read as a real one-unit limit and refused every order on MCX, BFO, and CDS.

<sup>Written for commit c3d302b2d68cbcffca625bb90b628dba10653c29. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/1998?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


marketcalls (commented):
Review score: 83/100 at `e666e5d78392420496dd3e12148036d913e362de`. Changes are needed before approval and merge.

- `broker/zerodha/api/funds.py:117`: Physical quantity units are reused as the P&L multiplier even when quotes use a different unit. One GOLDM contract is converted to quantity100, so a one-rupee price move now reports INR100 unrealized P&L instead of INR10. The changed lot sizes also feed sandbox arithmetic that assumes quantity and quoted-price units agree. Keep quantity conversion separate from price-value conversion: use Kite broker-computed P&L or an explicit quote-unit multiplier for funds, and audit sandbox/trade-value consumers when changing lotsize. Add GOLDM/GOLD/ZINC dimensioned P&L tests alongside CRUDEOIL round trips.

PR #1999, opened 2026-09-05
feat(trading): intraday position calculator with risk controls and symbol exit watch

Narasimha722:
## Position Calculator: intraday sizing, order type, risk controls and symbol exit watch

A complete upgrade to the charting terminal's order flow. The calculator popup that now appears before every Buy/Sell on the chart computes a safe maximum quantity, lets the user pick a trade type (Intraday / Overnight / GTT), an order type (Market / Limit) and a routing venue (NSE / BSE for cash equity), and optionally attaches risk legs (Stop Loss / Target / Trailing Stop) - which the backend now turns into an automatic square-off via the **symbol exit watch**, so the SL/target actually fire in both sandbox and live mode.

Buy Order

Sell Order

### What changed

**Calculator UX**
- NSE/BSE routing toggle for cash equity: live quote, leverage, max quantity and brokerage estimate all follow the chosen exchange; `onConfirm` reports it and the order is routed accordingly. F&O stays on its native exchange.
- Draggable dialog, position persisted to `localStorage` (`openalgo:position-calculator-pos`) and clamped to the viewport - including a fix for a drag bug where only the text moved out of a static frame.
- Trade Type segmented control: Intraday (`MIS`) / Overnight (`CNC`/`NRML`) / GTT. Leverage sizing applies to Intraday only; Overnight/GTT size at cash value.
- Order Type: Market executes now at LTP; Limit schedules at a chosen price (buy below LTP, sell above LTP) and is used as the sizing basis.
- 3D dark-glass UI: raised/inset tiles for Buy/Sell, exchange, order type and trade type; gradient glow cards; pure Tailwind, no new dependencies.
- Hidden brokerage estimator: shows per-component charges + total for the four brokers the platform carries a tariff sheet for (Fyers, Zerodha, Dhan, Groww), and backs a new Charges column in the order book.

**Symbol Exit Watch (backend)** - makes the Risk Management block real
- `database/symbol_exit_db.py` - new `symbol_exit_watch` table (one row per entry order that carried risk legs, `UNIQUE(order_id, mode)`, NullPool engine, session registered in the central teardown registry).
- `services/symbol_exit_monitor_service.py` - singleton monitor: gathers watches for the current trading mode, subscribes each symbol's LTP on the shared WebSocket proxy client, evaluates every tick with the shared `services/risk` core, and markets a single square-off (opposite side, same symbol/exchange/product) when the stop, target or trailed stop is reached. Mode rudder (a watch never exits the other mode), net-position guard (flat position closes the watch instead of double-exiting), entry-price seeding for market fills, persist throttle and retry cooldown.
- `place_order_with_auth` snapshots the risk legs before the live branch strips them and registers a watch on the success path of both the analyze (sandbox) and live branches - best-effort, so a failed watch never fails a placed order.
- Migration `upgrade/migrate_symbol_exit_watch.py` (idempotent, `--status` aware) registered in `migrate_all.py`.
- Tests `test/test_symbol_exit_watch.py` - 15 locked decisions (long/short SL and target breaches, trailing-stop tightening, registration gate).

### Transport detail
The calculator's risk legs ride on a single REST `placeorder` request (the chart feed socket has no slot for them). On success the backend records the exit watch; the broker always receives a clean single-leg order (risk keys are stripped before the live broker call so generic brokers never reject unknown keys). The exit - not the broker - acts on the levels, which is what makes SL/target work for every broker and every trade type.

### Verification
- `uv run pytest test/test_symbol_exit_watch.py` - 15 passed
- `uv run pytest test/risk test/test_strategy_module_risk.py` - 214 passed
- `npm audit` - 0 vulnerabilities; `pip-audit` - no known vulnerabilities
- `bandit` on the new/changed services - no medium/high findings
- Frontend lint + TypeScript build clean

### Screenshots
- `Buy_Order.png` - calculator on a Buy order
- `Sell_order.png` - calculator on a Sell order


marketcalls (commented):
Review score: 50/100 at `9ddcda9d7d8d8dc96b9be5ef090789ae7003b850`. Changes are needed before approval and merge.

- `services/symbol_exit_monitor_service.py:349`: Live protection is disabled by the global Analyzer toggle, and exit submission does not force the watch entry mode. sync loads only the current mode; _on_tick/_exit_worker skip live watches after a toggle. A toggle between the worker check and place_order can route a live exit to sandbox and then mark the live watch closed. Keep live watches active independently of the global UI mode. Bind all position/order lookups, cancellation and exit placement to the stored watch mode, using force_live=True for live protection. Add toggle-before-tick and toggle-between-check-and-submit regressions.

- `services/symbol_exit_monitor_service.py:624`: Broker acknowledgement immediately marks an exit watch executed. An accepted order can subsequently be rejected or partially filled; the watch is already removed and the remaining exposure loses protection. Persist the exit order ID and reconcile its fills and remaining watched quantity before closing. Keep unresolved/rejected/partial exits managed with an atomic retry/claim lifecycle.

- `services/symbol_exit_monitor_service.py:574`: Real OS exit threads call eventlet-managed trading services, and the tick handler performs database work while holding a real RLock. These cross-thread paths violate the explicit CLAUDE.md production boundaries and can block the single eventlet hub or touch green synchronization from a real thread. Dispatch application work through the supported green-thread path. Restrict real locks to in-memory snapshots and perform database/broker/event-bus work after release. Use centralized session cleanup for background tasks and verify under eventlet with concurrent ticks and exits.

- `frontend/src/lib/trading/terminal.ts:1278`: Adding any risk field bypasses the chart trading feed and its displayed-mode guard. A stale Analyze badge after another session switches the server to live now permits a real placement through raw api(placeorder), where the ordinary order path would refuse the mismatch. Extend the shared mode-checked placement path to carry risk fields. Test stale displayed/server mode with stop, target and trailing fields and assert no placement occurs on mismatch.

- `frontend/src/components/trading/PositionCalculator.tsx:543`: The DialogContent Enter handler confirms orders regardless of which control has focus. Enter on Cancel or Close bubbles into handleConfirm and can submit an order when the operator intends to cancel. Use explicit form submission/confirmation semantics and exclude cancellation and other buttons. Add keyboard regression tests for Cancel, Close, toggles and Confirm.

- `frontend/src/components/trading/ChartPane.tsx:1290`: The new calculator omits the fullscreen portal container used by the existing order ticket. Its default document.body portal is outside the fullscreen pane, making market/limit confirmation inaccessible in fullscreen. Forward a container prop through PositionCalculator to DialogContent and pass menuHost from ChartPane; verify confirmation and cancellation in fullscreen.

- `frontend/src/components/trading/PlaceOrderDialog.tsx:299`: SL-M risk validation uses LTP instead of the intended entry trigger. A BUY ticket with LTP 100, target 110 and stop-market trigger 130 passes even though its target is below the intended entry; the backend only checks relative levels for LIMIT. Use the trigger as the stop-market entry basis and revalidate whenever type, trigger, direction or price changes. Cover BUY and SELL transitions from calculator to SL-M.

- `services/brokerage_charges.py:264`: Currency and commodity derivatives are mapped to equity Futures/Options tariff rows. A Fyers CDS USDINR future probe returns STT 42.50 on turnover 85000, while this PR own Currency Futures CSV row specifies zero. Resolve currency/commodity segments and exchange rules correctly or reject unsupported estimates explicitly. Add vectors that distinguish these segments from equity F&O.

- `services/brokerage_charges.py:329`: The estimate routes accept non-finite numbers and do not validate JSON object shapes. price="nan" or quantity="inf" returns 200 success with invalid totals; a list root or a null batch member raises AttributeError and returns 500. Validate object roots/items, field types, positive finite values and finite computed turnover. Return a normal 400 or documented per-item error and add HTTP regression cases.

- `frontend/src/components/trading/PositionCalculator.tsx:630`: New calculator controls do not meet the current plain-text and accessibility standards. New icons contradict CLAUDE.md; numeric labels lack htmlFor/id associations and selected tiles only expose visual CSS state. Use plain-text controls, associate each label with its input and expose tile selection with aria-pressed. Keep feature documentation in the canonical docs tree.

Validation: Inspected the full monitor and surrounding watch database, placement integration and regression tests. Synthetic exact-method checks confirmed mode-dependent lost protection, missing force_live and ACK-only closure. Supplemental backend/frontend review is recorded separately. Independent frontend review reproduced mode-guard bypass, keyboard confirmation and SL-M validation defects with mocked I/O. All 21 brokerage tests passed; additional exact-route probes exposed tariff, non-finite and shape errors. No live browser/fullscreen, broker or production eventlet run was performed.

Issue #1394, opened 2026-05-08
CRITICAL ISSUE: Latest commit crashes


aravindgandavadi:
@marketcalls @Kalaiviswa 

After installing the latest version, I am getting the following errors in the log:

[OpenAlgo Fernet salt] Could not write FERNET_SALT to .env ([Errno 13] Permission denied: '/app/utils/../.env.tmp').
Migration aborted before any DB changes; safe to retry.

Deployed version: 4e4ef9b
Environment: Azure VM/Docker
Installation Method: install-docker-multi-custom-ssl.sh

Please fix it immediately. Let me know if you need more details.

marketcalls:
Hi @aravindgandavadi — fixed in a1455ff1.

**Root cause (two compounding issues):**

1. **`/app` is root-owned in the container.** The Dockerfile's `COPY --chown=appuser:appuser . .` chowns the *contents* of `/app/`, but `/app/` itself was created by `WORKDIR /app` as `root:root mode 755`. appuser can read/execute it but cannot create new files in it. The atomic-write helper tries `open("/app/.env.tmp", "w")` → **EACCES**, which is the exact error you saw.

2. **`.env` is a single-file bind mount.** Even if I fix #1, `./.env:/app/.env` makes `.env` live on the host filesystem while `.env.tmp` would live in the container's overlay filesystem. `rename(2)` refuses to span mount points and returns EXDEV/EBUSY. The standard atomic-write pattern fundamentally cannot work for any Docker user regardless of permissions.

**Fix in two layers:**

- **Python (`utils/env_check.py`)**: `_atomic_replace_text` now detects EACCES / EPERM / EXDEV / EBUSY and falls back to an in-place rewrite (open destination directly, truncate, write, fsync). Not crash-atomic in the strictest sense (a kill -9 mid-write could corrupt `.env`), but acceptable for a config file written once at startup and the only viable path for bind-mounted single files in Docker.
- **Dockerfile**: chown `/app` itself to appuser so fresh installs avoid EACCES entirely. The Python fallback still protects against the bind-mount EXDEV case.

**Upgrade procedure for affected deployments**

Your instance is currently in a clean recoverable state — the migration aborted *before* any DB changes (that's the "Migration aborted before any DB changes; safe to retry" message), so your stored broker tokens, API keys, and TOTP secrets are intact, still encrypted with the legacy static salt.

```bash
cd <your-instance-dir>          # e.g. /opt/openalgo/upstox.valmiki.trade
[...]

PR #2004, opened 2026-09-05
fix(docker): move base images from EOL bullseye to trixie

aravindgandavadi:
## Problem

Debian 11 "bullseye" reached end of LTS on 2026-08-31. Its security pool has since been drained while `bullseye-security` still serves an index advertising the removed files, so `apt-get update` succeeds and the install step then fails:

```
E: Failed to fetch .../systemd_247.3-7+deb11u8_arm64.deb    404  Not Found
E: Failed to fetch .../tzdata_2026b-0+deb11u1_all.deb       404  Not Found
E: Failed to fetch .../libtiff5_4.2.0-1+deb11u8_arm64.deb   404  Not Found
E: Failed to fetch .../libgbm1_20.3.5-1+deb11u1_arm64.deb   404  Not Found
```

This breaks **every** image build, not only fresh installs — any existing deployment that rebuilds hits it. The index carries `Valid-Until: Mon, 07 Sep 2026`, after which apt refuses the repository outright and even `apt-get update` fails.

## Why trixie and not bookworm

Bookworm's regular security support ended **2026-07-11**; it is already LTS-only and would need this same change again in 2028. Trixie has security support to 2028-08-09 and LTS to 2030-06-30.

## Changes

Three `FROM` lines, plus a comment recording why:

| stage | before | after |
|---|---|---|
| python-builder | `python:3.12-bullseye` | `python:3.12-trixie` |
| frontend-builder | `node:22-bullseye-slim` | `node:22-trixie-slim` |
| production | `python:3.12-slim-bullseye` | `python:3.12-slim-trixie` |

No package names changed — every explicitly installed package exists in trixie. The 404s were all transitive dependencies. Docs updated to match; the ASCII stage diagram is re-padded so column alignment is unchanged.

## Verification

Built and deployed on a live arm64 Azure VM (Ubuntu 24.04 host, Docker):

- Build completes 30/30 in 151s
- `groupadd --gid 1000 appuser` succeeds — the UID pin added for #1394 holds on trixie, which was the main risk in changing base images
- Kaleido static export verified in-container against Chromium 150: `fig.to_image(format="png", engine="kaleido")` → `KALEIDO OK`
- Instance came up healthy and is serving

## Image size impact

The production image grows from 2.14 GB to 2.71 GB on arm64. Cleanly attributable to the distro is the apt layer, **570 MB → 751 MB (+181 MB)**, consistent with `chromium` (214→276 MB) and `chromium-common` (17→85 MB). The remaining delta sits in the venv and source layers and is not cleanly attributable — the bullseye image I compared against was built from an older commit, so those layers mix code changes with any wheel-selection differences from the newer glibc.

Chromium is `150.0.7871.100` in both bookworm and trixie, so this cost is not avoidable by choosing bookworm.

## Not caused by this change

`/chart` in the Telegram bot fails with `If using all scalar values, you must pass an index` at `services/telegram_bot_service.py:227` when the history API returns a status dict instead of a DataFrame — e.g. on a non-trading day. Confirmed to fail identically on an unmodified bullseye instance, so it is pre-existing and unrelated. Filed separately.

---
## Summary by cubic
Moves all Docker base images from Debian 11 bullseye (EOL) to Debian 13 trixie to fix failing builds caused by bullseye's drained security pool.

- All three build stages now use trixie-based images (`python:3.12-trixie`, `node:22-trixie-slim`, `python:3.12-slim-trixie`).
- Bookworm was not chosen because its regular security support ended 2026-07-11; trixie offers support until 2028-08-09 and LTS to 2030.
- No explicit package names changed; only transitive dependencies were causing the 404s.
- Production image grows from 2.14 GB to 2.71 GB on arm64, driven mainly by Chromium version differences.
- Verified on a live arm64 VM: build completes, UID pin holds, and Kaleido static export works.
- Docs updated to match new base images and corrected the earlier image size estimate.

<sup>Written for commit 211bd7dd01c862d028f5ba9496c9996b30c90bc8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/2004?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,11 @@
 # ------------------------------ Python Builder Stage ----------------------- #
-FROM python:3.12-bullseye AS python-builder
+# Base images track Debian 13 "trixie". Debian 11 "bullseye" reached end of LTS
+# on 2026-08-31 and its security pool was drained days later, so every build
+# began failing with 404s on systemd, tzdata, libtiff5 and libgbm1 while the
+# package index still advertised them. Debian 12 "bookworm" is not the fix --
+# its regular security support ended 2026-07-11 and it is already LTS-only.
+# Trixie has security support to 2028-08-09 (LTS to 2030-06-30).
+FROM python:3.12-trixie AS python-builder
 RUN apt-get update && apt-get install -y --no-install-recommends \
     curl build-essential && \
     apt-get clean && rm -rf /var/lib/apt/lists/*
@@ -14,7 +20,7 @@ RUN pip install --no-cache-dir uv && \
     rm -rf /root/.cache
 
 # ------------------------------ Frontend Builder Stage --------------------- #
-FROM node:22-bullseye-slim AS frontend-builder
+FROM node:22-trixie-slim AS frontend-builder
 WORKDIR /app
 COPY frontend/package*.json ./frontend/
 RUN cd frontend && npm ci
@@ -23,7 +29,7 @@ RUN cd frontend && npm run build
 
 # --------------------------------------------------------------------------- #
 # ------------------------------ Production Stage --------------------------- #
-FROM python:3.12-slim-bullseye AS production
+FROM python:3.12-slim-trixie AS production
 # 0 – set timezone to IST (Asia/Kolkata) & install runtime dependencies
 #     chromium + fonts-liberation are required by Kaleido 1.x (plotly static
 #     image export) which drives a real headless Chromium via choreographer.
--- a/docs/design/11-docker/README.md
+++ b/docs/design/11-docker/README.md
@@ -13,7 +13,7 @@ OpenAlgo provides Docker support for containerized deployment with **3-stage bui
 
 ┌─────────────────────────────────────────────────────────────────────────────┐
 │                        Stage 1: Python Builder                               │
-│                        (python:3.12-bullseye)                                │
+│                        (python:3.12-trixie)                                  │
 │                                                                              │
 │  ┌───────────────────────────────────────────────────────────────────────┐  │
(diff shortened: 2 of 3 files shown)


marketcalls (approved):
Reviewed commit 211bd7dd01c862d028f5ba9496c9996b30c90bc8 against CLAUDE.md and the surrounding Docker build/runtime configuration. Approving the base-image change; I found no blocking regression.

The Python builder and runtime remain on Python 3.12 and the same Debian release. The official image definitions include the three selected tags on amd64 and arm64. UID/GID 1000, runtime dependencies, Chromium configuration, and eventlet/Gunicorn startup are preserved.

One nonblocking documentation correction: docs/docker/DOCKER_BUILD_GUIDE.md:108-112 lists approximately 750 MB + 1.1 GB + 50 MB + 10 MB, totaling about 1.91 GB, against a stated total of 2.7 GB. Please provide the measured missing layers or label these as selected layers. Avoid assigning the difference to the base image without measurement.

Validation: reviewed the full diff, build/runtime configuration and CI, and ran git diff --check. General CI checks passed, but Docker build jobs were skipped for this fork PR. The local Docker daemon was unavailable, so I did not independently reproduce the reported ARM64 build or Kaleido smoke test.

PR #2005, opened 2026-09-05
fix(installer): never start a container from a failed build

aravindgandavadi:
## Problem

In `install-docker-multi-custom-ssl.sh`, the deployment step was:

```bash
docker compose build
docker compose up -d
```

The build result is never checked and the script has no `set -e`, so a failed build falls straight through to `up -d`. Because the previous image is still tagged, the container restarts on the **old code**, the loop continues to the next domain, and the script prints `INSTALLATION COMPLETE` and exits 0.

The failure is silent in the worst way. The operator sees a green banner and a `healthy` container. The git checkout on disk is at the new commit and `.env` has been rewritten, but the running code is whatever last built successfully. Nothing indicates the update did not apply.

Observed in the wild when bullseye's security pool was drained after its 2026-08-31 EOL: the build failed on apt 404s, the script reported success, and the instance kept serving a five-day-old image. Verified afterwards — the image creation timestamp predated the run.

## Changes

- A failed `docker compose build` logs an error naming the domain, **skips `up -d`** so the running container is left untouched, records the domain, and continues with the remaining instances rather than aborting them.
- A container that fails to start is tracked the same way.
- The closing banner reports `INSTALLATION COMPLETED WITH ERRORS` and lists which instances were not updated.
- The script exits 1 when any instance failed, so CI and `&&` chains see it.

Deliberately not adding `set -euo pipefail`: in a 1,100-line interactive script full of `read` prompts and bare `grep`s it would cause more spurious aborts than it prevents.

## Verification

- `bash -n` clean
- Harness with a faked build failure for one of three domains: the other two deploy normally, the failed one never reaches `up -d`, the banner names it, exit code is 1
- Real run on an arm64 Azure VM: successful build still takes the happy path and prints the green `INSTALLATION COMPLETE` banner unchanged

Independent of the base-image fix; either can merge first.

---
## Summary by cubic
Fixes the installer so a failed `docker compose build` no longer silently starts the container from the previous stale image — that previously printed `INSTALLATION COMPLETE` while running old code. A failed build now also restores the instance's previous compose and Nginx config, so an already-running container stays reachable on its old ports instead of being pointed at new ports nothing serves.

- A failed build or container start skips `up -d`, restores the prior routing, logs an error naming the domain, and continues with remaining instances.
- The closing banner reports `INSTALLATION COMPLETED WITH ERRORS` and lists which instances were not updated.
- The script exits 1 when any instance failed, so CI and `&&` chains can detect the failure.

<sup>Written for commit 84bb35492012defd99639930ac88c77e76a3d8a7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/2005?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/install/install-docker-multi-custom-ssl.sh
+++ b/install/install-docker-multi-custom-ssl.sh
@@ -708,6 +708,10 @@ EOF
     fi
 fi
 
+# Instances whose image failed to build. Collected so one bad instance does
+# not abort the others, and so the final banner can tell the truth.
+FAILED_DOMAINS=()
+
 for i in "${!CONF_DOMAINS[@]}"; do
     DOMAIN="${CONF_DOMAINS[$i]}"
     BROKER="${CONF_BROKERS[$i]}"
@@ -1054,8 +1058,22 @@ EOF
     log "Starting Container for $DOMAIN..." "$BLUE"
     log "Building Docker image (includes automated frontend build, may take 2-5 minutes)..." "$YELLOW"
     cd "$INSTANCE_DIR"
-    docker compose build
-    docker compose up -d
+    # Never start a container from a stale image. A failed build leaves the
+    # PREVIOUS image still tagged, so an unconditional `docker compose up -d`
+    # restarts the old code and every later message -- including
+    # "INSTALLATION COMPLETE" -- reports success while nothing was updated.
+    if ! docker compose build; then
+        log "Error: Docker image build FAILED for $DOMAIN." "$RED"
+        log "       Its container was left untouched and is still running the" "$RED"
+        log "       previous image. This instance was NOT updated." "$RED"
+        FAILED_DOMAINS+=("$DOMAIN")
+        continue
+    fi
+    if ! docker compose up -d; then
+        log "Error: Container failed to start for $DOMAIN." "$RED"
+        FAILED_DOMAINS+=("$DOMAIN")
+        continue
+    fi
     
 done
 
@@ -1144,9 +1162,21 @@ EOF
 chmod +x /usr/local/bin/openalgo-ctl
 
 
-log "\n==============================================" "$GREEN"
-log " INSTALLATION COMPLETE" "$GREEN"
-log "==============================================" "$GREEN"
+if [ ${#FAILED_DOMAINS[@]} -gt 0 ]; then
+    log "\n==============================================" "$RED"
+    log " INSTALLATION COMPLETED WITH ERRORS" "$RED"
+    log "==============================================" "$RED"
+    log "These instances were NOT updated and are still running their" "$RED"
+    log "previous image:" "$RED"
+    for FAILED in "${FAILED_DOMAINS[@]}"; do
+        log "  - $FAILED" "$RED"
+    done
+    log "Scroll up for the build error, fix it, then re-run this script." "$YELLOW"
+else
+    log "\n==============================================" "$GREEN"
+    log " INSTALLATION COMPLETE" "$GREEN"
+    log "==============================================" "$GREEN"
+fi
 log "Management Command: openalgo-ctl" "$BLUE"
 log "  openalgo-ctl list" "$BLUE"
 log "  openalgo-ctl restart <domain.com>" "$BLUE"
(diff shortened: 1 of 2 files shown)


marketcalls (changes requested):
Reviewed commit bbb12eab5e962c711a1874f8ae9ab69c9206f3f3. The new guards correctly skip compose up after a failed build, continue deploying the remaining domains, and return exit status 1. There is one blocking regression in how that failure path interacts with the existing configuration writes.

P1: Preserve the running instance's routing when its build fails (install/install-docker-multi-custom-ssl.sh:1065-1070, followed by the Nginx reload at 1082).

The loop allocates new ports at lines 729-731 and overwrites both docker-compose.yaml and the Nginx configuration before building. The new continue leaves the existing container on its old port mappings, but the final Nginx reload activates the new upstream ports. An update that fails to build therefore makes the previously running instance unreachable through its domain.

I reproduced this with an isolated Bash harness using the actual allocator, Compose/Nginx generation, deployment branch and final reload, with Docker and system services stubbed. For the failed domain, the container retained ports 5001/8766 while Nginx was reloaded to target 5004/8769. Running the same scenario against the parent confirmed that its stale-image compose up kept the simulated mappings aligned; skipping that step introduces the mismatch.

Please preserve existing port mappings/configuration for failed updates, or stage and restore configuration so a failed domain retains its working routing before the shared Nginx reload. Add a regression case that checks the running container ports against the reloaded upstreams after a build failure, while another domain still deploys successfully.

Also adjust the new failure messages at lines 1067-1068 and 1169-1170: a fresh install has no previous container, and a failed compose up can leave a recreated container stopped. The banner should report the failure without promising the previous image is still running.

The healthcheck concern is valid, but detached startup without a readiness check predates this PR; it is a follow-up rather than the reason for requesting changes here.

Validation: bash -n and git diff --check passed. The isolated harness covered all-success, failed build, failed start, and a modeled unhealthy startup, including continued deployment and exit codes. No live Docker deployment was performed.

aravindgandavadi:
Thanks — the P1 is real, and the mechanism you describe is exactly right. The previous `docker compose up -d` was keeping the container and Nginx aligned by accident; skipping it broke that. Rebased onto merged #2004, fix in 84bb3549.

**Routing preserved on failed build.** The compose port mappings and the Nginx vhost upstreams are snapshotted immediately before regeneration and restored if the build fails, so the domain keeps pointing at the ports its container is actually listening on. For a fresh install there is nothing to restore, so the generated vhost and its symlink are removed rather than left pointing at a port that will never be served.

**No restore after a failed `docker compose up -d`.** Compose has already acted on the new file, so a container may exist stopped or partially recreated on the new ports; reverting Nginx alone would guarantee a mismatch. That path now reports the failure without claiming anything about what is running.

**Messages.** The per-domain error is conditional on update vs fresh install, and the banner reports which instances did not deploy successfully without asserting that the previous image is still serving.

**Regression test** — `test/test_installer_failure_routing.py`. A failed build alongside a domain that still deploys in the same run, asserting the failed domain's compose ports match its reloaded upstreams while the sibling advances to its new ports; plus a failed fresh install leaving no dead vhost. Both tests fail if `restore_instance_config` is neutered — I verified that rather than assuming it.

To make the failure path testable without writing to real Nginx config, those paths moved to `NGINX_SITES_AVAILABLE` / `NGINX_SITES_ENABLED` (defaulting to the previous literals), and sourcing with `OPENALGO_INSTALLER_LIB_ONLY=1` returns before any imperative section. The Portainer vhost paths are left as literals, out of scope here.

**`--wait`** left as a follow-up, as you indicated.

Validation: `bash -n`, `git diff --check`, both new tests pass. No live Docker deployment.

PR #2007, opened 2026-09-05
fix(tradebook): use fill_timestamp not order_timestamp, preserve per-fill trade_id

### LINKED ISSUE #2006: Tradebook API: wrong timestamp field (Zerodha) and dropped per-fill trade_id (Zerodha + Kotak)
## Description

`transform_tradebook_data()` uses the wrong timestamp field for a Zerodha trade record, and drops the broker's own per-fill trade ID entirely for both Zerodha and Kotak — leaving no stable per-fill identity for a consumer to key off of, and (very likely) causing intermittent missing-date timestamps we've observed in production.

## Root cause 1: `order_timestamp` used instead of `fill_timestamp` (Zerodha)

Kite Connect's own docs ([kite.trade/docs/connect/v3/orders/](https://kite.trade/docs/connect/v3/orders/)) document three distinct timestamps on a trade record:

| Field | Meaning |
|---|---|
| `order_timestamp` | "Timestamp at which the order was registered by the API" |
| `exchange_timestamp` | "Timestamp at which the order was registered by the exchange" |
| `fill_timestamp` | "Timestamp at which the trade was filled at the exchange" |

`broker/zerodha/mapping/order_data.py`'s `transform_tradebook_data()` currently does:

```python
"timestamp": trade.get("order_timestamp", ""),
```

`order_timestamp` is an **order-level** field — every fill under one multi-fill order shares the identical value, since it records when the order itself was placed, not when each individual fill executed. `fill_timestamp` is the field actually correct for a per-trade record.

### Live evidence this already causes a real bug

We observed this in production on 2026-08-18: one account's Trade Book rows showed bare `HH:MM:SS` values with no date, while an identical code path on another account (same OpenAlgo version, same broker plugin) showed full `YYYY-MM-DD HH:MM:SS` timestamps for trades on the same day. At the time we couldn't access the raw Kite response to confirm the exact cause, but `order_timestamp`'s documented semantics (an order-level field, not per-fill) are exactly the kind of field that could behave inconsistently across different order flows (e.g. a bracket/cover order vs a plain order) in ways `fill_timestamp` — always describing the fill itself — would not.

### Independent corroboration already in the codebase

`blueprints/pnltracker.py` (lines 318-319) already does this at the point of use, defensively:

```python
trade_timestamp = (
    trade.get("timestamp") or trade.get("fill_timestamp") or trade.get("fill_time")
)
```

This is a workaround for exactly this gap — falling back to `fill_timestamp` when the plain `timestamp` field (itself sourced from `order_timestamp`) looks unusable. Fixing it at the source removes the need for every consumer to carry this same defensive fallback.

## Root cause 2: `trade_id` dropped entirely (both Zerodha and Kotak)

Kite's `/trades` response includes `trade_id` ("Exchange generated trade ID") — a genuine per-fill identifier, distinct from `order_id` (shared by every fill under one order). `transform_tradebook_data()` never reads it.

Kotak's Neo API has the same field under a different name: `flId` (per [Kotak-Neo/kotak-neo-api-v2 docs/Trade_report.md](https://github.com/Kotak-Neo/Kotak-neo-api-v2/blob/main/docs/Trade_report.md)) — two fills under the same `nOrdNo` (order number) in the documented sample response carry two distinct `flId` values (`207983744` and `208109719`). This is also never read.

Without a stable per-fill ID, any consumer building something on top of the tradebook API that needs to dedup a repeated pull (e.g. a daily trade-history capture job) has no clean key — only `orderid`, which collides across every fill of a multi-fill order.

## Proposed fix

**Zerodha** (`broker/zerodha/mapping/order_data.py`):
```python
"trade_id": trade.get("trade_id", ""),
"timestamp": trade.get("fill_timestamp") or trade.get("order_timestamp", ""),
```
(fallback to `order_timestamp` kept for defensiveness, in case `fill_timestamp` is ever absent on a given trade type)

**Kotak** (`broker/kotak/mapping/order_data.py`):
```python
"trade_id": trade.get("flId", ""),
```
(Kotak's existing `timestamp: trade.get("exTm", "")` is already correct — `exTm` is the full-datetime exchange execution time — no change needed there)

Both are purely additive (`trade_id`) or a same-shape field swap (`timestamp`) — `services/tradebook_service.py`'s `format_trade_data()` passes every dict key through generically with no allowlist, so no other code needs to change for `trade_id` to reach consumers.

## Verification

- `python -m py_compile` on both files — clean.
- Confirmed no downstream code assumes a fixed key set that would need updating (`format_trade_data` is a generic pass-through).
- Live confirmation against a real account not yet done (filing this alongside the fix rather than after, since it rests on documented broker field semantics rather than a computed value — happy to report back real before/after timestamps from a live account once market hours resume).

## Regression coverage

Suggest a test confirming: (1) a Zerodha trade dict with both `order_timestamp` and `fill_timestamp` set to different values ends up using `fill_timestamp`; (2) a trade with only `order_timestamp` (no `fill_timestamp`) still gets a timestamp via the fallback; (3) `trade_id`/`flId` survive into the transformed dict unchanged for both brokers.

arsalanansari17:
Fixes #2006

## Problem

Two data-quality gaps in `transform_tradebook_data()`:

1. **Zerodha uses the wrong timestamp field.** Kite's own docs ([kite.trade/docs/connect/v3/orders/](https://kite.trade/docs/connect/v3/orders/)) document `order_timestamp` as 'when the order was registered by the API' (an order-level field, identical across every fill under one multi-fill order) vs `fill_timestamp` as 'when the trade was filled at the exchange' — the field actually correct for a per-trade record. The current code uses `order_timestamp`.
2. **`trade_id` (Zerodha) / `flId` (Kotak) — the broker's own per-fill identifier — is dropped entirely** for both brokers. Only `orderid`/`nOrdNo` survives, which is shared by every fill under one order, leaving no stable per-fill key for a consumer to dedup a repeated tradebook pull against.

## Evidence this isn't just theoretical

- We hit the timestamp issue in production: one account's Trade Book showed bare `HH:MM:SS` with no date on some rows, an identical code path on another account showed full datetimes for the same day.
- `blueprints/pnltracker.py` (lines 318-319) already independently falls back to `fill_timestamp` when `timestamp` looks wrong — informal corroboration this exact gap was already suspected, just never fixed at the source:
  ```python
  trade_timestamp = (
      trade.get("timestamp") or trade.get("fill_timestamp") or trade.get("fill_time")
  )
  ```
- Kotak's own SDK docs ([Kotak-Neo/kotak-neo-api-v2 docs/Trade_report.md](https://github.com/Kotak-Neo/Kotak-neo-api-v2/blob/main/docs/Trade_report.md)) show two fills under one `nOrdNo` carrying two distinct `flId` values in the documented sample response — confirming it's a real per-fill ID, not redundant with the order ID.

## Fix

**Zerodha:**
```python
"trade_id": trade.get("trade_id", ""),
"timestamp": trade.get("fill_timestamp") or trade.get("order_timestamp", ""),
```
(fallback kept for defensiveness in case `fill_timestamp` is ever absent)

**Kotak:**
```python
"trade_id": trade.get("flId", ""),
```
(Kotak's existing `timestamp: trade.get("exTm", "")` is already correct — full-datetime exchange execution time — no change needed there)

## Verification

- `python -m py_compile` — clean on both files.
- Confirmed `services/tradebook_service.py`'s `format_trade_data()` passes every dict key through generically with no allowlist, so `trade_id` reaches consumers with no other code changes needed.
- Live confirmation against a real account is pending market hours (filing alongside the fix since it rests on documented broker field semantics, not a computed value that needs numeric verification) — happy to report back real before/after timestamps once markets reopen.

## Regression coverage

Happy to add tests covering: a trade with both `order_timestamp` and `fill_timestamp` set differently uses `fill_timestamp`; a trade with only `order_timestamp` still gets a timestamp via fallback; `trade_id`/`flId` survive into the transformed dict unchanged for both brokers.

---
## Summary by cubic
Fixes two data-quality gaps in `transform_tradebook_data()`: uses `fill_timestamp` instead of `order_timestamp` for Zerodha fills, and exposes each fill's broker-level identifier (`trade_id` for Zerodha, `flId` for Kotak) under the canonical `tradeid` key so consumers can deduplicate repeated tradebook pulls. This addresses issue #2006.

- Zerodha now uses `fill_timestamp` (actual exchange fill time) with `order_timestamp` as fallback, fixing a known date-format bug where some rows showed bare times.
- Kotak's timestamp (`exTm`) was already correct; only the `tradeid` mapping was added.
- `tradeid` reaches consumers unchanged since `format_trade_data()` passes all dict keys through generically.

<sup>Written for commit 3044dc8d4cb0871a7311f6b88ec5b58e9dadc7e9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/2007?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/broker/kotak/mapping/order_data.py
+++ b/broker/kotak/mapping/order_data.py
@@ -226,6 +226,13 @@ def transform_tradebook_data(tradebook_data):
             "average_price": trade.get("avgPrc", 0.0),
             "trade_value": float(trade.get("fldQty", 0.0)) * float(trade.get("avgPrc", 0.0)),
             "orderid": trade.get("nOrdNo", ""),
+            # "flId" is Kotak's own per-fill trade ID (Kotak-Neo/kotak-neo-api-v2
+            # docs/Trade_report.md - two fills under the same order carry two
+            # distinct flId values). Previously dropped entirely, leaving no
+            # stable per-fill identity for a consumer to dedup repeated
+            # tradebook pulls against - only "orderid", shared by every fill
+            # under one order.
+            "trade_id": trade.get("flId", ""),
             "timestamp": trade.get("exTm", ""),
         }
         transformed_data.append(transformed_trade)
--- a/broker/kotak/mapping/order_data.py
+++ b/broker/kotak/mapping/order_data.py
@@ -231,8 +231,12 @@ def transform_tradebook_data(tradebook_data):
             # distinct flId values). Previously dropped entirely, leaving no
             # stable per-fill identity for a consumer to dedup repeated
             # tradebook pulls against - only "orderid", shared by every fill
-            # under one order.
-            "trade_id": trade.get("flId", ""),
+            # under one order. Emitted under OpenAlgo's own established key
+            # "tradeid" (no underscore) - the documented tradebook contract
+            # (docs/prompt/flow-import-format.md) and existing consumers
+            # (services/telegram_bot_service*.py, broker/groww's own
+            # adapter) already expect that exact key.
+            "tradeid": trade.get("flId", ""),
             "timestamp": trade.get("exTm", ""),
         }
         transformed_data.append(transformed_trade)
--- a/broker/zerodha/mapping/order_data.py
+++ b/broker/zerodha/mapping/order_data.py
@@ -180,8 +180,12 @@ def transform_tradebook_data(tradebook_data):
             # previously dropped entirely - only "orderid" survived, shared
             # by every fill under one order, giving no stable per-fill
             # identity for a consumer to dedup repeated tradebook pulls
-            # against.
-            "trade_id": trade.get("trade_id", ""),
+            # against. Emitted under OpenAlgo's own established key
+            # "tradeid" (no underscore) - the documented tradebook contract
+            # (docs/prompt/flow-import-format.md) and existing consumers
+            # (services/telegram_bot_service*.py, broker/groww's own
+            # adapter) already expect that exact key.
+            "tradeid": trade.get("trade_id", ""),
             "timestamp": trade.get("fill_timestamp") or trade.get("order_timestamp", ""),
         }
         transformed_data.append(transformed_trade)
--- a/broker/zerodha/mapping/order_data.py
+++ b/broker/zerodha/mapping/order_data.py
@@ -163,7 +163,26 @@ def transform_tradebook_data(tradebook_data):
             "average_price": trade.get("average_price", 0.0),
             "trade_value": trade.get("quantity", 0) * trade.get("average_price", 0.0),
             "orderid": trade.get("order_id", ""),
-            "timestamp": trade.get("order_timestamp", ""),
+            # Kite's own docs (kite.trade/docs/connect/v3/orders/) document
+            # three separate timestamps on a trade: order_timestamp ("when
+            # the order was registered by the API" - order-placement time,
+            # shared by every fill under one order), exchange_timestamp
+            # ("when the order was registered by the exchange"), and
+            # fill_timestamp ("when the trade was filled at the exchange") -
+            # the one actually correct for a TRADE record. order_timestamp
+            # was used here instead, which is very likely the root cause of
+            # a known bug (AlgoMirror KNOWN_ISSUES.md #2): one account's
+            # Trade Book rows showed time-only values with no date, while an
+            # identical code path on another account showed full datetimes -
+            # order_timestamp is the field genuinely liable to do that for a
+            # multi-fill order, since it's an order-level field being reused
+            # per fill.  "trade_id" is Kite's own per-fill identifier,
+            # previously dropped entirely - only "orderid" survived, shared
+            # by every fill under one order, giving no stable per-fill
+            # identity for a consumer to dedup repeated tradebook pulls
+            # against.
+            "trade_id": trade.get("trade_id", ""),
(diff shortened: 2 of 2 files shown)


arsalanansari17:
Fixed — the identifier is now emitted as `tradeid` (no underscore) for both Zerodha and Kotak, matching the documented tradebook contract and existing consumers (Groww's own adapter, the Telegram bot service, sandbox execution/position managers). Thanks for catching this.

marketcalls (commented):
Recheck score: 96/100 at 3044dc8d4cb0871a7311f6b88ec5b58e9dadc7e9. Merge-ready: the exact two-file diff is clean, exact-head CI metadata is green, official Zerodha/Kotak sources support the mapped fields, and isolated mapper/service tests passed with no security or functional blocker. Two non-blocking cleanups remain: document tradeid in docs/api/account-services/tradebook.md, and correct the Kotak comment because the cited sample rows have different nOrdNo values rather than two fills under one order. No live-account call was performed.

PR #2013, opened 2026-09-06
fix: require upstream WebSocket adapter readiness

linuxsmiths:
## Summary

- Treat the pool-level `connected` flag as necessary but insufficient when adapters connect asynchronously.
- Require every underlying adapter to report readiness before exposing pooled health.
- Delay local WebSocket authentication success until the upstream broker handshake is ready, with a bounded 30-second poll.
- Clean up unready adapters and pools on timeout.
- Include per-adapter connection state in proxy health statistics.
- Add explanatory comments documenting the distinction between pool state and adapter state.

This prevents a locally successful WebSocket authentication from masking an upstream broker rejection such as HTTP 401/403.

## Validation

- `python3 -m py_compile websocket_proxy/broker_factory.py websocket_proxy/connection_manager.py websocket_proxy/server.py`
- `git diff --check`

---
## Summary by cubic
Fixes WebSocket health and authentication so a locally successful login can no longer mask upstream broker rejections like HTTP 401/403.

- Pool health now requires the pool's connection flag and at least one underlying adapter to report actual readiness, using each adapter's `is_connected()` method when available; `FyersWebSocketAdapter` now exposes the underlying Fyers socket state.
- Per-adapter connection state is included in proxy health stats, so partially connected pools remain usable while their individual adapters stay observable.
- Authentication polls up to 30 seconds for adapter readiness, then fails cleanly, removes the client mapping, and cleans up only the adapter owned by the failed attempt on timeout, leaving concurrent attempts untouched.

<sup>Written for commit 82f86240ec601d5d150cfc0a5be5014c7264d5fe. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/2013?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/broker/fyers/streaming/fyers_websocket_adapter.py
+++ b/broker/fyers/streaming/fyers_websocket_adapter.py
@@ -1047,6 +1047,12 @@ class FyersWebSocketAdapter(BaseBrokerWebSocketAdapter):
             self.logger.error(f"Error sending data via ZeroMQ: {e}")
             self.logger.error(f"Data causing error: {data}")
 
+    def is_connected(self) -> bool:
+        """Check the outer adapter and the underlying Fyers socket."""
+        return self.connected and bool(
+            self.fyers_adapter and self.fyers_adapter.is_connected()
+        )
+
     def get_connection_status(self) -> dict[str, Any]:
         """Get connection status"""
         status = {
--- a/websocket_proxy/broker_factory.py
+++ b/websocket_proxy/broker_factory.py
@@ -222,7 +222,7 @@ class _PooledAdapterWrapper:
         # marks itself connected when it starts that work (ref ConnectionPool.connect()),
         # so also inspect the underlying adapters before reusing a cached pool
         # or reporting health.
-        return bool(self._pool.adapters) and all(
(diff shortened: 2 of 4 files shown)


linuxsmiths:
Follow-up concurrency fix: a timed-out authentication attempt now cleans up only when the current user mapping still owns the exact adapter instance it created. If a concurrent authentication has replaced it, adapter and pool cleanup are skipped so the newer ready connection cannot be evicted.

Commit: `fffa56783`.

linuxsmiths:
Follow-up authentication-state fix: when upstream adapter readiness times out, the failed client's `user_mapping` entry is now removed before sending the error. Later messages from that client must authenticate again instead of passing the local authentication guard.

Commit: `$(git rev-parse --short HEAD)`.

PR #2021, opened 2026-09-09
feat(optionsymbol): return the option quote in one batched Dhan request

### LINKED ISSUE #1816: Dhan: quote-path latency is dominated by avoidable rate-limit round trips, not by Dhan's 1 req/s limit
# Dhan: quote-path latency is dominated by avoidable rate-limit round trips, not by Dhan's 1 req/s limit

## Summary

On Dhan, an `optionsymbol` call followed by a quote fetch for the returned legs
measures **~1,125 ms** end to end. The cause is not Dhan's API being slow —
Dhan's non-quote endpoints return in 24–42 ms on the same connection and auth.
It is that the flow spends **two** separately-gated `/v2/marketfeed/quote`
requests where Dhan's API would serve both in **one**.

`DHAN_QUOTE_INTERVAL = 1.1` (`broker/dhan/api/data.py:28`) is **correct and
should not be lowered** — it is compliance with Dhan's documented 1 req/s on
`/v2/marketfeed/*` (error 805). Everything below is about issuing fewer
requests, not faster ones.

## Impact

Any flow that resolves an option symbol and then needs its price pays ~1.1 s on
Dhan, against ~30 ms of actual work. It compounds linearly: each additional
quote-category call adds another full second. [#855](https://github.com/marketcalls/openalgo/issues/855)
is the same defect at four calls, where it stops being slow and becomes an
outright client timeout. Nothing here is Dhan-specific in design — the same
redundant calls exist on every broker; Dhan is simply the one where the cost is
visible.

## Reproduction

On any Dhan-configured OpenAlgo instance:

1. Call `/api/v1/optionsymbol`. If the quote bucket is idle it returns in
   ~25 ms; otherwise it waits out `DHAN_QUOTE_INTERVAL`.
2. Immediately call `/api/v1/quotes` for the symbol just returned.
3. Step 2 always pays the **full 1.1 s**, because step 1 consumed the bucket's
   only slot for that second.

Total ~1,125 ms for a sequence Dhan would serve in one request. Swap the order
and the cost is identical — it is the second call that pays, whichever it is.
`tools/bench_openalgo.py` (n=50/endpoint) produced the tables below.

## Evidence

Dhan REST, n=50/endpoint, collected with `bench_openalgo.py`:

```
endpoint                 p50       p95     err
                    ---- no quote-category call ----
funds                 42.0ms    74.8ms      0
analyzerstatus        11.8ms    13.2ms      0
orderbook             25.9ms    29.3ms      2
tradebook             26.4ms    28.6ms      3
expiry                24.3ms    29.3ms      0
                    ---- reaches /v2/marketfeed/* internally ----
quotes_index        1100.0ms  1106.2ms      0
quotes_option       1099.7ms  1116.3ms      0
multiquotes_2leg    1100.1ms  1116.3ms      0
optionsymbol        1100.1ms  1124.3ms      0
positionbook        1100.0ms  1105.4ms      0
```

These are OpenAlgo API endpoints, not Dhan paths — the split is by whether the
handler ends up making a `/v2/marketfeed/*` call. Every endpoint that does pins
at 1099.7–1100.1 ms; every endpoint that does not is 11–42 ms. Same host, same
auth, same connection pool, so the gate is the only variable.

`positionbook` sitting in the lower group is worth a look on its own: a position
listing pays a full quote slot, presumably to price open positions. That makes
any exit-verification loop cost ≥1.1 s per poll on Dhan.

Concurrency on `multiquotes_2leg` — the global lock serializes perfectly:

```
threads       p50     req/s   err
1        1099.8ms       0.9      0
2        2199.7ms       0.9      0
4        4399.7ms       0.9      0
8        8799.7ms       0.9      0
```

The end-to-end `optionsymbol` + leg-quote sequence reproduces at **~1,122 ms
± 5 ms** across repeated runs, which is the signature of **exactly one full
`DHAN_QUOTE_INTERVAL` wait**: the first quote call finds the bucket idle and
returns in ~25 ms, the second pays the full 1.1 s.

For contrast, the identical sequence against a broker without a quote-side rate
limit completes in ~75 ms, confirming the delay is the limiter and not network,
host, or OpenAlgo overhead.

### 1. `optionsymbol` forces a second gated quote call on the caller — highest impact

`services/option_symbol_service.py:43` imports `get_quotes` and uses it to
resolve the underlying LTP. That is one `/v2/marketfeed/quote` request. The
caller receives an option symbol but no premium, so it must issue a second quote
request for the returned legs — and on Dhan that one pays the full 1.1 s gate.

Per the [Market Quote docs](https://dhanhq.co/docs/v2/market-quote/), Dhan's
`/v2/marketfeed/quote` accepts **up to 1000 instruments per request** — the
limit is per *request*, not per *instrument*. Underlying + CE + PE is three
instruments: **one request, identical cost.**

**Proposed:** an opt-in `include_quotes=true` on `optionsymbol` that returns the
resolved legs' quotes in the same response. Half the plumbing already exists:
`get_option_symbol(..., underlying_ltp=None)` at `:635` already accepts a
pre-fetched LTP "to avoid redundant quote requests", and the response already
carries `underlying_ltp`.

Effect on Dhan: 2 gated calls → 1, ~1,125 ms → ~30 ms. Broker-agnostic win;
Dhan just shows it most dramatically.

### 2. `get_depth` and `get_quotes` issue byte-identical requests, separately gated

`data.py:752` (`get_quotes`) and `data.py:1071` (`get_depth`) both POST to
`/v2/marketfeed/quote` with the same payload shape:

```python
payload = {exchange_type: [int(security_id)]}
```

Same endpoint, same body, two rate-limit slots. A caller wanting both pays
2.2 s for one request's worth of data — and `/quote` already returns depth, so
the second call adds no information.

**Proposed:** a short-TTL (~250 ms) response cache keyed on
`(endpoint, payload)`. Note this cannot serve stale data by construction — the
limiter already enforces ≥1.1 s spacing, so any two calls closer together than
the TTL would have returned the same upstream snapshot anyway.

### 3. Single-symbol quotes don't batch, though the API takes 1000 per request

`data.py:752` sends exactly one security id per request. `get_multiquotes`
(`data.py:945`) already batches correctly — the single-symbol path just doesn't
route through it. N sequential single-symbol quotes cost N × 1.1 s where one
batched call costs 1.1 s.

**Proposed:** a small coalescing window (~50 ms) merging concurrent
single-symbol quote requests into one batched marketfeed call. Largest win for
multi-leg option flows, which are the common case for this endpoint.

### 4. The WebSocket isn't used as a quote source — the structural fix

`broker/dhan/streaming/dhan_adapter.py` keeps no last-tick cache; ticks are
published to ZMQ and discarded. Meanwhile Dhan's live feed WS is the one Dhan
market-data path with **no per-request rate limit** — published limits are 5000
instruments per connection and 5 connections per user — and it benchmarks well:

```
connect 100.33ms   subscribe 0.15ms   first tick 507.16ms
index    gap p50 1001.64ms  p95 1028.84ms   0 stale
optionA  gap p50  992.77ms  p95 1109.47ms   0 stale
optionB  gap p50 1004.94ms  p95 1050.13ms   0 stale
```

**This pattern already exists in OpenAlgo — just at the wrong layer.** The
Python SDK's WebSocket client keeps an in-memory latest-quote-per-`(exchange,
symbol)` cache and exposes it as `client.get_quotes()`, documented as reading
"the snapshot without waiting for the next push". So a WS-backed quote cache is
already an accepted design here; it simply lives client-side and only benefits
SDK users who happen to be streaming, while REST `/quotes` callers pay the gate.

**Proposed:** keep the same last-tick-per-token cache server-side in the broker
adapter and let `get_quotes` serve from it when the symbol is already subscribed
and the tick is fresher than `DHAN_QUOTE_INTERVAL`. For any streaming symbol the
REST gate becomes irrelevant. Falls back to REST unchanged when not subscribed.

### 5. Minor: `/v2/marketfeed/ltp` where only an LTP is needed

ATM resolution (finding 1) needs one number — the underlying LTP — and calls
`/v2/marketfeed/quote`, which returns 5-level bid/ask depth, circuit limits, OI,
volume and trade timings for it.

To be explicit about what this does **not** buy: all three Market Quote
endpoints share the **same** 1 req/s bucket, so `/ltp` does not free a slot.
`data.py:60`'s single `quote` category for the whole `/v2/marketfeed/*` family
is therefore **correct** and should stay. The gain is payload and parse cost
only — real but second-order next to findings 1–4.

## Verified against Dhan's published limits

Per [Dhan's rate-limit documentation](https://docs.dhanhq.co/api/v2/guides/rate-limits)
([support summary](https://dhan.co/support/platforms/dhanhq-api/what-are-the-api-rate-limits-for-dhan/)):

| API category | Per second | Per minute | Per hour | Per day |
|---|--:|--:|--:|--:|
| Non-Trading | 20 | Unlimited | Unlimited | Unlimited |
| Order | 10 | 250 | 1,000 | 7,000 |
| Data (`/v2/charts/*`) | 5 | — | — | 100,000 |
| **Quote (`/v2/marketfeed/*`)** | **1** | **Unlimited** | **Unlimited** | **Unlimited** |

**The quote row is the important one: there is no per-minute, per-hour or
per-day ceiling on quote calls — only the 1/s spacing.** So consolidating two
calls into one is a pure latency win with no quota trade-off. There is no
budget being conserved by making fewer requests, and no downside to a caller
that needs quotes often; the only thing the limiter buys is compliance with the
per-second rule. That also means findings 1–3 cost nothing to adopt.

Per the [Market Quote docs](https://dhanhq.co/docs/v2/market-quote/): **1000
instruments per request**, and the 1 req/s applies uniformly to all three of
`/ltp`, `/ohlc` and `/quote`.

Per the [Live Market Feed docs](https://dhanhq.co/docs/v2/live-market-feed/):
**5000 instruments per connection, 5 connections per user**, 100 instruments per
subscribe message, and **no per-request rate limit**.

`DHAN_DATA_INTERVAL = 0.2` and `DHAN_QUOTE_INTERVAL = 1.1` both match these
exactly. The limiter is right; the request count is not.

## Relationship to #855

[#855](https://github.com/marketcalls/openalgo/issues/855) reports the same root
cause on `/api/v1/syntheticfuture` (4+ sequential gated calls → 8–10 s timeout)
and is still open. This report adds:

- the same defect on a second path, `optionsymbol` (finding 1)
- measurements isolating it to one `DHAN_QUOTE_INTERVAL` wait (~1,122 ms ± 5 ms)
- two defects #855 does not cover: the `get_quotes`/`get_depth` duplicate
  request (finding 2) and the absent WS tick cache (finding 4)
- confirmation that `/ltp` does **not** buy an extra slot, which #855 lists as a
  proposed fix

#855's `underlying_ltp` proposal is the same mechanism as finding 1 and the
parameter already exists at `option_symbol_service.py:635` — it is just not
wired through from `optionsymbol`.

## What is already right

- The per-category limiter design (`data.py:20-50`) is correct: independent
  buckets, timestamp reserved under the lock, sleep performed outside it, and
  both intervals match Dhan's published limits.
- `get_multiquotes` batches properly.
- 0 errors across every quote benchmark — the limiter is doing its job.

The ask is not to relax the limit. It is to stop spending two slots where Dhan
would accept one.

## Suggested order

**If only one lands, make it finding 1** — it is the smallest change, it closes
#855's `underlying_ltp` proposal on a second path, and the parameter it needs
already exists. Finding 2 is next and is close to free. Findings 3 and 4 are
larger and can wait for a maintainer's view on whether a request-coalescing
layer or a WS-backed quote cache is architecturally welcome — happy to open a PR
for any of these if the approach is agreed.

## Environment

- OpenAlgo @ `6bf0db4b0`
- Dhan live, AWS `c8g.large`
- REST benchmarks: n=50/endpoint via `bench_openalgo.py`

wolfgang-aura:
Closes #1816.

When option-symbol lookup is used with Dhan, resolving a strike can issue one quote request for the underlying and another for the option chain. Each request pays the broker's rate-limit wait, so a quote-enabled lookup can be much slower than the work requires.

The new `include_quotes` option is opt-in. For Dhan, the service requests the underlying and candidate option symbols together, reuses that response to resolve the strike, and returns the selected option quote. The path is exposed through REST, MCP, the agent tool, the Flow executor, and the Flow editor. Fyers and Zebu retain their adapter-specific request behavior, while unknown or oversized cases keep the existing fallback path. Default callers keep the existing response and quote behavior.

This follows the batched-multiquote pattern already merged in #1973. Only `frontend/src/` sources are touched; `frontend/dist/` is untouched and left to CI.

### How this was tested

- The maintained test selection passes 243 tests on `9117eb9cd0bf` and 253 with this change, in both cases with 6 existing test warnings and one pytest cache-path warning. Both counts were re-run against the final commit.
- The focused option and adapter-boundary tests passed 159 tests, and the Flow regression selection passed 182 tests.
- The pre-change reproducer made two quote calls for the quote-enabled lookup; the Dhan path now makes one combined multiquote call when the batch fits.
- `npm run build` passes in `frontend` on Node 25.9.0, so the Flow-editor change compiles. The rebuilt `frontend/dist/` was restored and is not part of this branch.
- `git diff --check` passed.

The tests used Windows and the run's Python 3.12.14 environment. The recorded Mailman environment metadata reports Python 3.14.3 because the venv was created from that interpreter. No live Dhan account or broker credentials were available, so the reported live latency was not re-measured locally.

The alternative was to send every broker through one generic batch size and split requests automatically. I did not use that design because Fyers and Zebu perform different upstream work inside their quote adapters. The combined optimization is limited to Dhan, while the other adapters retain their existing request behavior.

This change was drafted with AI assistance. codex (`gpt-5.6-luna`) wrote the patch and codex (`gpt-5.6-luna`) reviewed it under Mailman. The test results above come from the harness executing the commands, not from either agent's account of its own work.

---
## Summary by cubic
Adds an opt-in `include_quotes` option to option-symbol lookup that, on Dhan, resolves the strike and returns the selected option's quote from one batched multiquote call instead of two separate calls that each paid rate-limit latency.

- `include_quotes` is accepted by the REST endpoint, MCP server, agent tool, Flow executor, and the Flow editor.
- The response includes `quote` only when `include_quotes` is true; existing callers see the same response as before.
- Fyers and Zebu keep their current request pattern; unknown brokers or oversized batches fall back to the pre-existing path.

<sup>Written for commit 328394ceb28abc2c79c2524001d80d8d8b157a39. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/2021?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/api/options-services/optionsymbol.md
+++ b/docs/api/options-services/optionsymbol.md
@@ -116,6 +116,7 @@ curl -X POST http://127.0.0.1:5000/api/v1/optionsymbol \
 | expiry_date | Expiry date in DDMMMYY format | Mandatory | - |
 | offset | Strike offset: ATM, ITM1-ITM50, OTM1-OTM50 | Mandatory | - |
 | option_type | Option type: CE or PE | Mandatory | - |
+| include_quotes | Include the selected option's current quote in the response | Optional | false |
 
 ## Response Fields
 
@@ -128,6 +129,7 @@ curl -X POST http://127.0.0.1:5000/api/v1/optionsymbol \
 | tick_size | number | Minimum price movement |
 | freeze_qty | number | Maximum quantity per order |
 | underlying_ltp | number | Current underlying price |
+| quote | object | Selected option quote; returned only when `include_quotes` is true |
 
 ## Understanding Offset
 
@@ -149,6 +151,7 @@ curl -X POST http://127.0.0.1:5000/api/v1/optionsymbol \
 
 - The offset is calculated based on actual **strike intervals** in the database
 - **underlying_ltp** shows the current price used for ATM calculation
+- Set **include_quotes** to `true` when the selected option's quote is needed with the resolved symbol. The underlying and candidate option quotes are fetched in one batch when supported.
 - Use this endpoint to **discover the symbol** before placing orders
 - For placing orders directly with offset, use [OptionsOrder](../order-management/optionsorder.md)
 
--- a/frontend/src/components/flow/panels/ConfigPanel.tsx
+++ b/frontend/src/components/flow/panels/ConfigPanel.tsx
@@ -2299,6 +2299,21 @@ export function ConfigPanel() {
                     ))}
                   </div>
                 </div>
+                <div className="flex items-center justify-between rounded-lg border p-3">
+                  <div className="space-y-1">
+                    <Label htmlFor="option-symbol-include-quotes" className="text-xs">
+                      Include Quotes
+                    </Label>
+                    <p className="text-[10px] text-muted-foreground">
+                      Return the selected option quote with the resolved symbol.
+                    </p>
+                  </div>
+                  <Switch
+                    id="option-symbol-include-quotes"
+                    checked={Boolean(nodeData.includeQuotes)}
+                    onCheckedChange={(v) => handleDataChange('includeQuotes', v)}
+                  />
+                </div>
                 <div className="space-y-2">
                   <Label className="text-xs">Output Variable</Label>
                   <Input
--- a/frontend/src/lib/flow/constants.ts
+++ b/frontend/src/lib/flow/constants.ts
@@ -1538,6 +1538,7 @@ export const DEFAULT_NODE_DATA = {
     expiryDate: '',
     offset: 'ATM',
     optionType: 'CE' as const,
+    includeQuotes: false,
     outputVariable: '',
   },
   expiry: {
--- a/frontend/src/types/flow.ts
+++ b/frontend/src/types/flow.ts
@@ -425,6 +425,7 @@ export interface OptionSymbolNodeData {
   expiryDate: string // Format: 30DEC25 - can use {{variable}}
   offset: string // ATM, ITM1-10, OTM1-10 - can use {{variable}}
   optionType: 'CE' | 'PE'
+  includeQuotes?: boolean
   outputVariable?: string
 }
 
--- a/mcp/mcpserver.py
+++ b/mcp/mcpserver.py
@@ -1353,6 +1353,7 @@ def get_option_symbol(
     offset: str,
     option_type: str,
     expiry_date: str | None = None,
+    include_quotes: bool = False,
 ) -> str:
     """
     Get option symbol for specific strike and expiry.
@@ -1364,9 +1365,11 @@ def get_option_symbol(
         option_type: 'CE' for Call or 'PE' for Put
         expiry_date: Expiry date in 'DDMMMYY' format (e.g., '28OCT25'). Optional when
                      the underlying already includes an expiry.
+        include_quotes: Include the selected option quote in the response.
 
     Returns:
-        JSON with symbol, exchange, lotsize, tick_size, underlying_ltp
+        JSON with symbol, exchange, lotsize, tick_size, underlying_ltp, and the
+        selected option quote when include_quotes is true.
     """
     try:
         params: dict[str, Any] = {
@@ -1377,6 +1380,8 @@ def get_option_symbol(
         }
         if expiry_date is not None:
             params["expiry_date"] = expiry_date
+        if include_quotes:
+            params["include_quotes"] = True
         response = client.optionsymbol(**params)
         return json.dumps(response, indent=2)
     except Exception as e:
--- a/restx_api/data_schemas.py
+++ b/restx_api/data_schemas.py
@@ -149,6 +149,9 @@ class OptionSymbolSchema(Schema):
     option_type = fields.Str(
         required=True, validate=validate.OneOf(["CE", "PE", "ce", "pe"])
     )  # Call or Put option
+    include_quotes = fields.Bool(
+        required=False, load_default=False
+    )  # Include the selected option quote in the response
 
 
 class OptionGreeksSchema(Schema):
--- a/restx_api/option_symbol.py
+++ b/restx_api/option_symbol.py
@@ -15,7 +15,8 @@ Request Body:
     "expiry_date": "28OCT25",  // Optional if underlying includes expiry
     "strike_int": 50,  // Optional: Strike interval. If omitted, actual strikes from database are used (RECOMMENDED)
     "offset": "ITM2",  // ATM, ITM1-ITM50, OTM1-OTM50
-    "option_type": "CE"  // CE or PE
+    "option_type": "CE",  // CE or PE
+    "include_quotes": true  // Optional: include the selected option quote
 }
 
 Response:
@@ -25,7 +26,12 @@ Response:
     "exchange": "NFO",
     "lotsize": 25,
     "tick_size": 0.05,
-    "underlying_ltp": 23587.50
+    "underlying_ltp": 23587.50,
+    "quote": {
+        "ltp": 125.5,
+        "bid": 125.25,
+        "ask": 125.75
+    }
 }
 """
 
@@ -71,6 +77,7 @@ class OptionSymbol(Resource):
             )  # Optional - if not provided, actual strikes from database will be used
             offset = data["offset"]
             option_type = data["option_type"]
+            include_quotes = data.get("include_quotes", False)
 
             logger.info(
                 f"Option symbol request: underlying={underlying}, exchange={exchange}, "
@@ -86,6 +93,7 @@ class OptionSymbol(Resource):
                 offset=offset,
                 option_type=option_type,
                 api_key=api_key,
+                include_quotes=include_quotes,
             )
 
             return response, status_code
--- a/services/agent/tools/options.py
+++ b/services/agent/tools/options.py
@@ -487,6 +487,7 @@ class OptionsToolkit(OpenAlgoToolkit):
         expiry_date: str,
         offset: str,
         option_type: str,
+        include_quotes: bool = False,
     ) -> str:
         """Resolve one tradable option symbol from an offset relative to ATM.
 
@@ -525,6 +526,7 @@ class OptionsToolkit(OpenAlgoToolkit):
                 does not matter. An offset past the end of the listed ladder is
                 an error, not a clamp to the last strike.
             option_type: ``CE`` for a call or ``PE`` for a put.
+            include_quotes: Include the selected option quote in the response.
 
         Returns:
             JSON with ``symbol`` (the OpenAlgo option symbol to trade),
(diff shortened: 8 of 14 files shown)


Kalaiviswa:
Thank you for working on this and for adding comprehensive test coverage. The new include_quotes option is a useful improvement and addresses the main optionsymbol latency path for Dhan when the request fits into one batch.

 However, #1816 covers several additional sources of avoidable quote requests. The following work is still needed for the issue to be considered fully resolved:

 - Update existing consumers to use include_quotes=true; otherwise the current optionsymbol followed by /quotes flow remains unchanged.

 - Consolidate syntheticfuture into one batched request containing the underlying, ATM CE, and ATM PE where possible.

 - Avoid separate Dhan requests from get_quotes() and get_depth(), since both call /v2/marketfeed/quote with equivalent payloads. A short-lived shared response cache could address this.

 - Coalesce concurrent single-symbol quote requests into Dhan’s supported multi-instrument request.
 - Add a server-side latest-tick cache to the Dhan WebSocket adapter, with REST fallback when no fresh tick is available.

 - Optionally use /v2/marketfeed/ltp where only an LTP is required. This reduces payload size, although it shares the same one-request-per-second limit.

 - Add a live-Dhan latency verification to confirm the optimized path performs as expected outside mocked tests.

 Batch failures, missing quote data, and batches exceeding 1,000 instruments can also fall back to multiple rate-limited requests, which is reasonable but should be documented.

Thanks again—this is a valuable first step toward reducing Dhan quote latency.

wolfgang-aura:
Thanks for reading it that closely. Your first point is correct and it is the
one I want to answer properly. The rest I want to separate from this PR.

## The default

`include_quotes` defaults to `false` in all seven places it is threaded
through, so an existing caller that does `optionsymbol` and then `/quotes`
still makes two requests. Opting in was deliberate rather than an oversight:
turning it on by default adds a `quote` object to every `optionsymbol`
response, and spends a quote request on callers that only wanted the symbol.
That is a behaviour change and a cost, and it did not seem mine to make
unasked.

If you want it on by default, say so and I will flip it in this PR. Flipping
it is a small diff. Deciding to flip it is your call, not mine.

## The rest

Everything else on your list is real, and none of it is a change to this
patch. Consolidating `syntheticfuture`, sharing a response cache between
`get_quotes()` and `get_depth()`, coalescing concurrent single-symbol
requests, the WebSocket latest-tick cache with REST fallback, `/v2/marketfeed/ltp`
for LTP-only paths, and live-Dhan latency verification are six separate
changes to five different call paths. Landing them here would make one
reviewable patch into an unreviewable one.

I would rather this merge as what it is, which is the `optionsymbol` batching
step, and leave #1816 open for the rest. If the issue should stay open after
this merges, I am happy to edit the PR description so it says "part of #1816"
instead of closing it.

## Documentation

Fair, and cheap. I will document the fallbacks in
`docs/api/options-services/optionsymbol.md`: a failed batch, a batch response
missing quote data, and a batch over 1,000 instruments all fall back to
separate rate-limited requests, so `include_quotes` bounds the request count
rather than guaranteeing one request. Tell me if you want that in this PR or a
follow-up.

Which of the six, if any, would you want me to pick up next?

PR #1992, opened 2026-09-04
feat(websocket): guaranteed Quote-mode tick field contract

spsp21366-commits:
feat(websocket): guaranteed Quote-mode tick field contract

Adapters forward broker payloads verbatim, and several mappers only add
OHLC when the broker snapshot carries it (zerodha_mapping.py updates
open/high/low/close only 'if available'). Quote-mode clients could not
tell 'field absent' from 'field coming later', and every broker behaved
differently.

- websocket_proxy/tick_contract.py: stdlib-only normalizer guaranteeing
  ltp, open, high, low, close, volume, timestamp on every Quote-mode
  (mode 2) tick. Missing broker values become explicit null - never a
  fabricated 0, which would render as doji bars on OHLC charts.
  Broker-supplied values (including zeros) are passed through verbatim.
  No-op when the contract is already satisfied; idempotent.
- server.py zmq_listener: normalize Quote-mode ticks once before the
  MarketDataService feed and client fan-out.
- docs/websocket-quote-feed.md: correct the data-format examples to the
  actual wire envelope (payload nested under 'data', numeric mode,
  broker tag) and document the guaranteed field set and null semantics.
- test/test_quote_tick_contract.py: standalone unit tests mirroring the
  test_mode_normalization.py pattern, including a guard that ties the
  module's QUOTE_MODE constant to mode_utils.

---

Problem in one line: some adapters ship OHLC in Quote ticks, others are LTP-only; with no documented guarantee, clients must probe per broker or synthesize doji bars. Happy to adjust scope to maintainer preference.

---
## Summary by cubic
Guarantees every Quote-mode tick carries `ltp`, `open`, `high`, `low`, `close`, `volume`, and `timestamp`, including Depth ticks delivered to Quote subscribers on the mode-downgrade path. Missing broker values become explicit `null` (never a fabricated 0 that renders as doji bars), and broker-supplied values — including zeros — pass through untouched.

- Adds a stdlib-only normalizer (`tick_contract.py`) applied before fan-out; it runs on every mode ≥ Quote and is idempotent, so LTP stays lean and Depth→Quote deliveries carry the contract fields.
- Extracts the per-tick parse/normalize/fan-out from `zmq_listener` into `_handle_market_data` so the delivery path is testable without a live socket.
- Renames the module constant to `MODE_QUOTE`, fixing an ImportError when importing `websocket_proxy.server`.
[...]

PR #2046, opened 2026-09-11
fix(websocket): offload market-data dispatch off the shared event loop

AmirF194:
`zmq_listener` (`websocket_proxy/server.py`) shares one asyncio loop with `websockets.serve()`, and it calls `market_data_service.process_market_data()` directly, no `await`. That call fans out to `WebSocketExecutionEngine._on_market_data`/`_check_and_execute_order`, which does synchronous SQLAlchemy commits and margin reconciliation. A tick that triggers an order blocks the loop for that DB work, and a concurrent client's handshake to the same proxy times out, matching the `TimeoutError: timed out during opening handshake` in the issue.

Wrapped the dispatch in `await asyncio.to_thread(...)`. zmq_listener still awaits it, so ticks are still processed one at a time in order, but the loop is free to accept connections while the thread runs.

Went with your first suggestion but at the outer call site (`process_market_data`) rather than the inner `_check_and_execute_order`: it's the one place all of `_broadcast_update_priority`'s CRITICAL-priority synchronous work already funnels through (today just the sandbox engine, but not only ever that), and it's already wrapped in its own try/except. Didn't touch the subscription-filtering or separate-loop ideas, those are bigger changes.

Added `test/test_zmq_listener_offloads_market_data.py`: builds a bare `WebSocketProxy` (no real ZMQ/sockets), feeds it one tick whose `process_market_data` blocks for 0.3s, and counts how many times a concurrent heartbeat coroutine gets to run during that window. Without the fix the heartbeat never runs, the loop is stalled for the full 0.3s. A second test checks the offloaded call still receives the right parsed fields. Also added the file to `ci.yml`'s test list, it wasn't picking up `test/` by default.

Ran the repo's CI-safe test list plus this new file on Python 3.12 and 3.14 (the matrix extremes), all green except `test_precompress_assets.py`'s stale-variant test, which I confirmed flakes on its own, unrelated to this change. `ruff check` clean on both changed files; `ruff format --check` flags pre-existing formatting elsewhere in server.py, not on the lines I touched.

One thing I noticed but didn't check: #1992 is open on this same file and pulls this exact block into a new `_handle_market_data` method. It still calls `process_market_data` synchronously there, so this bug would survive it landing first, just at a different line, might need a rebase depending on merge order.

---
## Summary by cubic
Fixes WebSocket handshake timeouts by offloading market-data dispatch from the shared event loop. `zmq_listener` previously ran `process_market_data` synchronously, and that call could stall the loop with SQLAlchemy commits and margin reconciliation, blocking concurrent handshakes. The dispatch now runs in a thread via `await asyncio.to_thread`, still awaited so ticks stay ordered, but the loop remains free to accept connections.

**Notes**
- Adds regression tests for the offload and parsed-field delivery, and registers them in `ci.yml`.
- Touches the same block as #1992; a rebase may be needed depending on merge order.

<sup>Written for commit f25536f59174833b2e607bbc14e3150ef3000e5a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/2046?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,6 +65,7 @@ jobs:
             test/test_broker_credential_logging.py \
             test/test_option_symbol_service_validation.py \
             test/test_graceful_shutdown.py \
+            test/test_zmq_listener_offloads_market_data.py \
             -v --timeout=60
 
   # Frontend linting
--- a/test/test_zmq_listener_offloads_market_data.py
+++ b/test/test_zmq_listener_offloads_market_data.py
@@ -1,0 +1,109 @@
+"""Regression for #2019: zmq_listener must not run MarketDataService work inline
+on the event loop shared with websockets.serve()'s handshake accept path.
+
+Self-contained, like test_websocket_unsubscribe_contract.py: no real ZMQ
+context, socket or broker adapter.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import json
+import time
+from collections import defaultdict
+
+from websocket_proxy import server as proxy_server
+from websocket_proxy.server import WebSocketProxy
+
+
+class _OneTickThenIdleSocket:
+    """Stands in for the real zmq.asyncio socket: yields one market-data
+    message, then blocks (like a real socket with nothing more to deliver)
+    until the test tears the loop down.
+    """
+
+    def __init__(self, topic: bytes, payload: bytes) -> None:
+        self._delivered = False
+        self._topic = topic
+        self._payload = payload
+
+    async def recv_multipart(self):
+        if not self._delivered:
+            self._delivered = True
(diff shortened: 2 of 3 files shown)


AmirF194:
CI's clean at head 22e71a02 and it's been a week with no review. Let me know if anything's unclear about the fix.

PR #2058, opened 2026-09-14
fix(dhan_sandbox): build history fallback candles instead of crashing

### LINKED ISSUE #2057: dhan_sandbox: get_history crashes on every interval when the sandbox returns no candles
## Summary

On `dhan_sandbox`, `/api/v1/history` fails on **every interval** whenever the sandbox returns no candles, which the code expects it to do often: it has a placeholder-candle fallback for exactly that case. The fallback itself crashes, so the caller gets an error instead of placeholder data.

```
Error fetching historical data: unsupported operand type(s) for //: 'int' and 'str'   # 1m, 5m, 15m, 25m, 1h
Error fetching historical data: unsupported operand type(s) for +: 'int' and 'str'    # D
```

## Cause

`broker/dhan_sandbox/api/data.py`, the fallback in `get_history` (current `main`):

```python
interval_seconds = self.timeframe_map.get(interval, 300)                # line 678
...
trading_session_seconds = 6 * 60 + 15  # 375 minutes                    # line 692
num_candles = min(trading_session_seconds // interval_seconds, 75)      # line 693
...
candle_ts = base_ts + (i * interval_seconds)                            # line 701
```

`timeframe_map` holds Dhan's request codes as **strings**: `"5m": "5"`, `"1h": "60"`, `"D": "D"`. So:

- **Intraday:** `375 // "5"` raises at line 693.
- **Daily:** line 693 is skipped, but `i * "D"` is a string, and `base_ts + ""` raises at line 701.

There's also a units bug that the type error currently hides. `"5"` means **minutes**, but line 701 adds it to a Unix timestamp as **seconds**, so even after an `int()` conversion the candles would be spaced 5 seconds apart. `trading_session_seconds` is 375 minutes, despite its name and the comment above it claiming 22500 seconds.

## Reproduce

No Dhan session needed. Replace the HTTP call with one that returns no candles:

```python
import broker.dhan_sandbox.api.data as d

d.get_token = lambda symbol, exchange: "2885"
d.get_api_response = lambda *a, **k: {}      # sandbox returns no candles
bd = d.BrokerData("token")
bd.get_quotes = lambda *a, **k: {"ltp": 0}

for interval in ("5m", "15m", "1h", "D"):
    try:
        bd.get_history("RELIANCE", "NSE", interval, "2026-09-08", "2026-09-10")
        print(interval, "OK")
    except Exception as e:
        print(interval, "FAIL", e)
```

```
  5m: FAIL Error fetching historical data: unsupported operand type(s) for //: 'int' and 'str'
 15m: FAIL Error fetching historical data: unsupported operand type(s) for //: 'int' and 'str'
  1h: FAIL Error fetching historical data: unsupported operand type(s) for //: 'int' and 'str'
   D: FAIL Error fetching historical data: unsupported operand type(s) for +: 'int' and 'str'
```

The same `TypeError` shows up in a running `dhan_sandbox` instance's `log/errors.jsonl`, raised from `services/history_service.py` on `POST /api/v1/history`.

## Suggested fix

Convert the interval to seconds in the fallback: minutes × 60 for intraday, one day for `D`. Compute the session length in seconds (22500) so the candle count and spacing use the same unit. I'll open a PR with this and a regression test.

vibecoding-skills:
## What this PR does

Closes #2057.

When the Dhan sandbox returns no candles, `get_history` in `broker/dhan_sandbox/api/data.py` generates placeholder candles so charts and `/api/v1/history` still have data. That fallback crashed on **every** interval, so callers got an error instead. There are three separate bugs, each hidden behind the one before it.

## The three bugs

**1. The interval was a string.** The fallback took its step from `timeframe_map`, which holds Dhan's request codes as strings (`"5"` for 5m, `"D"` for daily).

```
intraday: 375 // "5"        -> TypeError: unsupported operand type(s) for //: 'int' and 'str'
daily:    base_ts + i * "D" -> TypeError: unsupported operand type(s) for +: 'int' and 'str'
```

**2. The units were wrong.** Those codes are minutes, but the candle timestamps added them as seconds. `trading_session_seconds` was 375, which is minutes, despite its name and a comment claiming 22500 seconds. Converting the code with `int()` alone would have spaced 5m candles five seconds apart.

**3. The list was never created.** The loop appends to `fake_candles`, but nothing in the file creates it. `fake_candles = []` went in with the fallback in `e090d255d` (25 Feb) and was deleted the same day in `4a15de87a`. It has not been in the file since. Fixing bugs 1 and 2 alone just moves the crash to `NameError: name 'fake_candles' is not defined`.

## The change

- Convert the interval to real seconds: minutes × 60 for intraday, one day for `D`.
- Measure the session in seconds: `(6 * 60 + 15) * 60 = 22500`.
- Create `fake_candles = []` before the loop.

The path taken when the sandbox *does* return candles is untouched.

## Testing

New `test/test_dhan_sandbox_history_fallback.py`. It replaces the HTTP call with one that returns no candles, so no Dhan session is needed. For every supported interval it checks the candle count and that candles are spaced by the interval in seconds (`1m`→60, `5m`→300, `15m`→900, `25m`→1500, `1h`→3600, `D`→86400). It also checks that intraday placeholders stay inside the 9:15 to 3:30 session.

```
uv run pytest test/test_dhan_sandbox_history_fallback.py -q
7 passed
```

All 7 fail on current `main`.

CI's fast backend selection plus `test_dhan_tick_size.py` and the new file: **73 passed**.

Ruff had been reporting bug 3 all along. `ruff check` on `broker/dhan_sandbox/api/data.py` drops from 10 findings to 8: both `F821 Undefined name 'fake_candles'` are cleared, and nothing new appears. The test file passes `ruff check` and `ruff format --check`.

---
## Summary by cubic
Fixes the Dhan sandbox history fallback so placeholder candles render instead of crashing on every interval when the sandbox returns no data. Closes #2057.

- The interval came from `timeframe_map` as a Dhan request code (`"5"` for 5m, `"D"` for daily), which raised `TypeError` on both the intraday and daily paths.
- Those codes are minutes, not seconds, so even after fixing the type, 5m candles would have been spaced five seconds apart; the fallback now converts minutes × 60 or uses one day for `D`.
- The loop appended to `fake_candles`, which was never created, so the fallback crashed with `NameError` once the type error was fixed; the list is created before the loop now.
- Adds `test/test_dhan_sandbox_history_fallback.py`, which mocks the empty HTTP response and verifies candle count and spacing for every supported interval.

<sup>Written for commit 621efcf6a5d9f1bc9273b23e0977c3c31c630747. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/2058?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Kalaiviswa:
Thanks for this PR, and for the careful write-up in #2057. Your diagnosis was right on all three counts: the string interval codes, minutes being used as seconds, and fake_candles never being initialised. It made the problem easy to act on.

This is now fixed in #2114, and I'm closing this PR without merging. Here's why we went a different way.

Your change makes the fallback work. We decided to remove the fallback altogether. No other broker in OpenAlgo makes up candles when history Angel, Fyers and Kotak all return anempty DataFrame with the standard columns, and some brokers have comments saying they deliberately avoid
fake bars. Even once the crash was fixould have:

- been dated today from 9:15, whatever asked for, and
- had made-up prices, so a chart, backtest or strategy running on the sandbox would treat invented data as
  real, with nothing to tell them apar

In #2114, an empty sandbox response no the columns timestamp, open, high,low, close, volume, oi, the same as live Dhan. That removes the crash in #2057, since the code that
crashed is gone.

Your test fixture, which fakes the HTTut a Dhan login, was a good pattern,and we used the same approach to test the new behaviour.

Thanks again for reporting and digging into it. Contributions like this are welcome.

PR #2063, opened 2026-09-14
fix(smartorder): stop reporting a skipped exit as a placed order

### LINKED ISSUE #2054: [Bug]: Close/action button clickable when no positions are open, firing false 'placesmartorder' alerts- whatsapp/telegram
That warning confirms the core bug: the order execution engine correctly detected that there were no open positions and aborted the order, but the notification engine fired anyway.

In OpenAlgo's backend, the flow should be:

Check open position → None found → Log warning → Stop.

Place order → Broker responds → Trigger alert.

What is happening instead is an inverted control flow or decoupled event trigger:

The API route/endpoint for placesmartorder likely triggers the Telegram and WhatsApp notification services before or independently of the position check result in the order placement adapter.

When you click the red X position window, the alert handler fires immediately (logging Telegram alert triggered for placesmartorder and queueing WhatsApp), while downstream the order execution service checks the state, logs No OpenPosition Found. Not placing Exit order., and halts

"The terminal explicitly logs No OpenPosition Found. Not placing Exit order., confirming the backend knows no position exists. However, the alert services (telegram_alert_service and whatsapp_alert_service) fire before this check completes, sending false 'order executed' notifications for an order that was intentionally skipped."

anishkun:
Closes #2054.

## What is actually happening

The report suspected an inverted control flow, with the alert firing before the position check completed. It is not that. Both channels subscribe to a dedicated `order.no_action` event, `send_order_alert()` is handed the broker response, and it runs after the adapter returns. The alerts were given the wrong event, and would have mislabelled the right one anyway.

A smart order targets a position size, so placing nothing is a correct outcome. Every broker adapter returns that from one branch, with no API call made and therefore no order id:

```python
elif position_size == current_position:
    if int(data["quantity"]) == 0:
        response = {"status": "success", "message": "No OpenPosition Found. Not placing Exit order."}
    else:
        response = {"status": "success", "message": "No action needed. Position size matches current position"}
    orderid = None
    return res, response, orderid   # res remains None as no API call was made
```

`services/sandbox_service.py` produces the same pair. `place_smart_order_service` matched only the second, so clicking the red X with no position open took the first and fell through:

- **Analyze mode**: `"No action" in message` is False, so it published `OrderPlacedEvent`. Telegram and WhatsApp rendered the `placesmartorder` template, which is `*Smart Order Placed*`, and appended `Order ID: {response.get('orderid', 'N/A')}`. The user was told an order went in, with `N/A` where the id should be, and the `message` explaining why nothing happened was never rendered. This is the false alert in the report.
- **Live mode**: the guard missed, `res` is None so the `res.status == 200` check fell through, and the else branch published `OrderFailedEvent` and returned **HTTP 500**. A benign no-op reached the user as an error.

`frontend/src/hooks/useSocket.ts` already treats **both** messages as informational, so the frontend and the backend disagreed about what a no-action smart order looks like. In live mode the `status === 'error'` branch is tested first, so the toast read `Error: No OpenPosition Found...` and the informational branch written for it was unreachable.

## The change

**Read the shape, not the wording.** The adapters do not share a wording ("No action needed", "No action required", "No position to square off", "No order action needed", "Position already matches target size of N", a lower-case upstox variant, and tradejini sends no message at all), and an earlier revision of this PR that matched a marker list still missed six of them. What they do share is the shape: an adapter that placed nothing made no API call, so it returns `res` None and no order id; an adapter that placed an order returns the response object and the id it parsed, which is what the service already gates `OrderPlacedEvent` on. `is_no_action_response(response, order_id)` checks success and the absence of an id, and the live guard is `res is None and is_no_action_response(...)`. The `res is None` half keeps the lost-id case honest: a response object with no parsed id is an order that went in, and still reaches `order.placed`.

**One predicate, shared.** The sandbox has the same shape (a placed sandbox order carries `orderid`, a no-action result does not), so the analyze path uses the same call. Folding the analyze check into it also fixes its precedence: written inline, the `and` bound tighter than the trailing `or`, so `{"status": "error", "message": "... Already Matched ..."}` satisfied the whole condition.

**The alert says what happened.** A smart order that placed nothing now uses a `placesmartorder_no_action` heading, `*Smart Order - No Action Taken*`, and reports the adapter's reason instead of `Order ID: N/A`. Only a success with a reason and no order id takes that path, so an order that did go in is untouched. Both channels changed identically, keeping them symmetric as the files intend.

**The live no-op is a success.** It returns 200 with the adapter's own wording, so the message names the real cause, with a fixed fallback ("No action needed. Position already at the requested size.") only when the adapter sent none. Nothing downstream matches on the live wording: `order_notification` goes out with status `info`, and the one text match left in the tree is the analyzer toast in `useSocket.ts`, which reads the sandbox's two fixed strings and is pinned to them both ways by a test.

One incidental fix worth flagging: the `SmartOrderNoActionEvent` message carried a stray leading space that reached the socketio toast verbatim. It is gone.

## Verification

```
uv run pytest test/test_smart_order_no_action.py -v
```

49 passed. The suite drives `place_smart_order_with_auth` with a stub adapter module and a fake event bus, parametrised over every distinct `(res, response, orderid)` triple a shipped adapter returns with nothing placed (13, collected by walking the AST of every smart-order function under `broker/`; a sweep test with no keyword filter reports any success wording the fixture does not exercise), plus the placed-order, lost-id and rejected shapes, and both sandbox wordings in analyze mode. Each asserts the result tuple and the exact topic sequence on the bus. No broker session, no running server, no market hours.

Against `origin/main`'s service, 15 of the 21 behavioural tests fail: every live no-action shape and the analyze no-position case. Against this PR's earlier marker-list revision (ab4687c), exactly six fail: definedge x2, groww, ibulls, upstox, tradejini.

The file is added to the `backend-test` list in `ci.yml`, since that job names its test files explicitly. Full CI backend list on this branch: **344 passed**. `ruff check` clean; the three service files were already outside `ruff format` on `main`, so they are left as they are.

I have no Telegram or WhatsApp account linked to a live instance, so the rendered text is asserted from the template and formatter rather than observed in a chat. Worth one real no-action smart order before merging.

There is a second, narrower half of the report - the close button being clickable at all when no positions are open - which is a frontend guard rather than this path. Happy to take that separately if you want it.


anishkun:
Self-review before this goes near a merge. Two changes in the follow-up commit.

**`_placed_nothing()` could mislabel a live order.** It tested only for a missing order id. An adapter that returns 200 but whose order id failed to parse produces that same shape, and would have been announced as "Smart Order - No Action Taken / No order was placed" for an order that did go in. A false statement about a live order is worse than the vague line it replaced.

The two shapes are actually distinct: a placed order carries an id and no message, a no-action result carries a message and no id. Both halves are now required, and anything else keeps the old heading and the old `Order ID: N/A`, saying no more than it used to.

**One test was too broad.** `test_frontend_and_backend_agree_on_what_no_action_looks_like` scanned every `message.includes` in `useSocket.ts`. An unrelated toast added anywhere in that file would have failed this test and pointed the blame at the backend. It now reads only the placesmartorder branch.

Test count is now 20, not the 18 quoted above; the two added fail on my previous commit.

anishkun:
Third finding from self-review, and the worst of them.

My marker list was built from the wording in the adapter this issue came from, `"No action needed"`. The wording is not uniform. `broker/definedge/api/order_api.py:388` returns

```python
response_data = {"status": "success", "message": "No action required"}
```

which matches no "needed" variant. The check I replaced was `"No action" in message`, so narrowing it to `"No action needed"` would have put definedge straight back on the defect this PR exists to fix: in analyze mode that response would once again publish `OrderPlacedEvent` and announce an order that was never sent. A fix that regresses one broker while fixing another is worse than no fix.

Markers are now the shortest distinguishing fragments matched case-insensitively, which also picks up "Position already matched." (zerodha, hdfcsky, arrow), "Position already aligned" (iiflcapital) and the sandbox's capitalised "No Action needed."

Two tests added. One pins every wording a shipped adapter actually returns, collected by grepping `broker/` rather than read off a single adapter. The other sweeps every `place_smartorder_api` in the tree for a success message no marker covers, so a new adapter wording fails at the point it is added. Both fail on my previous commit. Now 26 tests.

Unrelated, found while running the CI list: `test/test_precompress_assets.py` is flaky. It fails a different test on most runs and passes cleanly on others, because it writes `os.urandom(4096)` and then asserts on whether compression still helps. It is in the `backend-test` list, so it can fail CI for anyone. Happy to open a separate issue or a fix if useful.

Kalaiviswa:
Thanks for this, and for the write-up — the diagnosis is right. I verified locally: the tests fail on main and pass with the patch, and with a stubbed adapter the live path goes from order.failed + HTTP 500 to order.no_action + 200, while analyze mode goes from "Smart Order Placed / Order ID: N/A" to "Smart Order - No Action Taken".

A few things before merge:

1. Three adapters still miss the markers and stay on the old path: definedge returns "No position to square off" (order_api.py:344, reached by exactly the position_size == 0, current_position == 0 case in the issue) and "Position already at target size" (:349); groww returns "No order action needed. ..." (:2101, so "no action" does not match); tradejini returns a success with no message at all (:1128).
2. test_no_success_no_action_message catch those — it pre-filters on "noaction" / "openposition" before the coverage check, which is the same assumption as the markers, so ireports no misses today.
3. Related: the module comment lists definedge's wording as "No action required", but that literal isits default for missing parameters,
4. Would you consider one behavioural test that calls place_smart_order_with_auth with a stub broker module and a fake bus, and asserts the suite asserts source text, so itpins the shape of the patch rather than the routing.

Longer term the string matching is the fragile part — a structured flag on the adapter response would close the class. Happy to take that as a follow-up rather than hold this PR.

anishkun:
Thanks for the careful read - all four are right, and the walk you prompted found two more misses beyond the four you named: ibulls ("Position already matches target size of N", so "already matched" did not match "matches") and upstox ("No open position found. Not placing exit order.", lower-case and with a space, so "no openposition found" missed it). Six adapters in all were still on the 500 path. That is the marker list failing the way you said it would, so d84666b drops it rather than extending it.

**Shape instead of wording.** Every adapter that placed nothing made no API call, so it returns `res` None and no order id; every adapter that placed an order returns the response object and the id it parsed (the service already gates `OrderPlacedEvent` on `res.status == 200`). The live guard is now `res is None and is_no_action_response(response_data, order_id)`, where the predicate checks success and the absence of an id, on both the tuple and the dict. That takes tradejini's `{"status": "success", "orderid": ""}` with no message at all. The sandbox has the same shape (a placed sandbox order carries `orderid`, a no-action result does not), so the analyze path uses the same predicate. This is your structured flag without touching thirty adapters: the tuple already carries it.

The `res is None` half is what keeps the lost-id case honest: a response object with no parsed id is an order that went in, and still goes to `order.placed`.

**The live message** is now the adapter's own wording, passed through, with a fixed fallback ("No action needed. Position already at the requested size.") only when the adapter sent none. Nothing downstream matches on the live wording: the socket path emits `order_notification` with status `info`, the alerts print the message under the No Action Taken heading, and the only text match left in the tree is the analyzer toast in `useSocket.ts`, which reads the sandbox's two fixed strings. That contract is pinned both ways in `test_frontend_and_sandbox_agree_on_what_no_action_looks_like`.

**Tests.** As you asked, the suite now drives `place_smart_order_with_auth` with a stub adapter module and a fake bus. It is parametrised over every distinct `(res, response, orderid)` triple a shipped adapter returns with nothing placed (13 of them, including tradejini's), plus the placed-order, lost-id and rejected shapes, and the two sandbox wordings in analyze mode. Each asserts the result tuple and the exact topic sequence on the bus. The source-text tests are gone. The tree sweep walks the AST of every `place_smartorder_api` and `_place_smartorder_locked*` with no keyword filter and reports any success message the fixture does not exercise, which is what point 2 needed.

On point 3, small correction for the record: the `:388` "No action required" is the success-status fallback at the end of the function (the else of `if action and quantity > 0`); the missing-parameters default is `:300`, which carries status error and stays a rejection. It no longer matters, since no wording is matched.

Verification: 49 tests pass on this branch. Against `origin/main`'s service, 15 of the 21 behavioural tests fail (every live no-action shape, and the analyze no-position case). Against the previous commit ab4687c, the six that fail are exactly definedge x2, groww, ibulls, upstox and tradejini. Full CI backend list: 344 passed. `ruff check` clean; `place_smart_order_service.py` was already outside `ruff format` on main and is left that way.

PR #2064, opened 2026-09-14
fix(positions): no Close button on a position already squared off

anishkun:
The frontend half of #2054. The backend half is #2063; the two touch different files and can land in either order.

## The defect

A broker keeps reporting a position after it has been squared off: quantity 0, realised P&L attached. `/positions` shows those rows, correctly, because that figure is the whole reason they are still reported. It also rendered the red X on them, so the page offered to close a position that is not there. Clicking it sent a `placesmartorder` the engine then refused, logging:

```
No OpenPosition Found. Not placing Exit order.
```

An action was offered for something that cannot happen, and the failed attempt is what surfaced the alert defect in #2054.

## Why this shape of fix

The rule is already settled twice in this codebase, and this table is the one place that missed it.

The same page, in its own header:

```tsx
<Button variant="destructive" size="sm" disabled={stats.total === 0}>
```

`stats.total` counts only non-zero-quantity rows, with a comment saying so. So Close All already knows the difference between a row and an open position.

And `components/trading/dock/PositionsTable.tsx` states it outright in its header comment, then implements it with `{open && <Button .../>}`:

> the zero-quantity rows the broker still reports stay visible with their realised figure and no Close

This change does the same thing in `pages/Positions.tsx`: the row and its P&L stay, only the action goes. Omitting the button rather than disabling it is what the dock does, so the two position tables now behave identically.

One thing found while making the change: the quantity colour was

```tsx
position.quantity > 0 ? 'text-green-600' : 'text-red-600'
```

so a squared-off position rendered its `0` in the short colour. It is now muted, which is neither long nor short.

## Verification

```
cd frontend && npx vitest run src/pages/Positions.test.tsx
```

7 passed. Three fail on `main` - the two that assert the button is gone and the one that asserts the quantity is not red. The other four are guards that must pass either way: the button still appears on an open row, the closed row and its P&L are still rendered, and Close All is still disabled when everything is flat.

Frontend tests are auto-discovered, so nothing needed adding to the CI list.

`npm run build` (which typechecks) and `npm run lint` both pass. Full suite: 1961 passed. `frontend/dist/` is untouched.

Biome wanted to reformat two unrelated pre-existing spots in this file, the `EmptyState` import position and the `action={...}` block below it. I reverted both so the diff is only this change. The row block is reindented because the map body gained a `const` and an explicit `return`.

---
## Summary by cubic
Fixes the positions table offering a Close button on already squared-off positions, so it no longer sends an order the engine refuses. Zero-quantity rows keep their realised P&L but show no Close action, and their quantity is muted instead of red.

The quantity is coerced to a number before the check, since Zerodha sends the string `"0"` for closed positions, which a strict comparison reads as open and puts the button back.

Addresses the frontend half of #2054.

<sup>Written for commit 4376560f5842b35dda75bd7c9d1a0a02b358a63e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/2064?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


anishkun:
Self-review before this goes near a merge. Found a defect in my own change and fixed it in the follow-up commit.

The guard read `(position.quantity || 0) !== 0`. The quantity is typed `number` but does not always arrive as one: `broker/zerodha/mapping/order_data.py` defaults it to a string.

```python
"quantity": position.get("quantity", "0"),
```

`"0" !== 0` is true, so a squared-off Zerodha position was read as open and got its Close button back, which is the exact row this PR exists to clear. It was invisible on screen because the colour ternary coerces implicitly and stayed correct.

Now coerced once at the top of the row with `Number(position.quantity) || 0`, as `calculatePnlPercent` in this same file already does, and both the guard and the colour read that value.

Two cases added, a string zero and a string quantity. The first fails on my previous commit. Test count is now 9, not the 7 quoted above.

anishkun:
Correcting my own note above, in the direction of the defect being wider than I said.

I cited Zerodha, where `"quantity": position.get("quantity", "0")` only yields a string when Kite omits the field, which is rare. The common case is the Noren-family adapters, and there the string is what the broker always sends.

`broker/flattrade/mapping/order_data.py:399` and `broker/tradesmart/mapping/order_data.py:265`:

```python
"quantity": position.get("netqty", 0),   # raw Noren value
```

Noren returns its numeric fields as strings, so a squared-off position arrives as `"netqty": "0"` and is passed straight through. `broker/shoonya/mapping/order_data.py:311` is the same API and does coerce, `netqty = float(position.get("netqty", 0))`, which is the tell: one adapter in the family converts and two do not.

So under the strict `!== 0` I first wrote, **every closed position on Flattrade and Tradesmart would have kept its Close button** - not an edge case reachable only through a rare default, but the normal path for those two brokers. `Number(position.quantity) || 0` covers all of them.

Two adjacent observations, both outside this PR:

- `frontend/src/hooks/useLivePrice.ts:255` has the same assumption: `if (item.quantity !== undefined && qty === 0)` is false for `"0"`, so the closed-position branch that keeps realised P&L is skipped for those brokers. The hook spreads `{...item}` and never normalises `quantity`, which is how the raw string reaches the table in the first place.
- The four adapters disagree about the type they emit for the same field. Normalising quantity once in the position mapping, rather than at each reader, would remove the class. Happy to open either as a separate issue if you want it.

PR #2071, opened 2026-09-15
feat(oi-profile): chart overlay, multi-expiry sums and a shared answer cache

aakash-code:
## OI Profile: on the chart, across expiries, and a lot cheaper

OI Profile was one page at `/tools`. It is now also an overlay on the `/trading`
chart, it can sum several expiries the way Sensibull's does, and a single answer
serves every client that asks for it instead of each one paying the broker again.

OI Profile on the /trading chart

Calls sit above each strike line and puts below it. The solid fill is current open
interest; the dashed box is what that strike carried into the day, so the gap
between the two edges is the day's build or unwind without switching views.

### Service and API

- **`expiry_dates`** sums open interest per strike across up to four expiries. The
  first expiry drives spot, ATM and the futures leg; the rest only add OI.
- **`strike_count`**, **`include_change`** and **`include_candles`** let a caller ask
  for only what it draws. The chart overlay already has bars, and the change pass
  costs one broker history call per option leg.
- **Shared answer cache**, bounded on size and TTL, keyed by everything that defines
  the payload. A second chart on the same underlying costs nothing.
- **Previous-session OI is cached per leg for the session.** It is settled history and
  cannot change again until tomorrow, yet reading it is one history call per leg
  behind the process-wide 350 ms gate.
- **`market_open`** in the payload lets a live overlay stop asking for a number that
  cannot move until the next session.

### Bug fix: every strike reported as unchanged

Outside market hours the broker appends a daily candle for the new date carrying the
last quote, so the newest two rows are identical. Taking `candles[-2]` as "yesterday"
then compared a session against itself and reported zero change everywhere. Identical
trailing rows are now dropped before picking the previous session. Covered by three
tests, including the live-session case where `[-2]` is genuinely correct.

### Chart overlay

`strategies/indicators/oi_profile_live.js` now ships as the second reference
indicator (Open Range Breakout is the pure-`calc` case; this is the other half of the
surface: data the chart has never seen, fetched in `attach`, drawn by its own canvas
primitive).

- Previous-session outline, max pain marker, zero rule in change mode, and either
  colour convention (calls red like Sensibull and Upstox, or calls green like the
  `/oiprofile` page).
- Strike window of 5 / 10 / 25 / 50 around ATM — the setting that decides how much
  work a refresh is.
- Paints the fast open-interest answer first and upgrades it when the slower change
  answer lands, rather than leaving the chart empty while it runs.
- Jittered three-minute beat, one request in flight at a time, newest request wins,
  and nothing at all while the tab is hidden or the market is closed.

### Chart fix: "needs more history than the N bars loaded"

`warnIfStarved` treated an indicator whose plots are all hidden as starved. Anything
drawing through `draws()` or its own primitive has to declare a hidden placeholder
plot and return nulls, so every such indicator was reported as having nothing to
draw. It now skips the warning when every declared plot is hidden, and still warns
when only some are.

### Testing

- `test/test_oi_profile_multi_expiry.py` — 9 tests: multi-expiry sums, cache hit
  without touching the broker, both skip flags, `strike_count` plumbing and cache
  keying, and the previous-session rule.
- `frontend/src/lib/trading/indicatorStarvationWarning.test.ts` — 4 tests around the
  hidden-plot guard.
- Verified against a live Upstox session: two expiries summed (ATM CE 6,102,655 →
  6,897,605), cold request 0.74 s and warm request 0.000 s from the shared cache,
  and the overlay rendering on `/trading`.

### Not included

The WebSocket path. Upstox Chart 360 carries the whole chain over its market data
feed and aggregates in the browser, with no OI endpoint at all. OpenAlgo already
streams option quotes with OI through the proxy on 8765, so the same shape is
available to us — but it is a larger change than this one and belongs in its own PR.

Generated with [Claude Code](https://claude.com/claude-code)

---
## Summary by cubic
Makes OI Profile available as a chart overlay on `/trading`, lets it sum open interest across up to four expiries, and serves shared cached answers instead of each client paying the broker again.

**Service and API**
- `expiry_dates` sums OI per strike across multiple expiries; the first expiry drives spot, ATM, and the futures leg.
- `strike_count`, `include_change`, and `include_candles` let a caller request only what it draws.
- Existing single-expiry requests keep working unchanged.
- Repeat requests are served from a bounded TTL cache, so a second chart on the same underlying never reaches the broker.
- Previous-session OI is cached per option leg for the session.

**Chart overlay and fixes**
- The overlay exposes calls above strikes and puts below, with previous-session outlines, max pain, and configurable strike windows.
- It paints the fast open-interest answer first and upgrades it when the slower change answer arrives.
- Fixes the after-hours bug where every strike reported zero change because identical trailing candles were compared against each other.
- Fixes the starvation warning firing for indicators whose plots are all hidden.

<sup>Written for commit 2ead1f5fa307eebcddaca621ce306efdf099cee4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/2071?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


aakash-code:
Closing in favour of #2073, which supersedes this.

The overlay had not been through a live session when this was opened. It has now, and the session found four things worth fixing, one of them wrong rather than rough: Change in OI was anchored on the wrong bar. Checked against NSE's own option chain on 16-Sep-2026, NIFTY22SEP2623200PE carried 189,110 contracts with 119,065 of change, so NSE measures from 70,045 contracts - 4,552,925 at that leg's lot size of 65, which is the previous daily candle exactly. It reported 85% where NSE and Sensibull both reported 170%.

#2073 carries everything here plus that fix, rebased on current main, verified leg by leg against NSE for NIFTY, BANKNIFTY, RELIANCE, PATANJALI, COLPAL and HDFCBANK, and it adds the expiry tick list so the profile picks expiries the way Sensibull's does.

PR #2073, opened 2026-09-16
feat(oi-profile): put OI Profile on the chart, across expiries, verified against NSE

aakash-code:
## OI Profile: on the chart, across expiries, checked against NSE

OI Profile was one page at `/tools`. It is now also an overlay on the `/trading`
chart, it sums as many expiries as Sensibull's own does, and the figure the whole
thing hangs on - the previous session's open interest - is read from one file NSE
publishes for the entire market rather than asked of the broker one contract at a
time.

This replaces #2071, which was opened before the overlay had been run through a
live session. Everything in that PR is here, rebased on current `main`, with what
the market taught it since.

### On the chart

Calls sit above each strike line and puts below it. The bars are pinned to the
right edge in screen space rather than anchored to a time, because that is where
a profile belongs and panning must not carry it away. Hovering a strike names it
- strike, call and put - in lakhs and crores, the way the desks that read these
numbers write them. `/trading` gets a toolbar toggle and a settings gear, shown
only when the indicator is installed.

### The anchor, which is the part that was wrong

Change in OI is measured from the previous session's close. This was checked
rather than assumed, during the 16-Sep-2026 session against NSE's own option
chain:

| | contracts |
| --- | --- |
| NSE open interest, NIFTY22SEP2623200PE | 189,110 |
| NSE change in open interest | 119,065 |
| so NSE measures from | 70,045 |
| x lot size 65 | **4,552,925** = the previous daily candle, exactly |

An anchor taken from the current session's *opening bar* reads 6,614,400 on the
same leg, and reports 85% where NSE and Sensibull both report 170%. The overnight
step is real open interest and the day's change owns it.

### Where that anchor comes from, and why it changed

Reading it from the broker costs one history call per option leg. Behind the
process-wide 350ms gate in `services/history_service.py` that every history
caller shares, a chain is 25-30s, and an underlying nobody had opened that day
had no anchor for any of its legs. Two things followed, both seen in a live
session:

- `openalgo-charts` times its own history request out at **15s**, so the chart's
  request for the newly selected symbol expired behind the anchor pass. The user
  sees `History error: History request timed out` and the symbol never draws.
- Switching symbols twice in a minute put two chains' worth of broker calls in
  flight together. Thread count reached 50 and the Upstox feed dropped on
  ping/pong.

NSE already publishes that anchor for the whole market - the F&O bhavcopy, one
zipped CSV per session. `services/nse_oi_bhavcopy.py` resolves, downloads and
parses it into `{OpenAlgo symbol: previous close OI}`.

| | |
| --- | --- |
| Size | ~1.1 MB zipped |
| Rows | 35,657 contracts (35,010 options) |
| Underlyings | 216, which is all of them |
| Download and parse | 0.58-1.02s |
| Frequency | once per day, for every symbol |

It picks the session *before* the one on screen: the previous trading day during
a session, and the one before the last that traded on a weekend or holiday, so a
chart opened on a Saturday still shows Friday's build. An unpublished date 404s,
so holidays are skipped without a calendar walk.

### Nothing fetches an anchor inside a request

`_fetch_prev_session_oi` answers with the anchors already cached, reports
`oi_change_pending` when some are missing, and hands the rest to one worker.

- **One worker** (`max_workers=1`), module-level, so the broker never sees two
  chains warming at once.
- **Newest selection wins.** A later request supersedes the running job by
  generation check instead of queueing behind it; anchors it had already cached
  are kept, so nothing is refetched. Rapid symbol switching cannot stack calls.
- **Only the worker may download the file.** A request reads it if it is already
  in hand, which is why every symbol switch after the first is instant *and*
  complete rather than pending.
- A pending payload is **not** pinned in the shared profile cache, or a client
  polling for the rest would be handed the same gaps for the whole TTL.
- `app.py` warms the file from the existing `_restore_caches_background` thread,
  which already waits for the database and holds the app context the market
  calendar needs. So `oi_change_pending` is not seen in normal operation; it
  survives for a failed download, BFO and crypto.

Clients shorten their beat while pending - 10s on the page, 15s on the overlay -
instead of waiting out the three-minute beat.

### Verified

**Against NSE, leg by leg, live**, converted to contracts: BANKNIFTY, RELIANCE,
PATANJALI, COLPAL and HDFCBANK match row for row with most legs exact, and NIFTY
matches inside the drift between two snapshots taken minutes apart. Absolute open
interest matched Sensibull to a tenth of a percent throughout. On the chart, with
the same five expiries ticked as Sensibull's overlay, strike 23100 reads 16.40L
call and 83.05L put here against their 16.32L and 81.79L.

**The file against the broker path**, anchor for anchor, 134 live legs:

```
NIFTY 22SEP26 (82 legs)      identical 82/82   disagreements 0
TVSMOTOR 29SEP26 (52 legs)   identical 45/52   disagreements 0
                             file had 7 anchors the broker returned nothing for
                             broker had 0 anchors the file missed
```

Zero disagreements. Where both sources answer, the figure is identical. The
asymmetry runs one way only: 7 TVSMOTOR legs carry an anchor the broker's daily
history returned nothing for, including `TVSMOTOR29SEP263850CE`, which the old
path logged as unreadable during a live session.

**The file against the symbol master**, for 16-Sep-2026: all **216** NFO option
underlyings present with none the master does not know, and all **35,010** option
rows map onto a symbol already in `symtoken`, no spelling mismatches. The anchor
this PR is pinned to reconciles too - 7,467,135 in the file less its 2,914,210
day's change is the 4,552,925 above - so the file is in the broker's own units
with no lot-size conversion.

### Measured, market open

| | before | after |
| --- | --- | --- |
| NIFTY 22SEP26, 82 legs, anchor pass | 28.82s | 0.0003s |
| TVSMOTOR 29SEP26, 52 legs | 24.35s | 0.0003s |
| full request on a cold symbol | 34-40s (117s observed) | 0.16-0.28s |
| chart overlay, `include_change: false` | 0.5s | 0.5s |
| the whole NSE F&O market | ~17,700 calls, ~1h38m | one file, under 1s |

### Expiries are ticked, not counted

The dialog lists the underlying's nearest expiries for the symbol on the chart
and sums the ticked ones, with the nearest-N fallback shown pre-ticked so it is
visible rather than implied. The exchange resolves itself - the BSE indices to
BFO, everything else to NFO - so a SENSEX chart no longer asks NFO for options
that live elsewhere and draws nothing.

### Cost

Every strike is two legs to quote, so the strike window is a setting. The overlay
asks for no candles, refreshes on a three-minute beat with a random offset so many
open charts do not arrive at the broker in the same instant, stands down with the
tab, and stretches to fifteen minutes when the market is shut. The backend shares
one answer across clients for a short TTL, so a second chart on the same
underlying is free.

### Unknown is not zero

A zero anchor subtracted from a live number reports the leg's entire open interest
as today's build, so one unreadable leg paints a strike as a huge fresh write.

- A **failed or empty broker read** is cached as unknown and draws no change. It
  is cached rather than retried so a chain of unreadable legs cannot re-storm the
  broker every beat; the trade-off is that a transient failure leaves that leg
  blank for the session.
- An **explicit zero from NSE** is a real anchor of 0 - the whole of today's open
  interest is a genuine build. NSE lists zero-OI contracts explicitly (about
  16,800 of the 35,010), so a contract missing from an underlying the file *does*
  cover is also anchored at zero. An underlying the file does not cover at all -
  an F&O name listed today - still goes to the broker, because zeroing its whole
  chain would report every strike as built today.

### Not covered by the file

- **BFO** (SENSEX, BANKEX). BSE publishes an equivalent in a different format;
  not implemented, so BFO keeps the per-leg path.
- **Crypto.** No such file exists.
- **The drag-selected intraday window.** Dragging a range on the candles measures
  the build across that window, which needs intraday bars a daily settlement file
  cannot give. Still one broker call per leg, still inline - it is explicitly
  user-initiated and one-off. Historify's `market_data` already holds one-minute
  bars with an `oi` column for 5,970 NFO symbols, which is the obvious local
  source if this path starts to hurt.

### Every file in this PR

| File | What |
| --- | --- |
| `services/nse_oi_bhavcopy.py` | New. Resolves, downloads and parses NSE's F&O bhavcopy into previous-session OI per OpenAlgo symbol. |
| `services/oi_profile_service.py` | Multi-expiry sums, shared answer cache, strike window, windowed changes, and the anchor pass moved off the request behind one worker. |
| `blueprints/oiprofile.py` | Expiry list, strike count and `include_change` / `include_candles` on the endpoint. |
| `app.py` | Warms the bhavcopy from the existing cache-restore thread. 17 lines. |
| `strategies/indicators/oi_profile_live.js` | New. The chart overlay: screen-space bars, hover tooltip, expiry ticks, beat with jitter, shorter beat while pending. |
| `frontend/src/lib/trading/terminal.ts` · `customIndicators.ts` | Toolbar toggle, settings gear, and the `expiries` input type in the loader. |
| `frontend/src/components/trading/IndicatorSettingsDialog.tsx` · `ChartPane.tsx` | The expiry tick list and wiring it to the chart. |
| `frontend/src/pages/OIProfile.tsx` · `api/oi-profile.ts` | Expiry picker, drag window, strike count, and the shorter refresh while `oi_change_pending`. |
| `docs/oi-profile-anchors.md` | New. Both anchor sources, which session's file is correct, the unknown-is-not-zero rules, the verification, and the tuning variables. Registered in `docs/INDEX.md`. |
| `docs/custom-indicators.md` · `.claude/skills/chart-indicator/*` | The `expiries` input type documented and validated. |
| `docs/prompt/services_documentation.md` | One line pointing the service row at the new doc. |
| `test/*` | See below. |

### Note for reviewers

The input vocabulary for custom indicators is declared in three places - the
loader that registers a user module, the dialog that draws the row, and the
chart-indicator skill's validator. The expiry tick list is a new type, and all
three now agree. The loader's copy is the one that bites: a type it does not know
makes registration throw, and the chart then says `unknown indicator` with
nothing about the input that caused it. A test pins it.

`nsearchives.nseindia.com` needs only a browser User-Agent, no warmed cookie,
unlike `www.nseindia.com` (see `services/nse_events_service.py`). Failure is soft
throughout: no file means the per-leg path takes over, and it retries after ten
minutes rather than on every leg that misses.

### Checks

`test/test_nse_oi_bhavcopy.py` (new - builds a bhavcopy in memory, no network:
symbol mapping including fractional strikes, the 4,552,925 arithmetic, both
absence rules, NSE-only gating, the failure cool-off), plus
`test/test_oi_profile_multi_expiry.py` (the anchor definition, unknown-is-not-
zero, multi-expiry sums, and that a cold underlying answers at once, that a newer
selection abandons the chain being warmed, and that a failed warm pass cannot
wedge the pending state), `test/test_oi_profile_windowed.py`,
`test/indicators/oi_profile_live.test.mjs`, the customIndicators suite, tsc, and
the chart-indicator skill's `validate.mjs` and `coverage.mjs` all pass. The NSE
path is off by default in the multi-expiry tests so those keep exercising the
per-leg fallback.

Full repo suite run twice - once on this branch, once on a clean `upstream/main`
checkout - produces byte-identical failure sets (20 pre-existing entries, none of
them in these files), so nothing here breaks a test that passed before.

---
## Summary by cubic
Adds OI Profile as an overlay on the `/trading` chart, alongside `/tools`, summing up to six selected expiries with bars pinned to the right edge in screen space. Change in OI now anchors on the previous session's close from the cached NSE F&O bhavcopy instead of per-leg broker history calls, so cold symbols draw fast while missing anchors fill in the background; failed legs stay unknown rather than zero-anchored.

- The indicator's input types now match `main`: the `expiries` type is declared consistently across the loader, settings dialog, and chart-indicator validator (registration previously threw, showing `unknown indicator`), and the Underlying field was moved from `symbol` to `text` since `main` dropped the former.
- BFO, crypto, uncovered underlyings, and drag-selected intraday windows keep the per-leg broker-history path.
- A merge with current `main` retains `indicatorId` on indicator rows — the OI Profile toolbar toggle matches on it — alongside upstream's Alerts and comparison state, and a plotted indicator that produces no points now raises a toast instead of drawing nothing silently.
- Dropped the fork-only default indicator list the merge reintroduced: a fresh chart opens with nothing, as upstream.

**Verification**
- Tests cover bhavcopy parsing, multi-expiry aggregation, windowed changes, custom-indicator registration, and screen-space rendering.
- Live validation matches the broker path across 134 legs and confirms the previous-close anchor against NSE data.

<sup>Written for commit 139d25be316b9886075a0fe7aded03a06427affe. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/2073?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


aakash-code:
## Follow-up pushed: the Change-in-OI anchor no longer costs a broker call per leg

Pushed `344277945` onto this branch. One behavioural bug and its cause, plus the
data source that removes it.

### What was wrong

Change in OI needs the open interest each leg closed the previous session on.
This PR fetched that anchor inline, one broker history call per leg, so opening
a chart on an underlying nobody had looked at that day was a 30-90s request
(117s observed across multiple expiries). Two consequences, both seen in a live
session:

- `openalgo-charts` times its own history request out at **15s**, so the chart's
  request for the newly selected symbol expired behind the anchor pass. The user
  sees `History error: History request timed out` and the symbol never draws.
- Switching symbols twice in a minute put two chains' worth of broker calls in
  flight together. Thread count reached 50 and the Upstox feed dropped on
  ping/pong.

### What changed

NSE already publishes that anchor for the whole market — the F&O bhavcopy, one
1.1MB zip per session carrying 35,010 option contracts across all 216
underlyings. One download replaces every per-leg call, for every symbol, for the
day.

- `services/nse_oi_bhavcopy.py` resolves, downloads and parses it into
  `{OpenAlgo symbol: previous close OI}`, picking the session before the one on
  screen so a chart opened on a Saturday still shows Friday's build.
- The anchor pass moved off the request. A request answers with what is cached,
  reports `oi_change_pending`, and hands the rest to one worker; a newer
  selection supersedes the running job instead of queueing behind it, so rapid
  symbol switching cannot stack broker calls.
- Warmed at startup from the existing cache-restore thread, so
  `oi_change_pending` is not seen in normal operation.
- BFO, crypto and the drag-selected intraday window keep the per-leg path.

### Verification

Anchor-for-anchor against the existing per-leg path, 134 live legs:

```
NIFTY 22SEP26 (82 legs)      identical 82/82   disagreements 0
TVSMOTOR 29SEP26 (52 legs)   identical 45/52   disagreements 0
                             file had 7 anchors the broker returned nothing for
                             broker had 0 anchors the file missed
```

Against the symbol master, for the 16-Sep-2026 file: all **216** NFO option
underlyings present with none the master does not know, and all **35,010**
option rows map onto a symbol already in `symtoken` with no spelling mismatches.

The figure this PR is already pinned to also reconciles: NIFTY22SEP2623200PE
reads 7,467,135 in the file with a day's change of 2,914,210, and
7,467,135 - 2,914,210 = 4,552,925, the 15-Sep close verified against NSE. Same
units as the broker feed, no lot-size conversion.

### Measured, market open

| | before | after |
|---|---|---|
| NIFTY 22SEP26, 82 legs, anchor pass | 28.82s | 0.0003s |
| TVSMOTOR 29SEP26, 52 legs | 24.35s | 0.0003s |
| full request on a cold symbol | 34-40s | 0.16-0.28s |
| chart overlay, `include_change: false` | 0.5s | 0.5s |
| the one daily cost | - | one download, 0.58-1.02s |

### Notes for review

- **Absence in the file is evidence, with a guard.** NSE lists zero-OI contracts
  explicitly, so a contract missing from an underlying the file *does* cover is
  anchored at zero. An underlying the file does not cover at all (an F&O name
  listed today) still goes to the broker, because zeroing its whole chain would
  report every strike as built today.
- **A failed broker read stays unknown and draws nothing**, never a zero anchor.
  It is now cached as unknown rather than retried every beat, so a chain of
  unreadable legs cannot re-storm the broker; the trade-off is that a transient
  failure leaves that leg blank for the session.
- `nsearchives.nseindia.com` needs only a browser User-Agent, no warmed cookie,
  unlike `www.nseindia.com` (see `services/nse_events_service.py`). Failure is
  soft everywhere: no file means the per-leg path takes over and it retries
  after 10 minutes.

### Docs

`docs/oi-profile-anchors.md` is new and registered in `docs/INDEX.md`. It records
both anchor sources and why the order between them matters, which session's file
is correct on a trading day versus a holiday, the unknown-is-not-zero rules, the
verification above, what the file does not cover, and the tuning variables.

Tests: 29 pass across `test/test_nse_oi_bhavcopy.py` (builds a bhavcopy in
memory, no network) and the two existing OI Profile files. The NSE path is off
by default in `test_oi_profile_multi_expiry.py` so those keep exercising the
per-leg fallback. `test/indicators/oi_profile_live.test.mjs` still passes.

PR #2079, opened 2026-09-18
feat(dhan): add 30m, 4h, weekly and monthly history intervals

nimchand87:
## Summary

Dhan's history API serves 1, 5, 15, 25 and 60 minute candles and daily candles only. This adds **30m, 4h, W and M** for Dhan by aggregating native candles, so they appear in the `/trading` timeframe menu and are accepted by `/api/v1/history` like any native interval.

## How it works

- All logic is in a new file, `broker/dhan/api/resample.py`. The change to `data.py` is 7 lines: an import, one `timeframe_map.update(...)` so `intervals_service` advertises the new intervals, and an early return at the top of `get_history`.
- **30m and 4h** are built from 15m candles and anchored to the session open, not midnight: 09:15 for NSE, BSE, NFO, BFO and the indices, 09:00 for MCX, NCO, CDS and BCD. An NSE 4h day is therefore 09:15-13:15 and 13:15-close. 15m is the base because it divides evenly into buckets from either anchor. Buckets never span two days.
- **W and M** are built from daily candles and stamped on the Monday and the 1st, even when that day is a holiday. The requested window is widened to whole weeks or months (end clamped to today), so a chart paging history backwards never receives a partial bar at the edge of a window that would then replace the complete one.
- **Timestamps** follow the convention `data.py` already uses (a naive IST wall-clock datetime passed through `.timestamp()`). Buckets are computed on that wall clock and stamped back through the same conversion, so derived bars line up with native ones regardless of the server's timezone.

## Testing

- New `test/test_dhan_resample.py`, 14 tests: session anchors for NSE and MCX, no cross-day buckets, OHLCV aggregation, a holiday Monday, monthly bars, derived stamps matching the native daily convention, range widening, the clamp to today, and the intervals being advertised. All pass, and also pass with the process timezone forced to UTC (`TZ=UTC0`).
- Checked against a live Dhan session by calling `BrokerData.get_history` directly (SBIN, NSE, 18 Sep 2026):
  - the 4h 13:15 bar's volume equals the sum of the 13:15 and 14:15 hourly bars
  - each 30m bar equals its two 15m bars
  - the week of 14 Sep equals the sum of its four daily bars, with the high and low from the correct days
- CI-safe selection from CONTRIBUTING.md: 66 passed. `test_python_editor.py::test_editor_functionality` fails the same way on unmodified `main` on Windows (`WinError 32`, file in use), so it is unrelated to this change.
- `ruff check` and `ruff format --check` are clean on the new files.

## Notes

- No frontend change is needed. `intervals_service` already groups W and M, and the chart already handles them for other brokers.
- `25m` stays in Dhan's map and is still filtered out by `SUPPORTED_INTERVALS`, as before.
- Seen while testing, not changed here: the native `D` path does not retry an empty historical response and falls back to today's quote candle alone. W and M inherit that behaviour from `D`.

---
## Summary by cubic
Adds 30m, 4h, weekly, and monthly history intervals for Dhan by aggregating native candles, so these timeframes now appear in the menu and are accepted by `/api/v1/history`.

- 30m/4h aggregate from 15m anchored to the session open (09:15, or 09:00 for MCX/NCO/CDS/BCD), folding stray pre-open auction bars into the first bucket; W/M aggregate from daily bars stamped on Monday/1st even on holidays.
- Weekly and monthly requests widen the window to whole periods, clamped to today, so backwards-paging charts never get a partial edge bar.
- Aggregation lives in `broker/dhan/api/resample.py`; `data.py` only advertises the intervals and routes `get_history` calls to it.
- No frontend change is needed; `25m` remains filtered from `SUPPORTED_INTERVALS` as before.

<sup>Written for commit c8a6280412a9fb59f99a7bb35230fd289ffcf721. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/2079?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


nimchand87:
Pushed one more commit (c8a6280) fixing an edge case I found while testing older history.

Dhan's 15m history carries a stray pre-open bar stamped 09:07 or 09:08 on a few days a year (SBIN 2021-08-04, 2021-08-12 and 2021-08-13; RELIANCE and INFY on two days each in 2021). Bucketing it from the 09:15 anchor gave it a bucket of its own, so 30m showed an 08:45 bar and 4h a 05:15 bar on those days.

The fix clamps anything before the session open into the first bucket, so the auction bar's price becomes the opening price of the 09:15 bar and its volume is counted there. Added a test for both intervals. Re-checked SBIN, RELIANCE and INFY for all of 2021 against the live API: no derived bars before 09:15, and each stock now has 496 4h bars (248 sessions x 2).

The native 15m and 1m series still include those pre-open bars as Dhan returns them; this PR does not change native intervals.

Kalaiviswa:
Thank you for putting this together and for testing the edge cases so thoroughly. We appreciate the work
that went into it.

Our policy is to expose only the historical intervals a broker provides through its API. Since Dhan does
not provide 30-minute, 4-hour, weekly, or monthly candles directly, we’re unable to merge these as Dhan-
supported intervals.

Clients can still retrieve Dhan’s supported 15-minute and daily candles through OpenAlgo and aggregate
them in their own applications. Thank you again for the contribution and your understanding.

PR #2080, opened 2026-09-18
fix(installer): escape closing parentheses in update.bat echo lines

nimchand87:
## Summary

On Windows, `install/update.bat` stops with "uv is not installed" on a machine where uv **is** installed as a standalone tool (and not also as a Python module), so the update never gets past the prerequisite check.

## Cause

Several `echo` lines inside `if (...) ( ... )` blocks contain an unescaped `)`, for example:

```bat
if not errorlevel 1 (
    set "UV_CMD=uv"
    echo   [OK] uv found (standalone)
) else (
```

cmd reads that `)` as the end of the block, so the message is cut short and the `else` branch runs as well. Running the uv-detection block from `main` with uv on PATH prints:

```
  [OK] uv found (standalone
[ERROR] uv is not installed.
Install with: pip install uv
```

and the script exits.

## Fix

Escape the closing parenthesis as `^)` in the 7 `echo` lines that have one. Only the echoed text changes; no logic is touched.

With the fix, the same block prints `[OK] uv found (standalone)`, sets `UV_CMD=uv` and continues.

## Testing

Windows 11, cmd. I extracted the uv-detection block from `main` and from this branch into standalone scripts and ran both with uv on PATH, which produced the before and after output above.

---
## Summary by cubic
Escapes closing parentheses in `install/update.bat` echo lines so cmd no longer misparses the parenthesized `if` blocks. Previously, an unescaped `)` cut messages short and triggered the `else` branch, causing the update script to falsely report "uv is not installed" and exit even when uv is on PATH. Now the echoed text displays correctly and the script continues. Only the echoed text changes; no logic is touched.

<sup>Written for commit 9fbd85f68ef9a1bf169e20fa94f639a2735e5359. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/marketcalls/openalgo/pull/2080?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


marketcalls:
Merged, thank you. I reproduced this on Windows 11 (build 26200, cmd) before merging and your diagnosis is exactly right, including the detail that cmd falls through into the `else` branch rather than just truncating the message.

One thing worth recording: the failure is not limited to a standalone uv. Running the block from `main` with uv absent from `PATH` but available as a Python module gives the same result:

```
  [OK] uv found (Python module
[ERROR] uv is not installed.
Install with: pip install uv
exit code 1
```

So `install/update.bat` always exited at the prerequisite check, for every Windows user, whichever way uv was installed. The unescaped `)` has been there since the file was added in `bf8fa2a79`, so the Windows updater had never once run past that gate.

I checked the fix is complete: the only unescaped `)` left in an `echo` is the `Current version: ... (branch: ...)` line, which sits at block depth 0 where cmd handles it fine, and the repo's other two batch files are clean of the same pattern. All seven escaped lines print the intended text at depth 0, 1 and 2 with delayed expansion.

Because your fix lets the script run for the first time, it also exposed a second bug waiting behind it, which I have pushed to `main` as a follow-up in `491deca07`. Step 1 built its backup folder name from `wmic os get localdatetime`, and WMIC is gone from Windows 11 24H2 and later. With `wmic` missing, `TIMESTAMP` expanded to literal text containing a colon, `md` and `copy` both failed into suppressed output, and `BACKUP_COUNT` was counting the source files rather than the copies. Running the real Step 1 block against a simulated install printed:

```
[Step 1/5] Backing up databases...
  Backed up: openalgo.db
  Backed up: logs.db
  Backed up: historify.duckdb
  [OK] Backup location: ...\db\backup_~0,8...DT:~8,6
```

with nothing actually copied, and the update would then have gone on to `git pull` and run migrations against live trading databases. The follow-up reads the clock from PowerShell in the format `update.sh` already uses, stops the update if the folder cannot be created, and counts a database as backed up only once its copy is present.

Nothing for you to change here. Nice catch, and thanks for including the before and after output, it made this quick to confirm.

PR #2084, opened 2026-09-20
feat: store and serve OpenScript studies for the trading chart

marketcalls:
Stores and serves OpenScript studies for the `/trading` chart, and the alert
surface that grew around them.

## OpenScript studies

A trader's own `.oscript` files compile in the browser and register beside the
built-in studies, under **My scripts** in the indicator picker. A study's legend
row carries a braces button that opens its source in the script panel.

`openalgo-script@0.4.0` is pinned and published, which is what makes the
lockfile resolve again: `0.3.0` was pinned and never published, so `npm install`
in `frontend/` failed on ETARGET and CI never reached a test.

## Alerts

**A right-click makes the alert.** Right-click at the price to be watched, or on
a study plot or a drawing, and it is created there and then: armed to fire once
on a confirmed bar, expiring in two months, with a sound and a desktop
notification. No form, because the price is the one thing a form would ask for
and pointing at it has already given it.

**The toolbar's Alerts button opens the form**, for an alert needing a
condition, a trigger, an expiry or a message the defaults do not give it. Both
paths seed from one function, so the gesture produces exactly the alert the form
would have proposed; seeding them separately is how a right-click ends up
quietly making something different from what the form describes.

**The list moved to the right rail**, with a **Log** tab of what has fired this
session. It used to live inside the modal, so reading what was armed meant
covering the prices that would arm it.

**How you are told is chosen per alert.** Sound and a desktop notification are
on by default and neither leaves the machine. Telegram and WhatsApp go through
the same services order notifications already use and are off unless asked for,
because a chart alert quietly joining somebody's order feed is not a decision to
make on their behalf. A channel that refuses names itself, the others still go,
and nothing is retried.

**A chart with an armed alert keeps fetching while its tab is hidden.** Hiding
stopped the poll and the bar-close repair unconditionally, and the default alert
policy is precisely the one that waits for a bar to close. A chart with nothing
armed still stops.

**A message can carry values from the bar that fired it** — `{{ticker}}`,
`{{price}}`, `{{close}}`, `{{interval}}` and seven more — so a notification on a
locked phone carries the number rather than sending you back to the chart. One
spelled wrong is left as typed rather than blanked.

**A dragged alert lands on a tradeable price.** Dropping a line where the axis
read 1,260.55 used to store `1260.5486842105263`. It is rounded to the
instrument's tick, and a name we generated follows the price; a name you typed
is never touched.

## Chart 2.4.8, and two fixes that were ours

- **openalgo-charts 2.4.8**: press and hold the plot to pan. The
  `chart-indicator` skill moves with the pin, as `CLAUDE.md` requires; the index
  is unchanged at 387 exports and coverage still prints COVERAGE COMPLETE.
- **Panning through price**: `buildChart` pinned `mousePan: 'horizontal'`, so
  dragging never moved through price. The engine's default is `both` and the
  setting was never ours to make: it is the trader's under **Mouse drag** in the
  Axes tab, and pinning it also made their choice the one thing Reset would not
  return to.
- **Searching for a symbol with a hyphen**: the search box splits on operators
  so a half-typed expression can look up its second leg. Splitting on `-` made
  `BAJAJ-AUTO` unsearchable — the box split into `BAJAJ` and `AUTO`, and typing
  the hyphen emptied the list. Both the whole box and the leg are searched now.

  The same omission is in the engine's grammar, where `SYM_BODY` admits `&` and
  not `-`. That is why `M&M` loads and `BAJAJ-AUTO` does not, and it is written
  up for the library separately.

## Wording

"Armed" is gone from anything a trader reads: **One-Click ON**, "Active as soon
as it is saved", "Keep the tool selected after drawing". `wording.test.ts` fails
on the word returning to a visible string, because nothing pinned it before.

## What this does not do

Alerts are evaluated by the chart that is open, so closing the tab stops them.
Server-side alerts are recorded as a deliberate deferral in OpenScript's
`ROADMAP.md`, with the cost that decides it and a note to revisit while
implementing the next phases.

## Validation

- 2,607 frontend tests across 163 files; TypeScript, lint and production build.
- Every new test mutation-checked. Two found real defects while being written:
  `notificationsSupported` tested the key rather than the value and threw on a
  window carrying the name with nothing behind it, and `toAlertInput` wrote a
  payload member holding `undefined`, which is not JSON and which the engine
  validates. A third test passed under its own mutation and was rewritten to
  read rendered row order.
- Checked on the running page: the create form, the rail panel, grab-to-pan,
  dragging an alert line, and the readings grid drawing inside its own columns.
- Documentation updated: `docs/userguide/32-charting-terminal/README.md` and a
  stanza in `docs/CHANGELOG.md`.

Not claimed: no live-market run. The market was closed throughout, so a fired
alert was reproduced from a stored condition rather than from a tick, and the
Telegram and WhatsApp sends are covered by tests against their request shapes
rather than by a delivered message.


marketcalls:
The editor half of the language landed in `openscript` (commit `234531c`, version `0.3.0`), and this branch now colours a script with it instead of reading the token stream here.

**What changed in this repository**

- `frontend/src/lib/trading/openscriptHighlight.ts` asks `openalgo-script/editor` for the pieces of a source. The two rules it used to apply itself are gone: a reserved word was "a token whose kind is its own text", which is also true of every punctuation mark, and the test that a kind is made of letters answered yes to `hexColor`, `indent`, `newline` and `dedent`, so a colour literal and a line's leading spaces were both painted as identifiers. Comment recovery was here too and is now in the package, in one place, for every host.
- What stays here is the palette: which of the language's eleven kinds share a colour in this product, and the call versus value distinction for a library name, which is a decision about this editor rather than a fact about the language. A `color` kind and its CSS variable are added.
- Two tests for the two kinds that used to be painted as names. The existing tests are unchanged and still pass, including the one that says the spans reconstruct the source exactly. The frontend suite is green: 153 files, 2457 tests.

**One thing to do before this can build**

`frontend/package.json` now asks for `openalgo-script@0.3.0`, which is committed and pushed on `openscript` `main` but **not yet published to the registry**. Until it is, `npm install` here will not resolve it, and `frontend/package-lock.json` still names `0.2.0` on purpose: the lock's `resolved` and `integrity` come from the published tarball and cannot be written by hand.

So the order is: publish `openscript` `0.3.0` (the release workflow, dispatched against tag `v0.3.0`), then `npm install openalgo-script@0.3.0` in `frontend/` to update the lock, then merge.

To try it before publishing, `npm pack` in the `openscript` checkout and install the tarball into `frontend/`, which is how the suite above was run.

